PR #658, opened 2014-04-06
AddImage button

N0U:
Now it is posible to attach an image to a message. 
If an image is too big, a user will be asked if the image should be scaled. 
All images are coded in base64.


timothyparez:
>  I am interested in supporting small attachments 
Are we talking about supporting messages in binary format instead of text only?  
Allowing BitMessage to serve as a backend for any kind of data exchange, protected
against spam by the PoW of course.

Atheros1:
Yes, that is the goal.

timothyparez:
Awesome!

PeterSurda:
I am undecided just like @Atheros1 , however I am more inclined to accept because my focus is usability. However, there are conflicts and there are UI changes being done. @N0U , I recommend you try the latest UI changes in my forks, https://github.com/mailchuck/PyBitmessage, and try to create a pull request for that one. That will have a better chance of eventually getting into the official PyBitmessage.

PR #755, opened 2014-12-27
dont freeze UI when mass-moving to trash

themighty1:
The fix is two-fold:
1. Make a bulk query to SQL rather than one-by-one.
However, 1. alone does not solve the UI freeze.
2. Remove ranges of rows with model().removeRows()

Note: TableWIdget gets inserted/removed to/from multiple threads, so it is desirable to implement a Lock() on access.


PeterSurda:
This is now done through sqlExecuteChunked.

PR #782, opened 2015-02-21
german translation and tray icon

felidosz:
german translation and minimize interface on click at tray icon under linux


Atheros1:
Should this do something on some linux systems? It does nothing on Ubuntu.

felidosz:
Without this change i can't minimize the interface to tray with one click on the tray icon. You have to right click to open the tray menu and then choose show bitmessage.
The "if" enables the minimize-by-one-click-on-the-tray-icon just for windows systems (and disables it for Linux Systems)

Am 22. Februar 2015 03:28:03 MEZ, schrieb Jonathan Warren <notifications@github.com>:

Atheros1:
It doesn't work in Ubuntu for me. I don't think it causes any harm so it should be mergeable but it doesn't work for me. It would be good if someone on OSX could test it.

felidosz:
I have Xubuntu, maybe this is the reason ;)

PR #790, opened 2015-03-21
Refactored the code in api.py

timothyparez:
Some refactoring for the api.py source.
Instead of having a single, really long, method, the handler code for each
request is not defined in its own method.

This should make it easier to read, modify and maintain.


Atheros1:
Thank you very much Timothy

PR #797, opened 2015-05-31
Add dockerfile

must-:
Enables running a PyBitmessage daemon in a Docker container.


TcM1911:
I'm agreeing with @bashrc

PeterSurda:
I'm not familiar with docker other than what you can read on their website, but what @bashrc said makes sense. Unfortunately I can't accept it like this.

PR #798, opened 2015-06-02
set log level from config file

imrehg:
Read the log level settings from the "loglevel" value in the
"bitmessagesettings" section of the config file. The possible
values map directly to the ones allowed by the logger module:
DEBUG, INFO, WARNING, ERROR, CRITICAL
Only allow these values, and cast to upper case to be more easy
going in the settings.

### DIFF
--- a/src/debug.py
+++ b/src/debug.py
@@ -24,7 +24,9 @@ import helper_startup
 helper_startup.loadConfig()
 
 # TODO(xj9): Get from a config file.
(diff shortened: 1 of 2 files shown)


PeterSurda:
I need to take a closer look to how the logging works, but on first glance it looks ok.

PeterSurda:
Logging was rewritten, now you can configure it via a separate config file.

Issue #79, opened 2013-03-27
UPnP Support


spumer:
Please, make a UPnP support standart.

PeterSurda:
I think @lightrabbit's patch is ok, if he makes the improvements I suggested I plan to integrate it. I'll close this when I merge it.

PR #810, opened 2015-08-22
Add UPnP support.

lightrabbit:
This feature request had been put out https://github.com/Bitmessage/PyBitmessage/issues/79#1 but never implemented before, then I tried to implement it.

### DIFF
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2.7
+﻿#!/usr/bin/env python2.7
 # Copyright (c) 2012 Jonathan Warren
 # Copyright (c) 2012 The Bitmessage developers
 # Distributed under the MIT/X11 software license. See the accompanying
@@ -44,6 +44,7 @@ from debug import logger
 # Helper Functions
 import helper_bootstrap
 import helper_generic
+import upnp
     
 
 def connectToStream(streamNumber):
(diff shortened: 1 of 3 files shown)


peterwang1996:
The product of that broadcasting?

bashrc:
This might be a good idea, although I don't know enough about Upnp or the security implications to be sure.

lightrabbit:
I think there's no need to use 3rd-party UPnP library if we only create a port mapping on router with UPnP. We can simplely make a simple module only for creating port mapping.

PeterSurda:
Hello @lightrabbit ,

I have reviewed the patch, I haven't run it yet though. It looks like it may be working. Have you tried it yourself?

I don't like that if the port mapping fails, it will randomise it, and that it assumes the external and internal port are the same. I think that the user should be more in control of their own machine and would prefer if it only randomised the external port and kept the internal port fixed. I would add a new variable representing the external port to the shared.py, which initialises to the internal port and createPortMapping would set it once the mapping is established. Then you also need to change the assembleVersionMessage to use this external port instead of the one in the config file. Based on my understanding of the code, that should be sufficient to get it working.

I would also like to have some check for stack depth in createPortMapping, so that it cleanly fails if there's something wrong.

I would also prefer the ability to configure this option. It should default to off. So, I'd like you to add a variable to the config file that defaults to false, and a change to the QT Gui for this (preferably in the Listening Port section of the Network Settings tab you'll add a checkbox).

Once this change is in, I will review it again and test it as well.

PeterSurda:
@lightrabbit Any updates?

PeterSurda:
Hey @lightrabbit 

I tested it, fixed and merged it, it's now in mailchuck@b93308d7abedce21ae8ea4a4f5808be282f7abba

PR #812, opened 2015-09-13
Add proper filesystem check for NetBSD

krytarowski:
FreeBSD and NetBSD are compatible here.


PeterSurda:
I reviewed this one as well and it looks ok, I pulled it into testing and will merge into master during regular lifecycle.

PR #824, opened 2015-11-14
Single instance and pop up old instance

### LINKED ISSUE #630: Second bitmessage instance should pop up the window of the first one and quit, and not run in parallel
I launched the second copy of bitmessage from the same user on the same display, and it launched. It showed two windows and two tray icons.

This shouldn't happen. It should locate the first instance, make its window active and quit. Chrome browser, for example, does this.

The second instance also shouldn't be allowed to launch off the same database on another screen, since this doesn't make sense.

It is possible to find the window of the original instance through the window manager.

thelink2012:
Hopefully this fixes #630

### DIFF
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -44,7 +44,9 @@ from debug import logger
 # Helper Functions
 import helper_bootstrap
 import helper_generic
-    
+
+# singleton lock instance
+thisapp = None
 
 def connectToStream(streamNumber):
     shared.streamsInWhichIAmParticipating[streamNumber] = 'no data'
@@ -139,6 +141,8 @@ if shared.useVeryEasyProofOfWorkForTesting:
 
 class Main:
     def start(self, daemon=False):
+        global thisapp
+        
         _fixWinsock()
 
         shared.daemon = daemon
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -143,7 +143,7 @@ class Main:
 
         shared.daemon = daemon
         # is the application already running?  If yes then exit.
-        thisapp = singleton.singleinstance()
+        thisapp = singleton.singleinstance("", daemon)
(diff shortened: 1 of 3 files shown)


thelink2012:
Works fine on Windows, just tried on Ubuntu and it gives *IOError: [Errno 6] No such device or address: '/home/thelink2012/.config/PyBitmessage/singleton.lock'*, thoughts?

PeterSurda:
Don't use "/", use path.join.

Maybe the error happens because it hasn't created the config directory at that time yet?

thelink2012:
Okay, that seems to be working fine now. I've tested on Windows 8 and Ubuntu 14.04.3 LTS (throught vmware). Please enlight me of any issue.

btw, I'm wasn't entirely sure, but is `myapp.appIndicatorShow()` the correct method to call to bring up the old instance when the new one tries to launch? That seems to work, but still, I have this feeling there should be another method.

PeterSurda:
I'll review it later today.

PeterSurda:
It looks better than the first one, but what about if it is a daemon? It should also not run twice in such a case either.

PeterSurda:
Also, I think `myapp.appIndicatorShow` is the correct method to call.

PeterSurda:
Also, I would kindly ask to PGP-sign your commit, we now have this policy for all commits.

thelink2012:
Oh, didn't notice daemons deleted the singleton that soon. Anyway, fixed.
And PGP signed both commits.

PeterSurda:
Pulled through https://github.com/mailchuck/PyBitmessage/pull/117

I reviewed the code and tested it as far as I could without spending excessive amounts of time. It seems to work. The main windows does not move in front of other windows on Linux or OSX, only on Windows, but that's a separate issue because it misbehaves similarly on other occasions too.

Thanks.

PR #826, opened 2015-11-21
Solve restore window size bug

### LINKED ISSUE #243: Maximized Window bug
# Issue
A Maximized window is not restored properly from a minimized state.
# Affects
Windows
# Steps to reproduce
* Open bitmessage and change the window size, so it will not fill up the entire screen.
* Maximize the window
* Minimize the window
* Restore the window (make it visible again)
* Click on the maximize Icon on the top right again. The window will not go back to the previous size it had.

# Solution
It seems like the window size is stored somewhere. This should be removed and the window handling should be done by windows.

thelink2012:
Fixes #243

wow
?!!!??!!!

### DIFF
--- a/src/bitmessageqt/__init__.py
+++ b/src/bitmessageqt/__init__.py
@@ -700,10 +700,7 @@ class MyForm(QtGui.QMainWindow):
         if not self.actionShow.isChecked():
             self.hide()
         else:
-            if sys.platform[0:3] == 'win':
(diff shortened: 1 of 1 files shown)


PeterSurda:
Thanks it works.

thelink2012:
Ohhhh, sorry for not PGP signing, remembered just now.

PeterSurda:
I just edited the files myself, I didn't even bother with merging.

PR #843, opened 2016-01-23
Added "run" tartget for easy launch under Linux

Quix0r:
What I have changed:
- some empty lines added to make it more readable
- "run" target added to Makefile for easy launch from source
- Updated INSTALL.md accordingly

Under Windows you may still need do a "cd src" and "python.exe ./bitmessagemain.py" to launch it.


sigoa:
the bash menu from ffox seems even easier ! :+1:

Quix0r:
Fixed conflict. Execuse me, what do you mean with ffox and bash menu, plesse?

PR #859, opened 2016-05-11
Grammar: "need-not" -> "need not"

yurivict:
If you search for "need-not" in google books, you will see that it is overwhelmingly used without the hyphen: "need not". "need-not" is (possibly) a rare noun, and is also (possibly) a UK spelling of "need not" as a verb.


sp2004c:
this is too complex a change, so it should be done in the next century.   🐫

PR #868, opened 2016-06-04
Removed Â in front of ©

sandakersmann:
Removed Â in front of © in the translation files.

### DIFF
--- a/src/translations/bitmessage_ar.ts
+++ b/src/translations/bitmessage_ar.ts
@@ -1446,7 +1446,7 @@ The &apos;Random Number&apos; option is selected by default but deterministic ad
     </message>
     <message>
         <location filename="../bitmessageqt/about.py" line="69"/>
-        <source>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Copyright Â© 2012-2016 Jonathan Warren&lt;br/&gt;Copyright Â© 2013-2016 The Bitmessage Developers&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</source>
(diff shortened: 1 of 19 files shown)


PeterSurda:
Thank you. I noticed this but wasn't able to find out where it's coming from. I reviewed this now, this looks like some oddity in pylupdate4, it needs to be fixed elsewhere. If I accept this patch, next time I run pylupdate4 it will break again.

bmng-dev:
The issue is pylupdate4 only uses a simple parser with basic analysis which doesn't pick up the advanced way the encoding is currently conveyed and thus defaults to latin1. There are a few ways to deal with this:
1. Replace non-ASCII characters with the appropriate HTML character entity reference or numeric character reference (in this case `&copy;`) however this solution only works where the source text is HTML
2. Explicitly specify UnicodeUTF8 (or some other identifer that ends in UTF8) when calling _translate where the source text contains non-ASCII characters
3. Use the -trUTF8-function option when running pylupdate4 (`pylupdate4 -trUTF8-function _translate bitmessage.pro`) however not all python files in bitmessageqt folder use _translate (that is some use QApplication.translate explicity instead) and where the plural (n) argument is being used the encoding argument needs to be removed
4. Reencode files with translatable strings containing non-ASCII characters into latin1

PeterSurda:
Thank you for the elaborate feedback, @bmng-dev . It looks like you're spot on. I ended up with a combination of 2 and 3. After looking at pylupdate4's fetchtr.cpp, I replaced the offending _translate with __trUtf8 with 2 arguments, which just calls the old _translate. Reducing the number of arguments is necessary, as you wrote, so that it does not insert the plural forms. Since there is only one place in the code where this is a problem, it's not a big deal. Stupidly, CODECFORSRC in the .pro file does nothing.

I fixed it in d9ca8d1e5e11f58d8895c904e8e238e6939ffd03, but I need to update the english translation source separately (that's in a different branch which is polled by transifex daily).

PR #879, opened 2016-07-15
Torifier patch 1

torifier:


torifier:
clarify about packagemonkey

PeterSurda:
This isn't needed anymore as add/join chan was unified and address is now optional.

PR #880, opened 2016-07-15
Torifier usability newchandialog join+create

torifier:
clarify that new chan == join chan


torifier:
I tested the dialog OK here.

PeterSurda:
I refactored newchaindialog because I plan to have most of the UI refactored.

PR #888, opened 2016-07-16
Fix python shebangs from `#!/usr/bin/env python2.7*' to `#!/usr/bin/python2.7'.

### LINKED ISSUE #887: fix shebangs
If possible, could you fix the python shebangs in "src/bitmessagemain.py" and
"src/bitmessagecli.py"?
Our shebang fixer did not stumble on more, but if you had more like those two they could get fixed too.
`#!/usr/bin/env python2.7.x' `#!/usr/bin/env python2.7' should get fixed in my opinion.

If you regard it as not necessary, could you please explain why.

ng-0:
Fix python shebangs from `#!/usr/bin/env python2.7*' to `#!/usr/bin/python2.7'.

* src/bitmessagecli.py: fix it.
* src/bitmessagemain.py: same.

This fixes #887.

### DIFF
--- a/src/bitmessagecli.py
+++ b/src/bitmessagecli.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2.7.x
+#!/usr/bin/python2.7
 # Created by Adam Melton (.dok) referenceing https://bitmessage.org/wiki/API_Reference for API documentation
 # Distributed under the MIT/X11 software license. See http://www.opensource.org/licenses/mit-license.php.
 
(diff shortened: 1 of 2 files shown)


bmng-dev:
`/usr/bin/env python2` is a more flexible and portable way of specifying the interpreter to use. This will search $PATH for python2 instead of expecting python2.7 at exactly `/usr/bin/python2.7`. PyBitmessage itself will then check whether a supported version of python is being used and continue or stop as appropriate.

ng-0:
Okay, let me then ask why these are the only two files which use /usr/bin/env.

Gentoo for example follows upstream expectations as far as I know, and they are providing python, python$version, python$version.$minorversion binaries/symlinks.

I'm okay with manually patching this behavior away for the distribution I am working on for PyBitmessage.

PeterSurda:
Merged into v0.6.

PR #891, opened 2016-08-23
Uniformized python invocation lines: #!/usr/bin/env python

yurivict:


bmng-dev:
This should have been done in #888. Most of these modified files were only intended to be run on their own during their active development and a few do absolutely nothing so it would be better if the shebang lines are removed (along with the __main__ clauses) rather than corrected.

Wouldn't it be better to use `python2` rather than `python`? `python` may link to a 3.x version on some systems but PyBitmessage won't work in 3.x.

Whatever is decided don't forget to check/update the execute permissions. Ideally though this should all be left to distutils/setuptools which will automatically correct the shebang lines and execute permissions of "script" files during build and/or install.

yurivict:
I agree, python2 is better. If they do nothing for the user, you should probably not install them at all.

bmng-dev:
I meant some files when executed on their own do nothing. They are still essential for the running of PyBitmessage; they are the equivalent of .so/.dll files for C programs.

yurivict:
Then these invocation lines should be removed.

PeterSurda:
I'm for removing these lines and only keep where it make sense (bitmessagemain, opencl, ...), and to use python2.7 instead of env. I got rid of env in the first place, it breaks pidof -x and I don't like that.

yurivict:
You can't assume the paths will be the same on different systems. ```#!/usr/bin/env <exec>``` is a standard way to do this.

You need to report a bug to ```pidof``` instead of hardcoding executable paths.
On my system (FreeBSD) ```pidof``` can't find pids of python scripts at all, no matter what the line is.

PeterSurda:
I'll leave it for future decision then, but I'll remove the line altogether from scripts that don't have a __main__.

g1itch:
Outdated.

PR #899, opened 2016-10-07
Hide connection notifications - configuration option

mhitza:
New option on the User Interface settings tab, to hide connection
notification messages.

Still left unclear with my changes:
 - [x] Is settings.ui ever used, or should I just revert my changes there?
       Reverted the changes in settings.ui

 - [x] Should I regenerate the translation files, and how do I do that?
       No, [see comment](#issuecomment-252410748)


bmng-dev:
> Is settings.ui ever used, or should I just revert my changes there?

Once upon a time the .ui files were used to generate the associated .py files but over time different developers have treated .ui files differently:
1. Edit .ui files and regenerate the .py files
2. Ignore the .ui files and edit the .py files
3. Edit the .ui files and at runtime dynamically compile into memory on demand

I think the second case applies to settings.ui - it has been deprecated in favor of manually editing settings.py. I don't know why, in such cases, the .ui files haven't been removed to avoid the situation where contributers such as yourself prefer to edit .ui files (either manually or with Qt Designer) that are no longer used.

> Should I regenerate the translation files, and how do I do that?

This a part of the build/release process that is missing from the repository. It is best left to @PeterSurda to do since he is the only one (that I'm aware of) that knows the proper procedure. I know the basic procedure is to `cd` to the translations directory and run `pylupdate bitmessage.pro` but maybe @PeterSurda does more than that.

PeterSurda:
@bmng-dev is correct with respect to settings.ui. @mirrorwish did some refactoring to put more emphasis to the .ui files but it's only partial. I'd like to move back into the .ui files where possible, but for the time being settings.ui is not used.

With respect to translations, the current procedure is that I occasionally manually regenerate the .ts files with

    pylupdate4 src/translations/bitmessage.pro

Then I commit only bitmessage_en.ts into the live branch, and merge that commit into the "translations" branch. Transifex will then pick it up around 10 am (CET) daily, and translators can work on it. Then prior to a new version release I download the current translations and update the live branch, I have a small script that does automates that last part so that I don't have to fiddle around with the zip and rename the files manually.

PeterSurda:
If you want to keep the default value at False, then editing class_sqlThread.py is strictly speaking not necessary, just recommended. safeConfigGetBoolean catches a missing value and defaults to False.

mhitza:
Made updates from the comments, as separate commits. If it's fine this way after approval I'll squash them into a single commit before merge

PeterSurda:
I have a few comments.

First, the pull request should be against "v0.6" branch rather than master, for historical reasons. I'll probably swap that when 0.7 is released, I don't want to change it in the middle of a cycle.

Second, I want to keep all git changes PGP signed, and this does not appear to have any signature, so I would have to sign it with my key. And as far as I understand, unless the pull request can be done in fast-forward, I have to sign the merge anyway even if it's signed already and I can't do that from the website.

I personally would have kept safeConfigGetBoolean, but it's fine this way as well.

mhitza:
@PeterSurda squashed commits, signed, and rebased on top of 0.6 for an easy fast forward merge

PeterSurda:
Well I still haven't figured out how to keep the signature when merging from the GitHub website :-(. The instructions say that if I use "rebase", the commit will be done in a fast-forward, but the signature is gone. I guess I'll have to continue doing it locally.

PeterSurda:
Also, now there are contribution instructions on the wiki: https://bitmessage.org/wiki/Contribute

For translations, I setup a webhook for the translation source so that there is no need for me to run `pylupdate4` manually. Now whenever there's a commit, my server will pull the changes, run `pylupdate4` and upload the resulting source into transifex.

I still have to run `lrelease` to put translated strings into the source, but I prefer to do that manually so that I can PGP sign git commits. But even though the trigger is manual, i'll just run a script.

mhitza:
It's an actual rebase since my commit id is different. I've checked [the docs](https://help.github.com/articles/about-pull-request-merges/) and for fast forward the only way would seem with "squash and merge", since it's only one commit that doesn't get "squashed" further and the merge is fast forward. Basically and edge case.

bmng-dev:
> Well I still haven't figured out how to keep the signature when merging from the GitHub website :-(

Sorry, I always forget about the signing policy and the restrictions that places on the GitHub Workflow.

PeterSurda:
I googled around and it looks like GitHub will always do the merge with an extra commit, no matter which option you choose. So I'll keep doing that locally until it's fixed. I'll try to overwrite this one as well, even though this is not a great idea on public repos, I don't like having an unsigned commit either.

PeterSurda:
Well it turns out that I committed 9590f83820aaef782a817f288c3a0ce7fc40a568 after you last modified this pull request, so I can't fast forward anyway :-(

bmng-dev:
> It looks like GitHub will always do the merge with an extra commit, no matter which option you choose

No, I've successfully used the merge options and the result is as per the GitHub documentation:
- Merge commit = number of commits in the pull request + the merge commit added to repository
- Squash and merge = 1 commit added to repository
- Rebase and merge = exactly the number of commits in pull request added to repository

I offered the knowledge of alternative merge options in an attempt to streamline the process of accepting pull requests by taking advantage of GitHub features and reducing the burden on both contributors and maintainers but I forgot the the signing policy makes this useless. Having to repeatedly request and educate contributors to generate throwaway gpg keys to sign their changes (and typos) and throw it all away to rebase onto the correct branch (rather than doing it for them) and then performing a CLI merge is apparently worth it considering commit level signing solves some sort of code integrity problem.

PR #915, opened 2016-11-14
Update pybitmessage.desktop with new GenericName

dmp1ce:
Easier to search for application without needing to know it is Python based.


PeterSurda:
Hello, thank you. Please follow the instructions here: https://bitmessage.org/wiki/Contribute

Mainly the request needs to be against the v0.6 branch, able to be merged as a fast-forward, and you need to PGP sign it. I'm not going to do any more commits today so if you do it now I can merge it right away.

dmp1ce:
See #916

PR #937, opened 2017-01-09
Improve OpenSSL library version detection

wfr:
PyBitmessage depends on OpenSSL 1.0 but some GNU/Linux systems have
multiple versions installed. Try to locate the correct version.


wfr:
This should fix the problem some users are having when OpenSSL 1.1 is installed in parallel to OpenSSL 1.0:
https://bitmessage.org/forum/index.php/topic,5280.msg11431.html

PR #938, opened 2017-01-10
Improve OpenSSL library version detection

wfr:
PyBitmessage depends on OpenSSL 1.0 but some GNU/Linux systems have
multiple versions installed. Try to locate the correct version.

https://bitmessage.org/forum/index.php/topic,5280.msg11431.html

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -13,7 +13,7 @@ installed on your system.
 Here's a list of dependencies needed for PyBitmessage
 - python2.7
 - python2-qt4 (python-qt4 on Debian/Ubuntu)
-- openssl
+- openssl 1.0
 - (Fedora & Redhat only) openssl-compat-bitcoin-libs 
 
 ##Running PyBitmessage
--- a/src/depends.py
+++ b/src/depends.py
@@ -99,10 +99,10 @@ def check_openssl():
     import re
     if re.match(r'linux|darwin|freebsd', sys.platform):
         try:
-            import ctypes.util
-            path = ctypes.util.find_library('ssl')
+            from pyelliptic.find_library_version import find_library_version
+            path = find_library_version('ssl', '1.0')
             if path not in paths:
-                paths.append(path)
+                paths.insert(0, path)
         except:
             pass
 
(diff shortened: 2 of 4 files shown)


PeterSurda:
I don't see anything obviously wrong with it, give me a couple days for a closer code review and testing.

bmng-dev:
@PeterSurda perhaps you should check the landscape.io integration because landscape does highlight an obvious issue which looks like a showstopper on Windows.

[Landscape.io Code Review for Pull Request #938](https://landscape.io/github/Bitmessage/PyBitmessage/pulls/938)

bmng-dev:
@wfr that was fast. I don't see any other issues preventing this being accepted as it is.

PeterSurda:
Wouldn't it be better to support both 1.0.x and 1.1.0? If `SSLeay_version` fails, try `OpenSSL_version` and analogously with `SSLeay`. Or did something else change in 1.1.0? I don't have a suitable VM available now, can you test it? `src/bitmessageqt/support.py` would also have to be changed analogously.

PeterSurda:
So I checked it with Debian testing. It looks like there are more changed in OpenSSL 1.1.0 that make pyelliptic incompatible:

- `ECDH_OpenSSL` -> `EC_KEY_OpenSSL`
- `ECDH_set_method` -> `EC_KEY_set_method`
- `EVP_CIPHER_CTX_cleanup` -> `EVP_CIPHER_CTX_reset`
- `EVP_MD_CTX_create` -> `EVP_MD_CTX_new`
- `EVP_MD_CTX_init` -> `EVP_MD_CTX_reset`
- `EVP_MD_CTX_destroy` -> `EVP_MD_CTX_free`

Just renaming these looks sufficient.  The problem is:

- `EVP_ecdsa` was deprecated and the relevant code needs to be rewritten to use the `'ecdsa-with-SHA1'` method manually.

Anyone would like to fix that?

PeterSurda:
So I tested it and other than the singing / signature verification it seems to work. If someone could fix that, that would be great.

PeterSurda:
Well, good news, `EVP_ecdsa` can be replaced with `EVP_sha1`. I tried it and it works both for signing and verification. So I'll provide my own patch instead and you can keep using OpenSSL 1.1.x.

PR #952, opened 2017-02-23
Handle exceptions from python without ipv6 support

g1itch:
Hello.

PyBitmessage shown many exceptions on my linux laptop because of python installed without ipv6 support. I tried to handle them all properly. The only question that remains is the log message "Peer host is None!" from line 53 in src/class_outgoingSynSender.py.

### DIFF
--- a/src/protocol.py
+++ b/src/protocol.py
@@ -100,7 +100,10 @@ def checkIPAddress(host, private=False):
             return False
         return hostStandardFormat
     else:
-        hostStandardFormat = socket.inet_ntop(socket.AF_INET6, host)
+        try:
+            hostStandardFormat = socket.inet_ntop(socket.AF_INET6, host)
+        except ValueError:
+            return False
         if hostStandardFormat == "":
             # This can happen on Windows systems which are not 64-bit compatible 
             # so let us drop the IPv6 address. 


PeterSurda:
I'm now in feature freeze mode for 0.6.2 so I can't promise this will make it into 0.6.2 because I need to test it and I don't know if I have time. Please follow the contribution guidelines: https://bitmessage.org/wiki/Contribute, you're missing a PGP signature that is associated with your github account. Without the signature, I'll have to resign it with mine and then I'll be the committer and you won't receive a tip.

src/class_singleListener.py
```diff
@@ -86,12 +86,13 @@ def run(self):
...
             sock = self._createListenSocket(socket.AF_INET6)
-        except socket.error, e:
+        except socket.error as e:
             if (isinstance(e.args, tuple) and
                 e.args[0] in (errno.EAFNOSUPPORT,
                               errno.EPFNOSUPPORT,
                               errno.EADDRNOTAVAIL,
-                              errno.ENOPROTOOPT)):
+                              errno.ENOPROTOOPT,
+                              'getsockaddrarg: bad family')):
```
PeterSurda: Isn't this covered by errno.EAFNOSUPPORT?

PeterSurda:
Also, please tell me which linux distro has ipv6 disabled so I can test it.

g1itch:
In gentoo you can disable ipv6 by portage use-flag (USE="-ipv6").

src/class_outgoingSynSender.py
```diff
@@ -47,14 +47,19 @@ def _getPeer(self):
                         self.stop.wait(1)
                         continue
                     priority = (183600 - (time.time() - knownnodes.knownNodes[self.streamNumber][peer])) / 183600 # 2 days and 3 hours
+                try:
+                    peer_is_onion = peer.host.find(".onion") != -1
+                except AttributeError:
+                    logger.error("Peer host is %r!" % peer.host)
```
PeterSurda: Landscape.io complains here, it should be just a "," rather than "%".

src/class_singleListener.py
g1itch: This article https://docs.python.org/2/library/socket.html#socket.error says socket.error may return string instead of errno and i've seen it.

PeterSurda:
I don't think I'll have time to test it before 0.6.2 because I don't have a gentoo environment and no other VM without IPv6 support. Just hold on a week or two.

g1itch:
OK

PeterSurda:
I actually had to do some other fixes which may have fixed this one too (apart from the string). I haven't tested it on a IPv4-only linux, but I tested it on OpenBSD which cannot open a socket for both IPv4 and IPv6 and it worked fine, it fell back to IPv4-only. Can you retry?

g1itch:
Nothing changed for me. With your recent code pybitmessage still works, but if I run it in terminal I can see the following exceptions:
```
Exception in thread singleListener:
  File "/usr/lib64/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib64/python2.7/site-packages/pybitmessage/class_singleListener.py", line 88, in run
    sock = self._createListenSocket(socket.AF_INET6)
  File "/usr/lib64/python2.7/site-packages/pybitmessage/class_singleListener.py", line 46, in _createListenSocket
    sock.bind((HOST, PORT))
  File "/usr/lib64/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
error: getsockaddrarg: bad family
```
```
CRITICAL - Critical error in a receiveDataThread: 
Trac
RITICAL - Critical error in a receiveDataThread: 
  File "/usr/lib64/python2.7/site-packages/pybitmessage/class_receiveDataThread.py", line 220, in processData
    self.recaddr(payload)
  File "/usr/lib64/python2.7/site-packages/pybitmessage/class_receiveDataThread.py", line 612, in recaddr
    hostStandardFormat = self._checkIPAddress(fullHost)
  File "/usr/lib64/python2.7/site-packages/pybitmessage/class_receiveDataThread.py", line 552, in _checkIPAddress
    hostStandardFormat = socket.inet_ntop(socket.AF_INET6, host)
ValueError: unknown address family 10
```

PeterSurda:
Thanks for the feedback, I'll look at it after 0.6.2 as I have no way to test your patch on a IPv4 only system at the moment.

g1itch:
One exception still exists. Another distro which can be configured without IPv6 support is openwrt.

PeterSurda:
Well it looks pretty straightforward.

Issue #819, opened 2015-10-29
idea: QR Code export of identities


ghost:
QR Code export for identities would be nice to have for business cards, websites, and once we have an functional (whatever-floats-your-boat system) mobile platform client.

krosos:
on related note: is there a mime/link type for bitmessage?
something like bitmessage:BM-longhash?label=yourchanname-or-identityname
?

PR #961, opened 2017-03-01
QR-code dialog

g1itch:
To open QR-code dialog right click one of you bm-addresses and select "Show QR-code".

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -191,6 +191,9 @@ if __name__ == "__main__":
             # TODO: add keywords
             #keywords='',
             install_requires=['msgpack-python'],
+            extras_require={
+                'qrcode': ['qrcode']
+            },
             classifiers=[
                 "License :: OSI Approved :: MIT License"
                 "Operating System :: OS Independent",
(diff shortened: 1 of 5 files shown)


PeterSurda:
Thank you, I like this. Maybe this can be made optional as this requires a new dependency, so that the context menu item would only show up if the qrcode is available?

g1itch:
Yes, I will made changes.

PeterSurda:
If you could also add qrcode as an optional prerequisite to setup.py, that would be great.

g1itch:
With the above changes once `qrcode` installed QR-code menu item will appear. It may sound weird, but I think all optional functionality with additional dependencies may be rewritten as such plugins with extras (opencl support with `pyopencl` and `numpy`, tray menu with `pygobject`, etc).

BTW there is an issue #819

PeterSurda:
Thanks.

PeterSurda:
please rebase, I can't fast-forward merge.

PeterSurda:
Looks ok and I'm happy about the plugin architecture, I wanted to do something like this. But it looks like the plugins have to be installed first (e.g. using setuptools) before it can load them. Can you look at that or should I? Also I'm not sure if it works in frozen mode, have you tried it? I've been having problems with dynamically loading modules in frozen mode (both on Windows and OSX).

PeterSurda:
How about I merge this first and then the other issues are solved separately? If the plugin just doesn't activate in some cases, that's acceptable (as opposed to if it was crashing).

PR #963, opened 2017-03-01
Changed copyright year to 2017

sandakersmann:
Changed copyright year to 2017


PeterSurda:
Thanks. Please read the contribution guidelines: https://bitmessage.org/wiki/Contribute. Your commit is not PGP signed, and the translations are done via transifex.

However, I also think that the year should be made dynamic rather than bumped every now and then.

PR #966, opened 2017-03-05
Missed translations: namecoin and welcome message

g1itch:
There is also a question about translations of MB, kB, etc. in [bitmessageqt/networkstatus.py](../tree/v0.6/src/bitmessageqt/networkstatus.py#L37) ...


PeterSurda:
good catch

PR #981, opened 2017-03-16
More fixes for l10n

g1itch:
When I was looking for the reason why pylupdate4 not includes the string "Show QR-code" into ts-files, I noticed that almost any widget module defines the `_translate()` function and calls it with many default values for the positional arguments. Fix me, I think `_translate` should be imported from `tr` where it's already defined for all cases and with the proper defaults. Here I edited only the modules `__init__` and `networkstatus` because it will not probably be generated from ui-files.

The disadvantage of this changeset is that pylupdate4 will mark all strings in that changed `_translate()` calls as "obsolete" and translators on transifex will be forced to copy suggestion into translation for each of it.

### DIFF
--- a/src/bitmessageqt/__init__.py
+++ b/src/bitmessageqt/__init__.py
@@ -1460,19 +1460,50 @@ class MyForm(settingsmixin.SMainWindow):
                 # may manage your keys by editing the keys.dat file stored in
                 # the same directory as this program. It is important that you
                 # back up this file.', QMessageBox.Ok)
-                reply = QtGui.QMessageBox.information(self, 'keys.dat?', _translate(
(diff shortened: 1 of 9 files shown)


g1itch:
Hmm, this not works at all with keyword args. New strings should be marked "unfinished", not "obsolete". Investigating...

PeterSurda:
Thanks for doing this, let me know when you're done so that I can review it.

g1itch:
I think it's finished already (for this two modules). Not squashing yet to be able to drop undesirable commits if any (what about 'synchronisation' for example?).

g1itch:
Rebased and squashed a bit. There is a rebase artifacts between commits but the final patch should be correct. Of course I'll squash further all acceptable commits before you merge this pr.

PeterSurda:
Thanks I'll look at it.

g1itch:
This is work in progress currently. Please don't merge without prior notice. The commit 
9267416 has a rebase artifacts and needs to be edited in next rebase.

PR #985, opened 2017-03-25
Move OS and package detection care to setuptools install command

g1itch:
Hello!

I'm testing PyBitmessage on Ubuntu. It's annoying:

1. I run `python setup.py sdist`
    and **press Enter**
2. run `py2dsc-deb dist/pybitmessage-0.6.2.tar.gz`
    and **press Enter** 6 times until I'll get the deb

If you move those OS and package detection care to setuptools install command in that case I would press Enter only once. It seems to me it will be logical, because all that detection code is for the end user, which will need the dependencies installed only then installing PyBitmessage.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,8 +4,10 @@ import os
 import sys
 try:
     from setuptools import setup, Extension
+    from setuptools.command.install import install
     haveSetuptools = True
 except ImportError:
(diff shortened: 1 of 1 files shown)


PeterSurda:
Looks good, thank you.

PR #988, opened 2017-03-28
Return line is missing

SafetyKeeper:
Even if sanity check fails, data processing continues.

### DIFF
--- a/src/shared.py
+++ b/src/shared.py
@@ -438,6 +438,7 @@ def _checkAndShareGetpubkeyWithPeers(data):
         return
     if len(data) > 200:
         logger.info('getpubkey is abnormally long. Sanity check failed. Ignoring object.')
+        return
     embeddedTime, = unpack('>Q', data[8:16])
     readPosition = 20  # bypass the nonce, time, and object type
     requestedAddressVersionNumber, addressVersionLength = decodeVarint(


jakobvarmose:
This is a bug, but I think it is better to remove the check altogether instead of suddenly enforcing it. Changing what objects are shared with other peers is probably not a good idea.

PeterSurda:
@SafetyKeeper thank you, I'll accept it but you have follow https://bitmessage.org/wiki/Contribute (the commit doesn't have a PGP signature).

@jakobvarmose While I accept your argument from caution, getpubkey has almost no practical effect anyway. If your pubkey is expired, PyBitmessage will ll send a new one out anyway very soon, and if it isn't expired, the getpubkey will be ignored. I only haven't decided on decommissioning getpubkey altogether is that maybe it will have some effect once streams and/or onion routing is implemented.

jakobvarmose:
It doesn't matter if it is a getpubkey or another type of object. The main problem is that you can put it into the ackdata of a msg object, and then use it to identify who received your msg object.

PeterSurda:
@jakobvarmose I see.

PeterSurda:
Perhaps this check should be moved to class_objectProcessor.py/processgetpubkey instead then.

PeterSurda:
@jakobvarmose Currently, PyBitmessage only looks for ACKs in msg objects. However, you are correct that it shouldn't be like this.

PeterSurda:
Fixed both (sanity check and ackdata being object type independent).

PR #995, opened 2017-05-07
Update setup.py

Lvl4Sword:
- Fixes two KeyErrors on Debian Jessie
- gets rid of two unnecessary list comprehensions for readability
- [] is also considered False in Python, so simply doing a "if variable:" will suffice
- Spacing for readability
- ' pybitmessage' to 'PyBitmessage'
- No point in suggesting Windows users use the package manager :smile: 
- Got rid of unnecessary () around print since this is Python2
- Two new lines between every function, aligning with Pep8


PeterSurda:
Thank you, please consult contribution guidelines: https://bitmessage.org/wiki/Contribute

PR #1006, opened 2017-06-03
Change api.py

Lvl4Sword:
-    ListSubscriptions isn't as messed up as it was json-wise, and is more readable. ( HandleListAddresses still needs fixed with json as well as others. )
-    base64 is imported and used in _decode because .decode() appends a newline character to each encoded string.
-    _decode now doesn't use .decode() and instead uses unhexlify(text) and base64.b64decode(text).
-    .encode() converted to base64.b64encode()


Lvl4Sword:
THERE. Now I know the ins and outs of signing. Only took me how many times?

PR #1014, opened 2017-06-12
New packages not listed in setup.py

g1itch:
Hello!

I noticed you added some new packages in recent code. Particularly `fallback.umsgpack` which should make dependency on python-msgpack optional. I made several changes to make it possible to install pybitmessage with `umsgpack` or without any msgpack implementation (to use `fallback` package).

First commit is sufficient to fix setup failure in case you don't like other changes.


PeterSurda:
looks ok

PR #1047, opened 2017-09-11
Notification plugins

g1itch:
I moved all notification code (including sound) to plugins.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -285,7 +285,8 @@ if __name__ == "__main__":
                     '[notify2]'
                 ],
                 'notification.sound': [
-                    'fallback = pybitmessage.plugins.sound_playfile [sound]'
+                    'file.fallback = pybitmessage.plugins.sound_playfile'
+                    '[sound]'
                 ],
                 # 'console_scripts': [
                 #        'pybitmessage = pybitmessage.bitmessagemain:main'
--- a/setup.py
+++ b/setup.py
@@ -254,7 +254,9 @@ if __name__ == "__main__":
             install_requires=installRequires,
             extras_require={
                 'qrcode': ['qrcode'],
-                'pyopencl': ['pyopencl']
+                'pyopencl': ['pyopencl'],
+                'notify2': ['pygobject', 'notify2'],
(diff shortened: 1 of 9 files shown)


PeterSurda:
Thank you, it will take me a while to review it but in general I think it's fine.

PeterSurda:
What OSes did you test it on? Have you tried frozen mode (PyInstaller on Windows, Py2App on OSX)? I had problems with dynamic loading in frozen mode and wasn't able to resolve it cleanly.

g1itch:
I tested on my Calculate Linux and Ubuntu 16.04 (`indicator_libmessaging`). I have no OSX and unsure which of this plugins are appropriate for it (it seems pygobject or gir is a linux stuff). Unfortunately I still could not build PyBitmessage for Windows (to be precise I cannot build Qt) but here is only one plugin for Windows - it's `sound_playfile`.

PeterSurda:
I can test OSX then. Regarding windows, you don't really have to build qt, you can just install the official binary. I recommend building qt if you're creating a file targeted for wider distribution and want to make sure you have the latest library versions, like openssl or patched MSVC.

setup.py
```diff
@@ -177,6 +179,14 @@ def compilerToPackages():
 
 class InstallCmd(install):
     def run(self):
+        # prepare icons directories
+        os.makedirs('desktop/icons/scalable')
```
PeterSurda: This doesn't check if the directory already exists. Move the makedirs/copyfile to the end, just before install.run is called.

PeterSurda:
In general I'm happy about it but I still have to do some tests, like frozen mode, BSD-like systems. Please rebase so that fast-forward merge is possible. The problems I found are only minor and can be fixed afterwards.

setup.py
g1itch: Quote from `help(os.makedirs)`:
```
... any intermediate path segment (not
    just the rightmost) will be created if it does not exist.
```
Yes, I'll move this chunk to the end of `run()` method.

PeterSurda:
Actually, there probably isn't any need to rebase. I'll check it out tomorrow.

Issue #1035, opened 2017-08-02
CLI Command Line Arguments


Lvl4Sword:
Daemon and Portable Mode should be command line arguments. This allows other programs to use the CLI much easier.

sp2004c:
true, one can set BITMESSAGE_HOME , does not even have to be an extra line in a bash script. Same effect.

PR #1055, opened 2017-09-23
use getopt parser for command-line arguments

f97ada87:
getopt is the standard UNIX command-line parsing function

This PR implements a getopt-based parser for bitmessagemain, supporting both short (-h) and long (--help) command-line options; a small help text is also included. 
The -c (curses) option is ported to the new parser and there is a new option, -d, for selecting daemon mode from command line (which partially solves issue #1035)

### DIFF
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -29,6 +29,7 @@ from struct import pack
 from subprocess import call
 from time import sleep
 from random import randint
+import getopt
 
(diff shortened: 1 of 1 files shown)


f97ada87:
tested with Python 2.7 on Linux

PeterSurda:
I planned on adding options like these, with argparse, but after reviewing the documentation, it looks like getopt is also an acceptable approach.

PR #1058, opened 2017-09-25
fix truncation of received ackdata in objectProcessor

f97ada87:
According to the Bitmessage protocol specification, the message acknowledgement data can be any valid protocol object, giving as example "like another msg message". 
When the pyBitmessage objectProcessor checks a received object against the list of watched acks (in function checkackdata() ) it does so by comparing the last 32 bytes of the object (data[-32:]) to the full ack data (as previously stored in the keys of shared.ackdataForWhichImWatching ). 
This causes valid ack objects with a size other than 32 bytes to fail the comparison and be ignored. 

This is a dormant problem currently as the ack data is hard-coded to OpenSSL.rand(32) at several points in the code, effectively limiting its size to 32 bytes and enabling the comparison to work; however, changing the ack size to another value would cause the ack detection mechanism to stop working. 

For a simple test:
- change all instances of OpenSSL.rand(32) to OpenSSL.rand(320) in bitmessageqt/__ init __.py 
- (re)start the QT client
- send a message to an external address 
- watch the "sent" folder
If the message never progresses from "waiting for acknowledgement" state, then the problem is present. 

This PR updates the checkackdata() function to work correctly for ack objects of any size.

### DIFF
--- a/src/class_objectProcessor.py
+++ b/src/class_objectProcessor.py
@@ -97,14 +97,24 @@ class objectProcessor(threading.Thread):
         # Let's check whether this is a message acknowledgement bound for us.
         if len(data) < 32:
             return
-        if data[-32:] in shared.ackdataForWhichImWatching:
+        readPosition = 20  # bypass the nonce, time, and object type
+        # chomp version number
+        versionNumber, varIntLength = decodeVarint(
+            data[readPosition:readPosition + 10])
+        readPosition += varIntLength
+        # chomp stream number
+        streamNumber, varIntLength = decodeVarint(
+            data[readPosition:readPosition + 10])
+        readPosition += varIntLength
+
+        if data[readPosition:] in shared.ackdataForWhichImWatching:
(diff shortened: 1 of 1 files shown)


f97ada87:
Tested with Python 2.7 on Linux using test method above.

PeterSurda:
Thank you for your hard work. Yes, your analysis is correct. The fix appears to be correct too. In the future I would like to avoid duplicate code (object structure parsing is already in bmproto and bmobject) but at the moment the code can't be reused so it's ok. Please rebase (I made a commit between you created your commit and the pull request) so that I can fast-forward-merge.

PeterSurda:
As a side note, there were opinions in the past that os.urandom should be preferred to OpenSSL.rand, so if you want cryptographically secure random data in the future, use helper_random.randomBytes.

f97ada87:
Noted all points, thanks. I have pulled your changes and rebased.

PR #1060, opened 2017-09-28
Properly handle SIGTERM if daemon ran with -d 

g1itch:
There is a minor problem: if no `daemon` setting in `keys.dat` pybitmessage ignores `SIGTERM`. I suggest to check `singleistance.daemon` variable instead of config to determine daemon status.

### DIFF
--- a/src/class_singleCleaner.py
+++ b/src/class_singleCleaner.py
@@ -66,7 +66,7 @@ class singleCleaner(threading.Thread, StoppableThread):
             # If we are running as a daemon then we are going to fill up the UI
             # queue which will never be handled by a UI. We should clear it to
             # save memory.
-            if BMConfigParser().safeGetBoolean('bitmessagesettings', 'daemon'):
+            if shared.thisapp.daemon:
                 queues.UISignalQueue.queue.clear()
             if timeWeLastClearedInventoryAndPubkeysTables < int(time.time()) - 7380:
                 timeWeLastClearedInventoryAndPubkeysTables = int(time.time())
--- a/src/helper_generic.py
+++ b/src/helper_generic.py
@@ -6,7 +6,7 @@ from multiprocessing import current_process
 from threading import current_thread, enumerate
 import traceback
 
-from bmconfigparser import BMConfigParser
+import shared
 from debug import logger
 import queues
 import shutdown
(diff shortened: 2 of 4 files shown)


PeterSurda:
Looks ok but I have some other pending stuff to do first. Afterwards I'll ask you to rebase. You can also fix other places where daemon is checked since you're on it.(shutdown, tr, singlecleaner).

f97ada87:
Good catch, and sorry for the regression that I've caused. 
I think the "flake" commits belong in a separate PR as they are not relevant to this bugfix (SIGTERM handling). Other than that, it looks good. 

Tested successfully on Linux and an ACK from me (for commit 2a94647 only).

PeterSurda:
Please separate flake8 commits to a different pull request.

g1itch:
There is also `shared.daemon` introduced in c89d86a: I'm unsure maybe it should be used instead of `shared.thisapp.daemon`?

f97ada87:
I wondered about this too. There is currently a bunch of runtime variables scattered between `shared`, `state` and `thisapp` which seem to be the result of several incomplete migrations. 
Based on my reading of the current code, `shared.daemon` is legacy and should be retired, and `shared.thisapp.daemon` is a sore thumb; I think the long-term solution is to migrate both of these to `state.daemon` - as a separate PR. 
For the purpose of this PR though, I think `shared.daemon` can be left as-is. It works as intended. 

FTR `shared.daemon` was not introduced in c89d86a, it was just moved a few lines further down. I avoided changing too much of the status quo to keep the diff small, but it seems I did a pretty poor job at that.

f97ada87:
ad0119f tested OK on Linux with daemon=false and --daemon switch.

g1itch:
Actually I have a problem with desktop app in module `tr` with this code. Investigating...

g1itch:
Is something wrong with this PR?

PeterSurda:
Please rebase, I'm ready to merge, looks ok to me.

PeterSurda:
If you want, you can also get rid of `shared.daemon`, or I can do it after merging this. This includes removing one line in `bitmessagemain.py` and replacing `shared.daemon` with `shared.thisapp.daemon` in `class_singleCleaner.py`.

PeterSurda:
I think I figured out how to merge without rebasing in a way that doesn't distrupt the workflow, tip4commit and gpg signatures. Should make it easier to merge in the future.

PeterSurda:
@g1itch let me know if you'll get rid of shared.daemon (in a separate PR) or want me to do that.

PR #1061, opened 2017-09-30
Stealth ACK objects

f97ada87:
By default, message ACK objects are generated as 32-byte random strings with a message header attached. Because of this, they are easily identified on the network (they're the only msg objects of size 54). 
This PR implements some new types of ACK objects that blend in more realistically with the network traffic, and enables selection via a new configuration file option. Three stealth levels are defined currently:
0 - the default format described above
1 - a getpubkey request for a non-existent key
2 - a standard formatted message encrypted to a non-existent key
To support objects of types other than 2 (msg), the internal format of the ackdata has been changed to also include the type header (objtype/version/stream). The startup routine has been updated to convert the legacy (headerless) format to the new format automatically, to continue tracking of any unacknowledged messages after upgrade. 

Any comments and feedback are welcome.


f97ada87:
SORRY - Wrong branch, please ignore.

PR #1065, opened 2017-10-07
Check IP before adding to knownnodes

g1itch:
I rewrited `helper_bootstrap` a bit to ensure that only valid IP addresses will be added to knownnodes.

The problem is in last elements of this list:

```python
import socket
socket.getaddrinfo('bootstrap8080.bitmessage.org', 80)
```

You don't see it because of `except:` without exception type in [src/helper_bootstrap.py#L52](../blob/v0.6/src/helper_bootstrap.py#L52)

### DIFF
--- a/src/helper_bootstrap.py
+++ b/src/helper_bootstrap.py
@@ -9,6 +9,7 @@ import knownnodes
 import socks
 import state
 
+
 def addKnownNode(stream, peer, lastseen=None, self=False):
     if lastseen is None:
         lastseen = time.time()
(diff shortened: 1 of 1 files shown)


PeterSurda:
I want to rewrite it as well to move the lookups into asyncore when using SOCKS. Then I can get rid of SocksiPi. The asyncore SOCKS resolving hasn't been tested but it probably works, just there isn't a way to communicate with it. Maybe through two queues like sql thread? Perhaps I also need to add some error handling.

Also, the onion bootstrap IP should be added even if not on tor, in order to allow bootstrapping when on a transparent proxy.

PR #1067, opened 2017-10-11
send ack even if message encoding is zero

f97ada87:
In the definition of BM message objects, a value of zero in the "message encoding" field indicates that the message "may be ignored". In practice, this means that, after a successful decryption, the message body is discarded without being stored on disk or displayed in the UI. However, the other metadata processing (pubkey extraction etc) is still performed. 
In the BM protocol, the message ACK object is broadcast on the network when the message has been successfully **decrypted** (subject to some flags and conditions) - **not** when the message has been displayed to, or read by, the recipient. 
Therefore, I suggest that the ACK object should also be broadcast in the case of messages with zero-type encoding that have been successfully decrypted, even if (by design) they are not displayed to the recipient. 
This PR removes the restriction on ACK sending for zero-type encoded messages.

### DIFF
--- a/src/class_objectProcessor.py
+++ b/src/class_objectProcessor.py
@@ -591,7 +591,6 @@ class objectProcessor(threading.Thread):
         # Don't send ACK if invalid, blacklisted senders, invisible messages, disabled or chan
         if self.ackDataHasAValidHeader(ackData) and \
             not blockMessage and \
-            messageEncodingType != 0 and \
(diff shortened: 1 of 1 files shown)


PeterSurda:
This was disabled as it can be misused as a vector for stealth deanonymisation. The victim wouldn't see anything visually during the attack. Since anonymity is the top priority of Bitmessage, this takes precedence over specification consistency. Although perhaps this should be documented.

PR #1073, opened 2017-10-19
Use unicode instead of str for MessagingMenu

g1itch:
It's my own error ): I didn't notice it because used English locale in Ubuntu when tested this code. `UnicodeEncodeError` is raised here.


PeterSurda:
Please rebase, ready to merge.

PR #1075, opened 2017-10-19
Slightly redrawn the scalable icon

g1itch:
This version is still not perfect but without gaps and extra polygons.


PeterSurda:
I like the improvement

Issue #1044, opened 2017-09-05
Inconsistent message read status on Chans tab


g1itch:
Hello.

When I read the message received from chan in All accounts -> inbox on tab Messages and then switch to Chans tab I see that message still drawn with bold font, like unread.

g1itch:
It's weird: method `propagateUnreadCount` not uses any of it's four keyword parameters (`address=None, folder="inbox", widget=None, type=1`). But in method `tableWidgetInboxItemClicked` it called like this:
```python
            if propagate:
                self.propagateUnreadCount(
                    tableWidget.item(
                        currentRow,
                        1 if tableWidget.item(currentRow, 1).type ==
                        AccountMixin.SUBSCRIPTION else 0
                    ).data(Qt.UserRole),
                    folder, self.getCurrentTreeWidget(), -1
                )
```

PeterSurda:
As far as I remember, the arguments were used in the past for performance optimisation, but it lead to problems and the performance difference was very small. It probably needs to be looked at in a bit more detail.

g1itch:
Here's the method [`tableWidgetInboxItemClicked`](../blob/master/src/bitmessageqt/__init__.py#L3825) how it looks for me:
```python
    def tableWidgetInboxItemClicked(self):
        folder = self.getCurrentFolder()
        messageTextedit = self.getCurrentMessageTextedit()
        if not messageTextedit:
            return

        msgid = self.getCurrentMessageId()
        if msgid:
            queryreturn = sqlQuery(
                '''SELECT message FROM %s WHERE msgid=?''' % (
                    'sent' if folder == 'sent' else 'inbox'
                ), msgid
            )

        try:
            message = queryreturn[-1][0]
        except NameError:
            message = ""
        except IndexError:
            message = "Error occurred: could not load message from disk."
        else:
            tableWidget = self.getCurrentMessagelist()
            currentRow = tableWidget.currentRow()
            # refresh
            if tableWidget.item(currentRow, 0).unread is True:
                font = QtGui.QFont()
                font.setBold(False)
                tableWidget.item(currentRow, 0).setUnread(False)
                tableWidget.item(currentRow, 1).setUnread(False)
                tableWidget.item(currentRow, 2).setUnread(False)
                tableWidget.item(currentRow, 3).setFont(font)
            # propagate
            if folder != 'sent' and sqlExecute(
                '''UPDATE inbox SET read=1 WHERE msgid=? AND read=0''',
                msgid
[...]

PR #1084, opened 2017-12-04
translate: could not load message from disk

sigoa:
message = "Error occurred: could not load message from disk." 
TODO: _translate( this ! 
it happens frequently. my suggestion:

"Error occurred: could not load message from disk. -- **This is not a (received) BitMessage, but an error message due to a read storage problem."**


g1itch:
The method `tableWidgetInboxItemClicked` needs complete rewrite as for me. Also It potentially contains  message read status propagation bug, #1044.

g1itch:
If it happens frequently you could open an issue.

PR #1093, opened 2018-01-22
 to demo utf8, emoji logging without exception

sigoa:
to demo utf8, emoji logging without exception

as requested by Peter via BM recently

### DIFF
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -407,12 +407,16 @@ All parameters are optional.
         return {'address':address,'port':port}
 
     
-def emojiLogg():
+def LoggWithUnicode():
     from debug import logger
-    reload(sys)  
-    sys.setdefaultencoding(  'utf8')
-    logger.warning(str(     u'exc:äüöß   🤷 🤷  without exception  😃                         '.encode("UTF-8")   )          )
-    logger.warning(          ' äüß      🤷 🤷 🤷   -- just a test emoji warning to demo this  ')
+    #reload(sys)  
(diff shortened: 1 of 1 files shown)


src/bitmessagemain.py
```diff
@@ -405,12 +406,22 @@ def getApiAddress(self):
         port = BMConfigParser().getint('bitmessagesettings', 'apiport')
         return {'address':address,'port':port}
 
+    
+def emojiLogg():
```
PeterSurda: Please rename to something else than "emoji", e.g. something with utf8, unicode or encoding.

src/bitmessagemain.py
```diff
@@ -405,12 +406,22 @@ def getApiAddress(self):
...
+    from debug import logger
+    reload(sys)  
+    sys.setdefaultencoding(  'utf8')
+    logger.warning(str(     u'exc:äüöß   🤷 🤷  without exception  😃                         '.encode("UTF-8")   )          )
+    logger.warning(          ' äüß      🤷 🤷 🤷   -- just a test emoji warning to demo this  ')
```
PeterSurda: Please get rid of the logger.warning or replace it with something more neutral, like "logger encoding reset", and it should be .info or .debug, not .warning.

PeterSurda:
The commit is against the master branch, please rebase against v0.6, other than that it looks fine.

PeterSurda:
I tested it and it works if your locale is utf-8. If it isn't and the error message has non-ascii characters, it still is broken. However I figured out how to fix both.

PR #1094, opened 2018-01-25
utf-8 logging globally

sigoa:
against v.0.6


PeterSurda:
If this works then I'll be impressed.

PeterSurda:
This doesn't work.

Issue #1081, opened 2017-11-28
"Mark all messages as read" causes critical error


g1itch:
Hello!

I have 346 messages in `[chan] bitmessage` shown as unread. When I select "Mark all messages as read" context menu this causes PyBitmessage crash with many unicode characters dumped into console and finally the following messages:

> Here is the actual error message thrown by the sqlThread: too many SQL variables
CRITICAL - This program shall now abruptly exit!

So I probably have [SQLITE_MAX_VARIABLE_NUMBER](https://www.sqlite.org/limits.html#max_variable_number) < 999 in my system, but it seams there is no easy way to check the exact value.

[The source](../blob/v0.6/src/bitmessageqt/__init__.py#L2536)

PeterSurda:
Thanks. I thought I fixed that. I think I should create a new function over sqlExecute which would handle this limit.

reelsense:
What version are you using @g1itch ?

g1itch:
Latest PyBitmessage commit, Python 2.7.12, sqlite 3.19.3

g1itch:
Hmm, I don't understand this:
`if i % 999 == 999:`

It seems to be always False

PeterSurda:
good catch.

PeterSurda:
Fixed in 80fdb08f0369a86e4e65b740c0457b467fdff416

PR #1096, opened 2018-01-25
Quick fix for #1081

g1itch:
With this oneliner "Mark all messages as read" works without error.

### DIFF
--- a/src/bitmessageqt/__init__.py
+++ b/src/bitmessageqt/__init__.py
@@ -2542,7 +2542,7 @@ class MyForm(settingsmixin.SMainWindow):
             tableWidget.item(i, 2).setUnread(False)
             tableWidget.item(i, 3).setFont(font)
             # sqlite default limit, unfortunately getting/setting isn't exposed to python
-            if i % 999 == 999: 
+            if i % 999 == 0:
                 markread += partialUpdate(self.getCurrentFolder(), msgids)
                 msgids = []
 


PeterSurda:
perhaps it should be 998?

PeterSurda:
or even better, can you check if sqlExecuteChunked in helper_sql works correctly and use that instead?

g1itch:
If I understand correctly the objective is to take 999 items at a time, in this case the remainder from the division should be 0.

g1itch:
OK, I'll try.

PeterSurda:
i think if it's 0, it will execute the first chunk with just one item.

g1itch:
I don't understand how `sqlExecuteChunked()` can work without modification of `sqlStatement` because it contains `idCount` placeholders.
```python
        markread = sqlExecuteChunked(
            "UPDATE %s SET read = 1 WHERE ackdata IN(%s) AND read=0" % (
                'sent' if self.getCurrentFolder() == 'sent' else 'inbox',
                ','.join('?' * idCount)
            ),
            idCount, *msgids
        )
```

PeterSurda:
You're right, I missed that one. It can probably be fixed by having a placeholder (e.g. `{0}`) in the original statement, and replacing it the way you just did but inside each of `sqlExecuteChunked` loops.

g1itch:
It probably works, at least for me.

g1itch:
I'll drop 0097ea6 If you test and approve this changes.

PeterSurda:
Looks ok, I'll test it tomorrow and then merge.

PR #1097, opened 2018-01-26
add ,

sp2004c:


BitpopCoin:
I'm a grammar Nazi but I don't think that comma is needed.

PR #1098, opened 2018-01-30
The network switching menu

g1itch:
Hello

Here is the "File -> Go offline/Go online" menu entry switching `dontconnect` variable and "Work offline" option for `connectDialog` doing nothing. Also I had to add `.set_state("close")` and `.handle_close()` calls for each existing connection to disconnect it properly and update "Network Status" tab.


sigoa:
Was this not called a "non-bug" ?

"dontconnect" never worked for me, anyway.

g1itch:
@sigoa, this may be called a wish or feature. @PeterSurda  wrote he want such menu in `[chan] bitmessage`.

`dontconnect` works as planned: prevents connection on first run. This PR extends it's usage making it possible to manually switch the network on/off. Support for it is already in the [`network.connectionpool`](../blob/v0.6/src/network/connectionpool.py#L157) module except for actually closing connections.

PeterSurda:
Thank you, works fine :+1:

PR #1099, opened 2018-01-30
Patch 1 cli v

sigoa:
the cli app slightly updated, so e.g. it could be GUI-fied


PeterSurda:
I already merged your previous patch in v0.6. It will be merged into master later.

PR #1103, opened 2018-01-31
bitmessage.pro needs update

g1itch:
Hello!

I forgot to update the `bitmessage.pro` and many translation strings are disappeared from transifex. I hope most of them come back after this update at least as suggestions. Just checked .ts generated by `pylupdate4` after this update and I see not many changes.


PeterSurda:
I'm not sure they'll come back automatically but I think that can be fixed manually. I'll look at it in a couple of days.

PR #1110, opened 2018-02-02
Minor changes to dialogs

g1itch:
Hello.
Based on discussion in `[chan] bitmessage`
I hope this makes simple dialog better. Close if no.


PeterSurda:
It is better.

PR #1113, opened 2018-02-05
Run autopep8 on the whole python codebase

shak-mar:
> - try to explain what the code is about
> - try to follow [PEP0008](https://www.python.org/dev/peps/pep-0008/)

I thought I'd make it easier for everyone to follow PEP8;  it's easier to follow a code style when the whole code base consistently uses it.

There's still quite some violations, but it's a bit better at least.

> - make the pull request against the ["v0.6" branch](https://github.com/Bitmessage/PyBitmessage/tree/v0.6)
> - it should be possible to do a fast-forward merge of the pull requests

Tried to do so.  If the v0.6 branch updates faster than I can update this PR (won't invest much effort), this PR can easily be reproduced by the command given in the commit message.

> - PGP-sign the commits included in the pull request

I didn't PGP sign my commits since noone around here knows my key, so it'd avail close to nothing.


shak-mar:
Another good reason to reproduce this PR using th command given in the commit message is that it's a lot of changes that are hard to review by hand.  Since I guess you don't trust strangers that much, it's better for security to simply run the command yourself.

Oh, and Jesus loves you btw :)  That's even why I'm contributing ;)

PeterSurda:
I didn't know about autopep, so thank you. I'll run this myself and commit it like that.

In the meantime I found out how to do merges the way I want them to be, so it's not necessary for me to do a fast-forward merge anymore. Regarding PGP, its purpose isn't to verify your identity, but to maintain integrity of the repository. I don't need to know who you are, but I want to make sure that the the commit authenticity can be verified in the future.

shak-mar:
> Regarding PGP, its purpose isn't to verify your identity, but to maintain integrity of the repository. I don't need to know who you are, but I want to make sure that the the commit authenticity can be verified in the future.

Hmm, I'm not sure I understand...  Does this mean the PGP key would build trust along the way, i.e. “commit authenticity can be verified in the future” would mean you can verify future PRs are really from the same contributor?  The integrity of the repository is guaranteed (given one trusts sha1;  since the commit signing only signs the hashes, that trust is implicitly granted by talking about PGP signing as something worthwhile) via the sha1 hashing, but the authenticity of the latest commit hash needs to be verified out-of-band or via PGP signing;  so do you mean PGP signing would help against GitHub (or some vulnerability) allowing someone to add extra git commits?

Since you yourself regularly commit signed commits ontop of the history you have locally (and you probably trust the history you have locally), the integrity of the whole history reachable from commits you signed can be verified by knowing just your PGP key.  So if that's the purpose, PGP signing does make sense, but it's not necessary that all contributors sign their commits;  the only commits that need to be signed would be version bumps etc. because people would probably want to verify and checkout those.  (And all commits made earlier than that could be verified too because they're parents of signed commits and thus must've been in your (trusted) local history at some point)

PeterSurda:
> do you mean PGP signing would help against GitHub (or some vulnerability) allowing someone to add extra git commits

That's how I understand it.

> PGP signing does make sense, but it's not necessary that all contributors sign their commits

I think you're right, but it's easier to automate check if signing is mandatory. I have various git hooks/scripts which enforce all commits to be signed, it also protects against me disabling signing by mistake or someone trying to attack my local repositories (my PGP key is on a yubikey so it's more protected). I have for example hooks for importing translations from transifex, one will create a signed pull request, and my merge script fails it means it wasn't signed properly and something fishy is going on.

If you put your PGP pubkey into your github account's settings, and then add a -S when committing, that should fix things.

shak-mar:
Yup, that makes sense.  I think it's not worthwhile to sign and then force-push my two PRs though?  You'll probably reproduce them yourself anyways;  and I'd probably want to generate a new PGP key just for git signing to keep things separated.

PeterSurda:
You can try it just to see if it works, you should be able to review it with "git log --show-signature" and the commits will have a [Verified] text next to them (check PR #1104 for example).

Issue #919, opened 2016-11-23
basic feature: discard button for "send" tab


ProZiro:
A basic discard (reset) button is needed under the "send" tab so that if one decides to not send the message, all fields can be reset so nothing is inadvertantly left to be accidently sent or other problematic issue.

PR #1116, opened 2018-02-05
New button "Blank" on tab "Send" to blank all fields (#919)

g1itch:
Hello!

Since new 0.6.3 release is soon (as you wrote), I decided to add also this button to current (non-refactored) code. Because I need it myself from time to time.

However I'm not sure about label: should it be "Discard" like proposed in #919 or maybe "Clear"?

### DIFF
--- a/src/bitmessageqt/__init__.py
+++ b/src/bitmessageqt/__init__.py
@@ -147,6 +147,8 @@ class MyForm(settingsmixin.SMainWindow):
             "clicked()"), self.click_pushButtonAddSubscription)
         QtCore.QObject.connect(self.ui.pushButtonTTL, QtCore.SIGNAL(
             "clicked()"), self.click_pushButtonTTL)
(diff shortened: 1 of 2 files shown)


PeterSurda:
Excellent. I'm not competely sure what to put there either (in English). Also "Cancel", "Reset" or "Abandon" are possible. I'll do some research.

PeterSurda:
I thought about it, and I think the best solution would be to create a new temporary tab for each new composed message. The tab would have a "Cancel" button and also an x mark. This is how other email clients work now. But this would require  a more significant redesign, so this PR is acceptable for now. As it's now, "Clear" or "Reset" are probably the best choices. I'll leave it up to you which one you choose.

g1itch:
It's hard for me to imagine such number of tabs nested in each other. But when refactoring be finished it will be possible to freely experiment with ui-files. Maybe someone would design a proper forms.

I've chosen "Clear" term.

PR #1117, opened 2018-02-06
Change unread status of message in "All Accounts" when it changes in "Chans"

g1itch:
Hello!

I completely rewritten `MyForm.tableWidgetInboxItemClicked()` and now message unread status seems to be consistent between "Chans" and "All Accounts".

I also tried to make other methods using the `QTableWidgetItem.setUnread()` similar as much as possible and commented out unused code in it.


PeterSurda:
thanks I'll look at it later today

PR #1122, opened 2018-02-12
Marked folder names for translation

g1itch:
Hello!

I slightly edited the `foldertree` module to really translate folder names. However, I'm not sure it is safe.


PeterSurda:
It should be, that's why I used the DisplayRole/EditRole separation. Let me do a quick review.

Issue #1138, opened 2018-02-17
Can't add entries to black list using Add Entry button


PeterSurda:
@g1itch:

Came over in bitmessage chan:

Unhandled exception
Traceback (most recent call last):
  File "PyBitmessage/src/bitmessageqt/blacklist.py", line 59, in click_pushButtonAddBlacklist
    if self.NewBlacklistDialogInstance.ui.labelAddressCheck.text() == _translate("MainWindow", "Address is valid."):
AttributeError: 'AddAddressDialog' object has no attribute 'ui'

------------------------------------------------------
Can't add entries to black list using Add Entry button
is anyone experiencing the same problem?

g1itch:
Hmm, I see the source. Fortunately `blacklist` module is not part of merged `ui-refactoring` branch, so this is not my mistake. Testing patch now.

PeterSurda:
odd I haven't had this problem before

g1itch:
7800272

g1itch:
Excuse me, part of it is mine. After dialog refactoring this code started to fail because it relies on old dialog members.

PeterSurda:
thanks

PR #1139, opened 2018-02-17
Quick fix for refactoring artifact in blacklist module (#1138)

g1itch:
I missed the `blacklist` module when changing address dialogs.


g1itch:
The right way to obtain address and label from address dialog now is
```python
dialog = dialogs.AddAddressDialog(self)
dialog.exec_()
try:
    address, label = dialog.data
except AttributeError:
    return
```

I planned to rewrite those code related to `bitmessage.ui` in the second part of refactoring.

PR #1150, opened 2018-03-09
Startup problem in openwrt

g1itch:
Hello.

There is no lib symlinks in openwrt, so pybitmessage doesn't start there without this patch.


PeterSurda:
Tested on OpenWRT 17.01.4, both the bug and the fix confirmed.

PR #1153, opened 2018-03-12
qrcode plugin fixes

g1itch:
Hello.

This is my technical debt on qrcode plugin:
  - all unused code removed;
  - menu entry is shown on any `QTreeWidgetItem` with address;
  - such plugins now have prefix `address.`;
  - 'BM:' is a schema for QR-code now.

Should I make that plugins also available for Address book and Blacklist entries?

### DIFF
--- a/src/plugins/menu_qrcode.py
+++ b/src/plugins/menu_qrcode.py
@@ -70,7 +70,7 @@ def connect_plugin(form):
             dialog = form.qrcode_dialog
         except AttributeError:
             form.qrcode_dialog = dialog = QRCodeDialog(form)
-        dialog.render(str(form.getCurrentAccount()))
+        dialog.render('bitmessage:' + str(form.getCurrentAccount()))
         dialog.exec_()
 
     return on_action_ShowQR, _translate("MainWindow", "Show QR-code")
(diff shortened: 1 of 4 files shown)


PeterSurda:
Can look at the codacy review and add docstrings? Don't worry about the "name doesn't conform", I'll try to adjust the check so that it doesn't complain.

PeterSurda:
I changed the method and function naming regexp, let's see if it is better now.

PeterSurda:
Maybe it takes a while for the changes to kick in.

src/plugins/menu_qrcode.py
```diff
@@ -70,7 +70,7 @@ def on_action_ShowQR():
             dialog = form.qrcode_dialog
         except AttributeError:
             form.qrcode_dialog = dialog = QRCodeDialog(form)
-        dialog.render(str(form.getCurrentAccount()))
+        dialog.render(str(form.getCurrentAccount()).replace('-', ':'))
```
PeterSurda: I think it should be `bitmessage:BM-aldkfjaolsfkjalfgkalkg`

PR #1157, opened 2018-03-15
fixed pep8 issue

MahendraNG:
I have Pep8 errors and removed unused imports.


src/depends.py
```diff
@@ -35,24 +37,29 @@ def check_hashlib():
     import hashlib
     if '_hashlib' not in hashlib.__dict__:
         logger.error(
-            'The RIPEMD-160 hash algorithm is not available. The hashlib module is not linked against OpenSSL.')
+            'The RIPEMD-160 hash algorithm is not available\
+            . The hashlib module is not linked against OpenSSL.')
```
g1itch:
```python
logger.error(
    'The RIPEMD-160 hash algorithm is not available.'
    ' The hashlib module is not linked against OpenSSL.')
```

PeterSurda: One step a time

PR #1164, opened 2018-03-20
Foxmod

ffox8:
minor clarification. 

also syncs my branch  ;-)


PeterSurda:
It can also be installed using setuptools.

sigoa:
... which is subsumed under "install from github sources" since BM is not on pypi yet.

PR #1169, opened 2018-03-21
Random (shuffle, sample, randrange) in helper random.

MahendraNG:
Random methods(shuffle, sample, randrange)


MahendraNG:
updated, have a look.

PR #1171, opened 2018-03-21
Address operations: done flake8 fixes and rewrited

g1itch:
Hello.

This is my variant of cleanup for code related to address operations. I picked it out from my keystore-plugin branch.

### DIFF
--- a/src/class_singleWorker.py
+++ b/src/class_singleWorker.py
@@ -34,7 +34,6 @@ from helper_sql import sqlQuery, sqlExecute
 # This thread, of which there is only one, does the heavy lifting:
 # calculating POWs.
 
-
 def sizeof_fmt(num, suffix='h/s'):
     for unit in ['', 'k', 'M', 'G', 'T', 'P', 'E', 'Z']:
         if abs(num) < 1000.0:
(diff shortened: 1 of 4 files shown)


g1itch:
As for `tr._translate()` which codacy considers as "Access to a protected member _translate of a client class" I can rewrite that code with `from tr import _translate` using `_translate()` instead like it done in ui-related modules.

g1itch:
Also should I replace `privSigningKeyHex` with underscore. What's your current convention about unused variables in tuple assignment?

PeterSurda:
@g1itch forget about `_tr` at the moment. Regarding unused variables, use `_` and then put a comment in the previous (or next) line with all named variables.

src/class_singleWorker.py
```diff
@@ -390,383 +452,623 @@ def sendOutOrStoreMyV4Pubkey(self, myAddress):
...
+                            ackdata,
+                            tr._translate(
+                                "MainWindow",
+                                "Problem: You are trying to send a"
+                                " message to yourself or a chan but your"
+                                " encryption key could not be found in"
+                                " the keys.dat file. Could not encrypt"
+                                " message. %1"
+                            ).arg(l10n.formatTimestamp()))
+                    ))  # log or show the address maybe?
```
PeterSurda: Well, this changes the status, so the recipient address is in the same row of the message list.

g1itch:
"Use implicit string literal continuation to break that literal across multiple lines inside the parens" means to not use `\` to break the strings. Any consequent strings are concatenated implicitly, inside parens you don't need `\` for line continuation.

g1itch:
I have a problem with codacy... https://ibb.co/irwZzc

g1itch:
@PeterSurda, should I turn `_getKeysForAddress` and `_doPOWDefaults` into functions or maybe staticmethod?

PeterSurda:
@g1itch maybe separately, I don't want to do big PRs as it's difficult to review. I'll merge this now.

PR #1172, opened 2018-03-21
Foldertree flake8

g1itch:
Hello.

There is flake8 fixes and new subclassing in module `bitmessageqt.foldertree` picked up from my `ui-refactoring` branch.

### DIFF
--- a/src/bitmessageqt/foldertree.py
+++ b/src/bitmessageqt/foldertree.py
@@ -1,10 +1,9 @@
 from PyQt4 import QtCore, QtGui
-from string import find, rfind, rstrip, lstrip
 
 from tr import _translate
 from bmconfigparser import BMConfigParser
 from helper_sql import *
-from utils import *
+from utils import avatarize
 from settingsmixin import SettingsMixin
 
 # for pylupdate
@@ -22,7 +21,7 @@ class AccountMixin(object):
     SUBSCRIPTION = 4
     BROADCAST = 5
 
-    def accountColor (self):
+    def accountColor(self):
         if not self.isEnabled:
             return QtGui.QColor(128, 128, 128)
         elif self.type == self.CHAN:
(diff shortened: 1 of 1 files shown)


src/bitmessageqt/foldertree.py
```diff
@@ -107,44 +114,63 @@ def defaultLabel(self):
         elif self.address is None or self.type == AccountMixin.ALL:
             return unicode(
                 str(_translate("MainWindow", "All accounts")), 'utf-8')
-        if retval is None:
-            return unicode(self.address, 'utf-8')
-        else:
-            return retval
 
+        return retval or unicode(self.address, 'utf-8')
```
PeterSurda: I'm not sure this is correct. I think it returns a boolean which is either true or false.

PeterSurda: It probably should be more like `return retval if retval is not None else unicode(self.address, 'utf-8')`

g1itch: It will return False only is `unicode(self.address, 'utf-8')` is False. But well it should be tested off course. That `foldertree` module is really essential for ui operation.

g1itch: You can try yourself:

In [2]: def test_or_return(foo, bar):
   ...:     return foo or bar
   ...: 

In [3]: test_or_return(1, 2)
Out[3]: 1

In [4]: test_or_return(None, 2)
Out[4]: 2

In [5]: test_or_return('', 2)
Out[5]: 2

In [6]: test_or_return('', '')
Out[6]: ''

In [7]: test_or_return('', False)
Out[7]: False

PeterSurda: ok

src/bitmessageqt/foldertree.py
```diff
@@ -502,26 +485,22 @@ def data(self, role):
...
-        stringList = []
-        text = unicode(path.toUtf8(), encoding="UTF-8")
-        splitIndex = rfind(text[0:self.widget().cursorPosition()], ";") + 1
-        str = text[splitIndex:self.widget().cursorPosition()]
-        str = rstrip(lstrip(str))
-        stringList.append(str)
-        return stringList
-        
+        text = unicode(path.toUtf8(), 'utf-8')
+        return [text[:self.widget().cursorPosition()].split(';')[-1].strip()]
```
PeterSurda: Did't you forget `lstrip`? I think that applies to situations when there is a space in front of the label.

g1itch: There is `''.strip()` instead which is supposed to replace both `rstrip` and `lstrip`

PeterSurda:
Were these additional commits just a rebase?

g1itch:
Yes, I just moved the branch to v0.6 tip, no additional commits.

PR #1175, opened 2018-03-21
BMinstallMenu

ffox8:
handy menu to clone, update and run BM


PeterSurda:
I would create a contrib directory and put it there, not in dev. dev is for testing new features.

sigoa:
lets hope the new sub-dir PRq will be activated

PR #1178, opened 2018-03-21
BMinstallMenu - the actual bash script

ffox8:
super easy download + run of Bitmessage , modify it to get more maintainer workflow done. very good for beginners who use Linux.


sigoa:
the smilie   :-)   in   option  9  needs to be removed !  At the bottom line of the script.  :cactus:

PR #1186, opened 2018-03-23
TravisRunner , put tests in here

sigoa:
TravisRunner doing nothing yet. put tests in there

to make meaningful Travis stuff, we can put python unit test etc. in here.

However,  on Travis Ubuntu, not everything installs - such as:   **pyqt4**  seems to be a problem


sigoa:
look into my fork, I got Travis running nicely !  :+1:

PR #1189, opened 2018-03-23
Flake8 for bitmessagemain, class_singleCleaner, helper_generic, shutdown

g1itch:
Hello.

This another portion of flake8 fixes is from #1060.

### DIFF
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -82,13 +82,11 @@ def connectToStream(streamNumber):
         state.maximumNumberOfHalfOpenConnections = 9
     else:
         state.maximumNumberOfHalfOpenConnections = 64
-    try:
-        # don't overload Tor
-        if BMConfigParser().get(
-                'bitmessagesettings', 'socksproxytype') != 'none':
-            state.maximumNumberOfHalfOpenConnections = 4
-    except:
-        pass
+
+    # don't overload Tor
+    if BMConfigParser().safeGet(
+            'bitmessagesettings', 'socksproxytype', 'none') != 'none':
+        state.maximumNumberOfHalfOpenConnections = 4
 
     with knownnodes.knownNodesLock:
         if streamNumber not in knownnodes.knownNodes:
@@ -158,27 +156,33 @@ def _fixSocket():
         socket.IPV6_V6ONLY = 27
 
 
+def _getApiAddress():
+    if BMConfigParser().safeGetBoolean(
+            'bitmessagesettings', 'apienabled'):
+        return [
+            BMConfigParser().get('bitmessagesettings', 'apiinterface'),
+            BMConfigParser().getint('bitmessagesettings', 'apiport')
+        ]
+
+
 # This thread, of which there is only one, runs the API.
 class singleAPI(threading.Thread, helper_threading.StoppableThread):
     def __init__(self):
         threading.Thread.__init__(self, name="singleAPI")
+        self.address = _getApiAddress()
         self.initStop()
 
     def stopThread(self):
         super(singleAPI, self).stopThread()
         s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         try:
-            s.connect((
-                BMConfigParser().get('bitmessagesettings', 'apiinterface'),
-                BMConfigParser().getint('bitmessagesettings', 'apiport')
-            ))
+            s.connect(self.api_address)
             s.shutdown(socket.SHUT_RDWR)
             s.close()
         except:
             pass
 
     def run(self):
-        port = BMConfigParser().getint('bitmessagesettings', 'apiport')
         try:
             from errno import WSAEADDRINUSE
         except (ImportError, AttributeError):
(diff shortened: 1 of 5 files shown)


src/bitmessagemain.py
```diff
@@ -73,17 +77,17 @@ def connectToStream(streamNumber):
...
-    try:
-        # don't overload Tor
-        if BMConfigParser().get('bitmessagesettings', 'socksproxytype') != 'none':
-            state.maximumNumberOfHalfOpenConnections = 4
-    except:
-        pass
-    
+
+    # don't overload Tor
+    if BMConfigParser().safeGet(
```
PeterSurda: I'm not fully sure the safeGet functions handle missing values like this. Perhaps instead of this change, just add a default socksproxytype in bmconfigparser and remove 'none'?

src/bitmessagemain.py
```diff
@@ -148,46 +155,62 @@ def inet_pton(family, host):
...
         self.initStop()
-        
+
     def stopThread(self):
         super(singleAPI, self).stopThread()
         s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         try:
-            s.connect((BMConfigParser().get('bitmessagesettings', 'apiinterface'), BMConfigParser().getint(
-                'bitmessagesettings', 'apiport')))
+            s.connect(self.api_address)
```
PeterSurda: this should probably be `self.address`

src/bitmessagemain.py
```diff
@@ -233,66 +259,82 @@ def start(self):
...
-                apiNotifyPath = BMConfigParser().get(
-                    'bitmessagesettings', 'apinotifypath')
-            except:
-                apiNotifyPath = ''
-            if apiNotifyPath != '':
-                with shared.printLock:
-                    print('Trying to call', apiNotifyPath)
+        singleAPIThread = singleAPI()
+        singleAPIThread.daemon = True
+        singleAPIThread.start()
```
PeterSurda: It should only run when the api is enabled.

g1itch: I moved that check into `singleAPI.__init__`

src/bitmessagemain.py
g1itch: It does. And it's already used in many places in code doing the same `try: ... except:`

Look at the [def](/Bitmessage/PyBitmessage/blob/master/src/bmconfigparser.py#L72)

PeterSurda: I know it's used in other places but that doesn't mean it works correctly. Maybe it does. If you tested it then it's fine.

src/bitmessagemain.py
PeterSurda: For configuring the API yes, but not for starting the thread.

g1itch: It will not start if `apienaled = true` not set in config, look at line 200 and `_getApiAddress` (re)definition

PeterSurda: This is a little non-obvious. But it will still start the thread, even if it quits right away. Perhaps the init should raise an exception and then these 3 lines be wrapped in try/except instead.

g1itch: Thread will be stopped then because [serve_forever](/Bitmessage/PyBitmessage/blob/master/src/api.py#L61) not called

PeterSurda: What is the point of starting it when it quits right away? Isn't it better not to start it in the first place?

g1itch:
I've dropped that controversial commit for now.

src/bitmessagemain.py
```diff
@@ -73,17 +77,19 @@ def connectToStream(streamNumber):
...
     else:
         state.maximumNumberOfHalfOpenConnections = 64
     try:
         # don't overload Tor
-        if BMConfigParser().get('bitmessagesettings', 'socksproxytype') != 'none':
+        if BMConfigParser().get(
+                'bitmessagesettings', 'socksproxytype') != 'none':
             state.maximumNumberOfHalfOpenConnections = 4
     except:
         pass
```
omkar1117: Write log in the place of pass

src/bitmessagemain.py
```diff
@@ -334,40 +368,39 @@ def start(self):
...
-                        ' If you want to run in daemon mode, see '
-                        'https://bitmessage.org/wiki/Daemon'
-                        'You can also run PyBitmessage with'
-                        ' the new curses interface by providing'
-                        ' \'-c\' as a commandline argument.'
-                    )
-
+        if not state.enableGUI:
+            BMConfigParser().remove_option('bitmessagesettings', 'dontconnect')
+        elif daemon is False:
```
omkar1117: elif not daemon:

src/bitmessagemain.py
g1itch: Maybe it's better to remove try .. except?
```python
if BMConfigParser().safeGet('bitmessagesettings', 'socksproxytype') != 'none':
    state.maximumNumberOfHalfOpenConnections = 4
```

src/bitmessagemain.py
g1itch: This place is controversial and changes is not final. It depends on how these changes will be merged with #1221. Maybe this statement will be more like
```python
if not state.enableGUI:
    ...
elif daemon:
    ...
else:
    if state.curses:
        ...
```

g1itch:
Of course I've tested `BMConfigParser.safeGet` locally. But it seems I should rather write separate test case for config.

src/bitmessagemain.py
PeterSurda: safeGet shouldn't throw an exception, if it does, it should be fixed. I agree to have the try/except removed, and perhaps add the socksproxytype none to the config defaults.

src/bitmessagemain.py
```diff
@@ -16,10 +16,12 @@
 os.chdir(app_dir)
 sys.path.insert(0, app_dir)
 
+
```
omkar1117: Import should be placed in a order.
Please maintain orders
* import statements first
* from statements should be next.

g1itch: It's impossible in this module. PyBitmessage won't start after such changes.

PeterSurda:
would it be possible to split this into several PRs, it's difficult to handle such a big PR (review/merge)

g1itch:
Yes. Can you please formulate a criterion for the split?

PeterSurda:
In this case I'd prefer one file per PR.

g1itch:
May part of that PS's contain one commit instead? Because some commits here are containing multiple files.

PeterSurda:
Give me today to review it, if I'm ok then I can merge it as it is, if not then we'll see

g1itch:
OK, I split this into five PRs. But sorry, I cannot find where I can request your review for them?

PeterSurda:
@g1itch not sure, I have been doing this manually so far.

PR #1192, opened 2018-03-29
Add travis.yml and requirements.txt

PeterSurda:
- these are very minimalistic, with input from @g1itch, @sigoa and
  @venkateshpitta

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: python
+python:
+        - "2.7"
+install:
+        - pip install -r requirements.txt
+script: src/bitmessagemain.py -t
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,4 @@ python:
         - "2.7"
 install:
         - pip install -r requirements.txt
-        - sudo apt-get install python-msgpack python-qt4 python-pyopencl
 script: src/bitmessagemain.py -t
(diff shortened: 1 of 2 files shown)


MahendraNG:
To run travis successfull, you need to install namecoin
To install namecoin, follow given instrauctions:
sudo sh -c "echo 'deb http://download.opensuse.org/repositories/home:/p_conrad:/coins/xUbuntu_16.04/ /' > /etc/apt/sources.list.d/home:p_conrad:coins.list"
sudo apt-get update
sudo apt-get install namecoin

sudo sh -c "echo 'deb http://download.opensuse.org/repositories/home:/p_conrad:/coins/xUbuntu_16.04/ /' > /etc/apt/sources.list.d/home:p_conrad:coins.list"
sudo apt-get update
sudo apt-get install namecoin-gui

After that,
mkdir -p ~/.namecoin \
&& echo "rpcuser=`whoami`" >> ~/.namecoin/namecoin.conf \
&& echo "rpcpassword=`openssl rand -hex 30/`" >> ~/.namecoin/namecoin.conf \
&& echo "rpcport=8336" >> ~/.namecoin/namecoin.conf

PeterSurda:
Well, I could run the OpenCL tests on a separate machine using the same config file, but I agree at the moment it's not necessary. I moved pyopencl, msgpack and PyQt to apt-get because the didn't work with pip on travis (tries to build pyopencl from source and that fails due to missing headers, pyqt isn't available through pip, and it protested regarding msgpack for some reason as well). But for the time being I'll probably just leave them commented.

PR #1198, opened 2018-04-02
Move test mode into state

PeterSurda:
- test mode variable is now in state

### DIFF
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -325,12 +325,21 @@ class Main:
             upnpThread = upnp.uPnPThread()
             upnpThread.start()
 
-        if daemon == False and BMConfigParser().safeGetBoolean('bitmessagesettings', 'daemon') == False:
+        if daemon == False and BMConfigParser().safeGetBoolean('bitmessagesettings', 'daemon') == False and not state.testmode:
             if state.curses == False:
                 if not depends.check_pyqt():
-                    print('PyBitmessage requires PyQt unless you want to run it as a daemon and interact with it using the API. You can download PyQt from http://www.riverbankcomputing.com/software/pyqt/download   or by searching Google for \'PyQt Download\'. If you want to run in daemon mode, see https://bitmessage.org/wiki/Daemon')
-                    print('You can also run PyBitmessage with the new curses interface by providing \'-c\' as a commandline argument.')
-                    sys.exit()
+                    sys.exit(
+                        'PyBitmessage requires PyQt unless you want to'
+                        ' run it as a daemon and interact with it using'
+                        ' the API. You can download PyQt from'
+                        ' http://www.riverbankcomputing.com/software/pyqt/download'
+                        ' or by searching Google for \'PyQt Download\'.'
+                        ' If you want to run in daemon mode, see'
+                        ' https://bitmessage.org/wiki/Daemon\n'
+                        'You can also run PyBitmessage with the'
+                        ' curses interface by providing \'-c\''
+                        ' as a commandline argument.'
+                    )
 
                 import bitmessageqt
                 bitmessageqt.run()
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -327,7 +327,7 @@ class Main:
 
         if (daemon == False and 
             BMConfigParser().safeGetBoolean('bitmessagesettings', 'daemon') == False and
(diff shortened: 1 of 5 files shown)


g1itch:
```patch
diff --git a/src/bitmessagemain.py b/src/bitmessagemain.py
index 453010a..18508b2 100755
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -328,9 +328,18 @@ class Main:
         if daemon == False and BMConfigParser().safeGetBoolean('bitmessagesettings', 'daemon') == False:
             if state.curses == False:
                 if not depends.check_pyqt():
-                    print('PyBitmessage requires PyQt unless you want to run it as a daemon and interact with it using the API. You can download PyQt from http://www.riverbankcomputing.com/software/pyqt/download   or by searching Google for \'PyQt Download\'. If you want to run in daemon mode, see https://bitmessage.org/wiki/Daemon')
-                    print('You can also run PyBitmessage with the new curses interface by providing \'-c\' as a commandline argument.')
-                    sys.exit()
+                    sys.exit(
+                        'PyBitmessage requires PyQt unless you want to'
+                        ' run it as a daemon and interact with it using'
+                        ' the API. You can download PyQt from'
+                        ' http://www.riverbankcomputing.com/software/pyqt/download'
+                        ' or by searching Google for \'PyQt Download\'.'
+                        ' If you want to run in daemon mode, see'
+                        ' https://bitmessage.org/wiki/Daemon\n'
+                        'You can also run PyBitmessage with the'
+                        ' curses interface by providing \'-c\''
+                        ' as a commandline argument.'
+                    )
 
                 import bitmessageqt
                 bitmessageqt.run()
```

> Help on built-in function exit in module sys:
> 
> exit(...)
>     exit([status])
>     
>     Exit the interpreter by raising SystemExit(status).
>     If the status is omitted or None, it defaults to zero (i.e., success).
>     If the status is an integer, it will be used as the system exit status.
>     If it is another kind of object, it will be printed and the system
>     exit status will be one (i.e., failure).

I guess all that `sys.exit()` calls should be treated as failure.

My current problem is that I cannot even repeat my successful `Build #2` (from my testing branch). The same `.travis.yml` no longer works ):

PeterSurda:
@g1itch not sure why you posted the change to `sys.exit` here. That's a separate issue, but I agree it's a good idea, please make a PR.

I'll work on fixing the travis build.

PeterSurda:
@g1itch I think #1199 is better than this one.

PeterSurda:
I see.

PeterSurda:
Tested in a VM, ReceiveQueue_x can stall, preventing correct shutdown. This is not specific to test mode, it happens in other situations as well. I think that's for a separate pull request.

PeterSurda:
Let me know @g1itch if it's ok now, Mahendra is taking a few days off, so I depend on you for code review now :-)

g1itch:
I thought it would be a quick and dirty solution for the first step of CI integration. To prevent obvious errors like extra spaces we should ensure BM starts successfully. In that case BM should start in foreground and stop without errors. Travis CI treats it as successful build.

Unfortunately such solution is not quick ): How do you think that hang on exit is specific for docker or Ubuntu 14? I cannot reproduce it locally and reducing the number of ReceiveQueue threads (8df59a4) not helps.

And vice versa, in the second step, I think `pybitmessage -t` should be called in [setUpClass() of TestSuite](https://docs.python.org/2/library/unittest.html#setupclass-and-teardownclass) and run as daemon stopping by timeout when all tests (API calls) are finished. I'll write an example for that soon. |: And this changes does not fit into such scenario.

PeterSurda:
I think what is happening is that the sql thread closes before receive data thread, and then it is waiting for a sql response which never comes. I'll add the fix to the PR.

Regarding setUpCleass/tearDownClass, I think than can be a separate PR.

PeterSurda:
Looks like it. I noticed it happens more often if data is being downloaded while shutdown is initiated.

PeterSurda:
You can approve whenever you're happy.

g1itch:
I'm worried that the changes to use `state.testmode` are quite extensive.

PeterSurda:
I'm include to agree, now that I think about it. daemon is basically the inverse of UI (plus forking), so your suggestion sounds better. Can you make a new PR, I'll make a new PR for the shutdown fixes.

PR #1199, opened 2018-04-02
Stop script if state.testmode.

MahendraNG:
Stop script if state.testmode.


g1itch:
This is another false positive. Please look into travis-ci build details and my explanation in #1198.

PR #1203, opened 2018-04-05
Changes for testmode

g1itch:
These are changes requested in #1193 made after reading #1198 and #1199. I also changed Travis CI configuration a bit. Should I further reduce timeout to 10 sec or 30 is ok?


PeterSurda:
I think 30s is ok, let me take a look at it.

PR #1206, opened 2018-04-06
helper_generic formating

MahendraNG:
helper_generic formatting for Codacy.

### DIFF
--- a/src/helper_generic.py
+++ b/src/helper_generic.py
@@ -1,5 +1,5 @@
 """This module is to perform generic oprations for threading."""
-"""and some conversion operations."""
+"""And some conversion operations."""
 
 import socket
 import sys
@@ -12,6 +12,7 @@ import shared
 from debug import logger
 import queues
 import shutdown
+from debug import logger
 
 
 def powQueueSize():
@@ -20,8 +21,8 @@ def powQueueSize():
         try:
             if thread.name == "singleWorker":
                 curWorkerQueue += thread.busy
-        except Exception:
-            print ""
+        except Exception as err:
+            logger.info("Thread error %s", err)
     return curWorkerQueue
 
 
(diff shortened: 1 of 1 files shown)


src/helper_generic.py
```diff
@@ -1,26 +1,30 @@
...
     curWorkerQueue = queues.workerQueue.qsize()
-    for thread in enumerate():
+    for thread in threading.enumerate():
         try:
             if thread.name == "singleWorker":
                 curWorkerQueue += thread.busy
-        except:
-            pass
+        except Exception:
+            print ""
```
PeterSurda: Maybe this can put an entry into the log or something, at INFO severity. In the past, there were threads being quickly created and destroyed, which could have caused an issue here without locking, but the current implementation launches all the threads at start and then keeps them running forever. There are minor exeptions like the UPnP thread, which can be turned on or off during runtime, but I don't see it as a problem here.

src/helper_generic.py
```diff
@@ -30,28 +34,37 @@ def convertIntToString(n):
...
 
+
 def signal_handler(signal, frame):
-    logger.error("Got signal %i in %s/%s", signal, current_process().name, current_thread().name)
+    logger.error("Got signal %i in %s/%s", signal,
+        current_process().name,
+        threading.current_thread().name)
     if current_process().name == "RegExParser":
-        # on Windows this isn't triggered, but it's fine, it has its own process termination thing
+        # on Windows this isn't triggered, but it's fine,
```
PeterSurda: The RegExParser process/thread was eliminated, this section can be removed.

src/helper_generic.py
```diff
@@ -30,28 +37,33 @@ def convertIntToString(n):
...
+            code.append('File: "%s", line %d, in %s' % (
+                filename, lineno, name))
             if line:
                 code.append("  %s" % (line.strip()))
     print "\n".join(code)
 
+
 def signal_handler(signal, frame):
-    logger.error("Got signal %i in %s/%s", signal, current_process().name, current_thread().name)
-    if current_process().name == "RegExParser":
```
g1itch: @PeterSurda, what about this condition, is it obsolete?

PR #1208, opened 2018-04-07
Helper startup format

MahendraNG:
helper_startup formating


PeterSurda:
@g1itch it did overlap so I closed the other PR.

PR #1210, opened 2018-04-07
helper_threading formating

MahendraNG:
helper_threading formating

### DIFF
--- a/src/helper_threading.py
+++ b/src/helper_threading.py
@@ -5,7 +5,11 @@ import threading
 
 try:
     import prctl
-
+except ImportError:
+    def set_thread_name(name):
+        """Set the thread name for external use (visible from the OS)."""
+        threading.current_thread().name = name
+else:
     def set_thread_name(name):
         """Set a name for the thread for python internal use."""
         prctl.set_name(name)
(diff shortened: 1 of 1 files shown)


src/helper_threading.py
```diff
@@ -12,17 +16,20 @@ def _thread_name_hack(self):
     threading.Thread.__bootstrap_original__ = threading.Thread._Thread__bootstrap
     threading.Thread._Thread__bootstrap = _thread_name_hack
 except ImportError:
-    def set_thread_name(name): threading.current_thread().name = name
+    def set_thread_name(name):
```
PeterSurda: Can you add docstring? This sets a name for the thread for python internal use.

src/helper_threading.py
```diff
@@ -1,9 +1,13 @@
+"""Helper threading perform all the threading operations."""
+
 from contextlib import contextmanager
 import threading
 
 try:
     import prctl
-    def set_thread_name(name): prctl.set_name(name)
+
+    def set_thread_name(name):
```
PeterSurda: This sets the thread name for external use (visible from the OS).

PR #1211, opened 2018-04-07
helper_inbox formating

MahendraNG:
Helper Inbox formating


g1itch:
There are more important problems with this module:
 - it's rarely used in code though must be
 - it covers very small subset of SQL-queries used in code not having "sql" in it's name

Personally, I dropped this module in favour of more complete [helper_db](/g1itch/PyBitmessage/blob/ui-refactoring/src/helper_db.py) in my `ui-refactoring` branch.

PR #1212, opened 2018-04-07
helper_msgcoding formating

MahendraNG:
Helper msgcoding formatting


src/helper_msgcoding.py
```diff
@@ -140,22 +142,26 @@ def decodeSimple(self, data):
         self.body = body
 
 if __name__ == '__main__':
```
PeterSurda: Can you make this into a unit test? https://docs.python.org/2/library/unittest.html

PR #1214, opened 2018-04-09
component control switches

f97ada87:
This PR implements the component control switches discussed in the comments of PR #1149


src/class_singleCleaner.py
```diff
@@ -65,7 +65,7 @@ def run(self):
             # If we are running as a daemon then we are going to fill up the UI
             # queue which will never be handled by a UI. We should clear it to
             # save memory.
-            if shared.thisapp.daemon:
+            if shared.thisapp.daemon or not state.enableGUI: # FIXME redundant?
```
g1itch: Yes, maybe just `if not state.enableGUI:`

src/helper_generic.py
```diff
@@ -66,7 +67,7 @@ def signal_handler(signal, frame):
         "PyBitmessage", "MainThread"):
         return
     logger.error("Got signal %i", signal)
-    if shared.thisapp.daemon:
+    if shared.thisapp.daemon or not state.enableGUI: # FIXME redundant?
```
g1itch: Not sure if this needed

f97ada87:
Thanks for the comments. 

There's an unlikely corner case that I can think of, and for this reason I left the redundancy in place. If `shared.thisapp.daemon` is changed during runtime, _after_ the initialization code in `bitmessagemain`,  the variable `state.enableGUI` will not magically auto-update, resulting in inconsistent behaviour. 
I couldn't find any such cases in the current code, however, I didn't check every single PR in the queue so I can't be 100% sure. I left those overlaps in the code (with FIXME notices) to cover this small uncertainty window, so they can be cleaned up in a subsequent PR once the dust settles. They are harmless in any case. 

If the consensus is that they're unnecessary clutter, I can clean them up within this PR.

PeterSurda:
I'm really happy that there are now constructive debates and development is steadily progressing. This hasn't been the case for many years. Let me know how I can make it easier for you to contribute.

I haven't had that much time for development in the last two months or so, as I'm focusing on having a solid and secure infrastructure. That may take another month or so, then I'll do more development again.

PeterSurda:
Regarding `shared.thisapp.daemon` vs. `state.enableGUI`, I think we'll leave it as it is, including FIXMEs. Later this can be revisited and one of these two options removed, or, if new requirements appear, they could stay.

PR #1217, opened 2018-04-11
Added some communication channels to README.md

sandakersmann:
Added some communication channels to README.md


sandakersmann:
Yes. It is better with a link. You don't get a badge for Reddit, so it will be a little messy with one badge and one link.

PR #1220, opened 2018-04-15
Update broken link for windows.

bashgnu:
Update copyright years and download links.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -43,7 +43,7 @@ That's it! *Honestly*!
 
 #### Windows
 On Windows you can download an executable for Bitmessage
-[here](https://bitmessage.org/download/windows/Bitmessage.exe).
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -43,7 +43,7 @@ That's it! *Honestly*!
 
 #### Windows
 On Windows you can download an executable for Bitmessage
-[here](https://bitmessage.org/download/windows/Bitmessage.exe).
+[here](https://github.com/Bitmessage/PyBitmessage/releases/download/0.6.3.2/Bitmessage_x86_0.6.3.2.exe).
```
PeterSurda: Better would be to point to releases without a version.

PeterSurda: Although since the parent repo doesn't seem to exist anymore, maybe this can be done separately.

PR #1221, opened 2018-04-16
Implement unittest based tests in test mode

g1itch:
Hello.

I remember that you didn't like the idea to implement tests through API. And I didn't like to commit empty files. This PR contains both. But I suggest to take it as basis for testing.

It contains two partial test cases I wrote using API and `psutil` as well as empty test case which is aim to be run in main thread. All that implemented using standard `unittest` package. Travis CI build will fail if any test fail.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+python_prctl
+psutil
--- a/src/api.py
+++ b/src/api.py
@@ -1456,7 +1456,9 @@ class MySimpleXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
     def _handle_request(self, method, params):
         if method not in self.handlers:
             raise APIError(20, 'Invalid method: %s' % method)
-        return self.handlers[method](self, params)
+        result = self.handlers[method](self, params)
+        state.last_api_response = time.time()
+        return result
 
     def _dispatch(self, method, params):
         # pylint: disable=attribute-defined-outside-init
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -367,7 +367,7 @@ class Main:
                         time.time() - state.last_api_response >= 30):
                     self.stop()
         elif not state.enableGUI:
-            from tests import core as test_core
+            from tests import core as test_core  # pylint: disable=relative-import
             test_core_result = test_core.run()
             state.enableGUI = True
             self.stop()
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -367,10 +367,11 @@ class Main:
                         time.time() - state.last_api_response >= 30):
                     self.stop()
         elif not state.enableGUI:
-            from tests import core
-            test_core_result = core.run()
+            from tests import core as test_core
+            test_core_result = test_core.run()
             state.enableGUI = True
             self.stop()
+            test_core.cleanup()
             sys.exit(
                 'Core tests failed!'
                 if test_core_result.errors or test_core_result.failures
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -367,7 +367,15 @@ class Main:
                         time.time() - state.last_api_response >= 30):
                     self.stop()
         elif not state.enableGUI:
+            from tests import core
+            test_core_result = core.run()
+            state.enableGUI = True
             self.stop()
+            sys.exit(
+                'Core tests failed!'
+                if test_core_result.errors or test_core_result.failures
+                else 0
+            )
 
     def daemonize(self):
         grandfatherPid = os.getpid()
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -185,7 +185,10 @@ class Main:
             elif opt in ("-c", "--curses"):
                 state.curses = True
             elif opt in ("-t", "--test"):
-                state.testmode = daemon = True
+                state.testmode = True
+                if os.path.isfile(os.path.join(
+                        state.appdata, 'unittest.lock')):
+                    daemon = True
                 state.enableGUI = False  # run without a UI
                 # Fallback: in case when no api command was issued
                 state.last_api_response = time.time()
@@ -363,6 +366,8 @@ class Main:
                 if (state.testmode and
                         time.time() - state.last_api_response >= 30):
                     self.stop()
+        elif not state.enableGUI:
+            self.stop()
 
     def daemonize(self):
         grandfatherPid = os.getpid()
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -25,10 +25,8 @@ import getopt
 # Used to capture a Ctrl-C keypress so that Bitmessage can shutdown gracefully.
 import signal
 import socket
-from datetime import datetime
+import time
 from struct import pack
-from subprocess import call
-from time import sleep
 
 from helper_startup import (
     isOurOperatingSystemLimitedToHavingVeryFewHalfOpenConnections
@@ -189,6 +187,20 @@ class Main:
             elif opt in ("-t", "--test"):
                 state.testmode = daemon = True
                 state.enableGUI = False  # run without a UI
+                # Fallback: in case when no api command was issued
+                state.last_api_response = time.time()
+                # Apply special settings
+                config = BMConfigParser()
+                config.set(
+                    'bitmessagesettings', 'apienabled', 'true')
+                config.set(
+                    'bitmessagesettings', 'apiusername', 'username')
+                config.set(
+                    'bitmessagesettings', 'apipassword', 'password')
+                config.set(
+                    'bitmessagesettings', 'apinotifypath',
+                    os.path.join(app_dir, 'tests', 'apinotify_handler.py')
+                )
 
         # is the application already running?  If yes then exit.
         if state.enableGUI and not state.curses and not depends.check_pyqt():
@@ -206,7 +218,7 @@ class Main:
             )
         shared.thisapp = singleinstance("", daemon)
 
-        if daemon and not state.testmode:
+        if daemon:
             with shared.printLock:
                 print('Running as a daemon. Send TERM signal to end.')
             self.daemonize()
@@ -346,12 +358,11 @@ class Main:
             BMConfigParser().remove_option('bitmessagesettings', 'dontconnect')
 
         if daemon:
-            if state.testmode:
-                sleep(30)
-                # make testing
-                self.stop()
             while state.shutdown == 0:
-                sleep(1)
+                time.sleep(1)
+                if (state.testmode and
+                        time.time() - state.last_api_response >= 30):
+                    self.stop()
 
     def daemonize(self):
         grandfatherPid = os.getpid()
@@ -362,7 +373,7 @@ class Main:
                 shared.thisapp.cleanup()
                 # wait until grandchild ready
                 while True:
-                    sleep(1)
+                    time.sleep(1)
                 os._exit(0)
         except AttributeError:
             # fork not implemented
@@ -383,7 +394,7 @@ class Main:
                 shared.thisapp.cleanup()
                 # wait until child ready
                 while True:
-                    sleep(1)
+                    time.sleep(1)
                 os._exit(0)
         except AttributeError:
             # fork not implemented
--- a/src/network/networkthread.py
+++ b/src/network/networkthread.py
@@ -1,11 +1,12 @@
 import threading
 
-from bmconfigparser import BMConfigParser
+import network.asyncore_pollchoose as asyncore
+import state
 from debug import logger
 from helper_threading import StoppableThread
-import network.asyncore_pollchoose as asyncore
 from network.connectionpool import BMConnectionPool
-import state
+from queues import excQueue
+
 
 class BMNetworkThread(threading.Thread, StoppableThread):
     def __init__(self):
@@ -15,8 +16,12 @@ class BMNetworkThread(threading.Thread, StoppableThread):
         logger.info("init asyncore thread")
 
     def run(self):
-        while not self._stopped and state.shutdown == 0:
-            BMConnectionPool().loop()
+        try:
+            while not self._stopped and state.shutdown == 0:
+                BMConnectionPool().loop()
+        except Exception as e:
+            excQueue.put((self.name, e))
+            raise
 
     def stopThread(self):
         super(BMNetworkThread, self).stopThread()
--- a/src/tests/apinotify_handler.py
+++ b/src/tests/apinotify_handler.py
@@ -1,0 +1,14 @@
+#!/usr/bin/env python
+
+import os
+import sys
+import tempfile
+from datetime import datetime
+
+
+if __name__ == '__main__':
+    if sys.argv()[1] == 'startingUp':
+        with open(
+            os.path.join(tempfile.gettempdir(), '.api_started'), 'wb'
+        ) as start_file:
+            start_file.write(datetime.now())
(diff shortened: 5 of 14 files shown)


src/bitmessagemain.py
```diff
@@ -25,9 +25,10 @@
 import errno
 import socket
 import ctypes
+import subprocess
```
omkar1117: imports should follow the order.

import  should be first.
from statements should be next to them.

src/bitmessagemain.py
```diff
@@ -342,8 +368,8 @@ def start(self):
             # Populate with hardcoded value (same as connectToStream above)
             state.streamsInWhichIAmParticipating.append(1)
 
-        if daemon == False and state.enableGUI: # FIXME redundant?
-            if state.curses == False:
+        if daemon is False and state.enableGUI:  # FIXME redundant?
```
omkar1117: `if not daemon and state.enableGUI:  # FIXME redundant?` 
       `if not state.curses: `

src/tests/test_api.py
```diff
@@ -0,0 +1,171 @@
+import unittest
+import xmlrpclib
+import base64
+import json
+from time import sleep
+
```
omkar1117: Remove empty line

src/tests/test_api.py
```diff
@@ -0,0 +1,171 @@
...
+    _threads_count = 13  # minus one ReceiveQueue
+    _seed = base64.encodestring(
+        'TIGER, tiger, burning bright. In the forests of the night'
+    )
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestAPI, cls).setUpClass()
+        cls.addresses = []
+        cls.api = xmlrpclib.ServerProxy(
```
omkar1117: Check the URL better add it in the constant.

src/tests/test_process.py
```diff
@@ -0,0 +1,101 @@
+import unittest
+import subprocess
+import os
+import signal
+import tempfile
+from time import sleep
+from datetime import datetime
```
omkar1117: Import should be in a order.

src/tests/test_api.py
g1itch: For what purpose?

src/bitmessagemain.py
g1itch: I'd better drop this hunk. This PR is not about style.

src/bitmessagemain.py
g1itch: This PR is not about style, I have #1189 for that. Here I retained the original order of imports.

src/tests/test_api.py
g1itch: pep8 says the opposite: You should put a blank line between each group of imports.
https://www.python.org/dev/peps/pep-0008/#imports

src/bitmessagemain.py
g1itch: And also, can you give me a proof link? I don't see such requirements in pep8.

src/tests/test_process.py
g1itch: prooflink, please

src/tests/test_api.py
PeterSurda: In most cases I guess @omkar1117 has a point, but this is a unit test and I think it isn't necessary.

src/tests/test_process.py
PeterSurda: Imports should be in alphabetic order after being grouped. It's the flake8 H306 rule, even though it's not a part of PEP8 as far as I understand.

src/bitmessagemain.py
PeterSurda: https://www.python.org/dev/peps/pep-0008/#imports has a requirement regarding grouping, and while alphabetic order isn't a part of PEP8 as far as I see, it's a commonly used standard.

src/tests/test_process.py
g1itch: It's rather esoteric rule because I've never seen it's application while have flake8 checks enabled in emacs python-mode.

I've followed it but only for newly created modules.

g1itch:
Oops. I just tried to separate code cleanup ):

src/api.py
```diff
@@ -20,7 +20,9 @@
 
 import shared
```
omkar1117: imports should be placed in a order.

src/bitmessagemain.py
```diff
@@ -218,17 +236,34 @@ def start(self):
...
+                state.testmode = True
+                if os.path.isfile(os.path.join(
+                        state.appdata, 'unittest.lock')):
+                    daemon = True
+                state.enableGUI = False  # run without a UI
+                # Fallback: in case when no api command was issued
+                state.last_api_response = datetime.now()
+                # Apply special settings
+                config = BMConfigParser()
+                config.set(
```
omkar1117: `
CONFIGURATIONS = [('bitmessagesettings', 'apienabled', ('bitmessagesettings', 'apiusername', 'username'), ('bitmessagesettings', 'apipassword', 'password'), ('bitmessagesettings', 'apinotifypath', os.path.join(app_dir, 'tests', 'apinotify_handler.py')]`

for data in CONFIGURATIONS:
                                 config.set(data)

src/tests/test_process.py
```diff
@@ -0,0 +1,112 @@
+"""
+Common reusable code for tests and tests for pybitmessage process.
+"""
+
+import os
+import signal
+import subprocess
```
omkar1117: imports should follow a order

g1itch: Which order?

src/bitmessagemain.py
omkar1117: Any update here?

src/tests/test_api.py
```diff
@@ -0,0 +1,172 @@
...
+"""
+
+import base64
+import json
+import unittest
+import xmlrpclib
+from time import sleep
+
+from test_process import TestProcessProto
+
```
omkar1117: Add a constant as:
URL = "http://username:password@127.0.0.1:8442/"

src/tests/test_api.py
```diff
@@ -0,0 +1,172 @@
...
+    _seed = base64.encodestring(
+        'TIGER, tiger, burning bright. In the forests of the night'
+    )
+
+    @classmethod
+    def setUpClass(cls):
+        """Setup XMLRPC proxy for pybitmessage API"""
+        super(TestAPI, cls).setUpClass()
+        cls.addresses = []
+        cls.api = xmlrpclib.ServerProxy(
```
omkar1117: Replace with URL

cls.api = xmlrpclib.ServerProxy(URL)

src/tests/test_api.py
```diff
@@ -0,0 +1,172 @@
...
+                print('API start detected!')
+                return
+            sleep(1)
+
+    def _add_random_address(self, label):
+        return self.api.createRandomAddress(base64.encodestring(label))
+
+    def test_user_password(self):
+        """Trying to connect with wrong username/password"""
+        api_wrong = xmlrpclib.ServerProxy("http://test:wrong@127.0.0.1:8442/")
```
omkar1117: api_wrong = xmlrpclib.ServerProxy(URL)

omkar1117:
@g1itch , Please comment over the Comments I make with the Commit, so that it will be easy to refer.

src/bitmessagemain.py
g1itch: This looks ugly

PeterSurda:
@g1itch It looks like github doesn't send notifications if a PR was rebased, and it isn't even obvious if you look at this page. I can only see it if I manually pull the PR locally and look at the logs. Maybe next time you rebase and someone is waiting for a reply, just add a new comment saying that you rebased.

src/bitmessagemain.py
PeterSurda: Hmm not sure whether this is necessary, I think it can be kept as it is. It can always be changed later.

g1itch: @PeterSurda, I thought of moving this into `testConfig` when it will load a sort of fixtures and merge them into `keys.dat` and then run bitmessage and check results. But for now it's too complex. I suspect this PR will be merged never.

g1itch:
When I rebase PR's branch you can see all the commits below the last comment made before rebase. Otherwise only new pushed commits are appearing there.

src/tests/test_process.py
```diff
@@ -0,0 +1,112 @@
...
+        for pfile in cls._files:
+            try:
+                os.remove(os.path.join(cls.home, pfile))
+            except OSError:
+                pass
+
+    @classmethod
+    def tearDownClass(cls):
+        """Ensures that pybitmessage stopped and removes files"""
+        # os.kill(cls.pid, signal.SIGTERM)
```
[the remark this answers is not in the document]
g1itch: Ant suggestions?

src/tests/test_process.py
```diff
@@ -0,0 +1,112 @@
...
+        cls.process.send_signal(signal.SIGTERM)
+        try:
+            cls.process.wait(5)
+        except psutil.TimeoutExpired:
+            print(open(os.path.join(cls.home, 'debug.log'), 'rb').read())
+            cls.process.kill()
+        finally:
+            cls._cleanup_files()
+
+    def _test_threads(self):
```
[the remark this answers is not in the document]
g1itch: where?

requirements.txt
```diff
@@ -1 +1,2 @@
 python_prctl
+psutil
```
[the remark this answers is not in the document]
g1itch: This done according to Travis-CI doc and common conventions. PyBitmessage itself has no dependencies.

src/tests/test_process.py
```diff
@@ -0,0 +1,112 @@
...
+    _threads_count = 14
+    _files = (
+        'keys.dat', 'debug.log', 'messages.dat', 'knownnodes.dat',
+        '.api_started', 'unittest.lock'
+    )
+
+    @classmethod
+    def setUpClass(cls):
+        """Setup environment and start pybitmessage"""
+        cls.home = os.environ['BITMESSAGE_HOME'] = tempfile.gettempdir()
```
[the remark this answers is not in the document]
coffeedogs: Working at the top of /tmp and writing out files like `debug.log` is bad etiquette bordering on irresponsible. You're already using the tempfile module so why not create a temporary directory for your files? You can then get rid of _files and just remove the tempdir at cleanup.

src/tests/test_process.py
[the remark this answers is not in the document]
coffeedogs: Remove the commented out code.

src/tests/test_process.py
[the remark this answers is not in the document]
coffeedogs: pylint doesn't like multiline comments as docsrings even if pep8 does. We can either be consistent (something something hobgoblin minds) or we can add a pylint exception to avoid `Missing function docstring (missing-docstring)`. I prefer the former.

src/tests/test_process.py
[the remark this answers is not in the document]
g1itch: Travis-CI sets up a clean environment for each build.

requirements.txt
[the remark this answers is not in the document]
coffeedogs: So the docs [say](https://docs.travis-ci.com/user/languages/python/#Dependency-Management):
```
You can manually override this default install phase, for example:
install: pip install --user -r requirements.txt
```
So we could have something like `pip install .[testing]` it seems. I'm all in favour of pulling in requirements from txt files to keep it DRY (instead of the other way round as I did with checkdeps.py) I'd just ask that we call it test_requirements.txt or requirements-dev.txt or something since, as you say, PyBitmessage itself has no dependencies so requirements.txt should be empty or missing.

src/tests/apinotify_handler.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
coffeedogs: I wasn't sure which manual you meant so I read [pytest](https://docs.pytest.org/en/latest/goodpractices.html), [travis](https://docs.travis-ci.com/user/languages/python/#Examples), an [SO question](https://stackoverflow.com/questions/17001010/how-to-run-unittest-discover-from-python-setup-py-test) about what our objectives appear to be and the [linked article](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure) by Cristian Mărieș and all either support tests/ at the project root or encourage it. Four out of the six examples at the bottom of the travis link use a test directory at the top level.

src/tests/test_process.py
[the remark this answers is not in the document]
g1itch: Why? Just to satisfy a lint? Don't you understand what this method does? `TestProcess.test_threads()` has docstring, it's enough as for me.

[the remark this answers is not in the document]
coffeedogs: Never mind, my bad, with all the multi-line `#` comments I missed that this method starts with `_` so the linter doesn't even care.

I was only thinking about helping you get a green codacy commit. If creating a docstring on public methods is clearly not required then we should leave a lint marker disabling the check in the most specific place possible. 'satisfying the linter' is about being able to use the return code as part of our CI and as far as humans go, 'alert-blindness' is a thing. If you're used to ignoring pages of lint warning you're making it harder to see ones that matter (more).

src/tests/apinotify_handler.py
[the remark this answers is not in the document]
g1itch: Cannot find now, it was a long time ago. Maybe you are right on this point. But I don't think it's matters.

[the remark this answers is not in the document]
g1itch: > Four out of the six examples at the bottom of the travis link use a test directory at the top level.

OK, I would choose `simplejson` or `tornado` as an example. And they place tests inside of the package dir.

PeterSurda:
@coffeedogs @g1itch Can we maybe have a group call where we discuss how issues like this can be addressed better in the future?

g1itch:
@PeterSurda, I'm not sure it's necessary because the final decision is your.

src/bitmessagemain.py
```diff
@@ -238,8 +255,28 @@ def start(self):
             elif opt in ("-c", "--curses"):
                 state.curses = True
             elif opt in ("-t", "--test"):
-                state.testmode = daemon = True
+                state.testmode = True
+                if os.path.isfile(os.path.join(
+                        state.appdata, 'unittest.lock')):
+                    daemon = True
+                else:
+                    from tests import core as test_core
```
PeterSurda: Please use a comment to suppress the pylint warning, something like

    from tests import core as test_core # pylint: disable=relative-import

PeterSurda:
@g1itch You're free to merge as soon as the travis check succeeds. I haven't check it in detail, maybe it's failing due to the bootstrap servers being overloaded?

PeterSurda:
Well, regarding 6a96bf6, I would have to test it and I don't have time for that. I would try to restart the asyncore loop if there is an exception (during normal run, during tests it's ok like it is), but since it isn't handled now, it shouldn't create new problems.

src/network/networkthread.py
```diff
@@ -15,8 +16,12 @@ def __init__(self):
...
 
     def run(self):
-        while not self._stopped and state.shutdown == 0:
-            BMConnectionPool().loop()
+        try:
+            while not self._stopped and state.shutdown == 0:
+                BMConnectionPool().loop()
+        except Exception as e:
+            excQueue.put((self.name, e))
+            raise
```
g1itch: If I remove `raise`, any exception will be ignored.

PeterSurda:
Code looks good, I'll run some tests in a VM and if all goes well I'll approve.

PR #1222, opened 2018-04-16
implement netcat operating mode

f97ada87:
This PR enables a special headless operating mode (which I named "netcat" mode due to similarities with the Unix utility) where all object processing is disabled. Instead, raw objects received from the network are output to STDOUT unprocessed, also, any valid raw objects received on STDIN are broadcast to the network. 

This is a re-implementation of PR #1149 , using the switches defined in PR #1214 . The discussions from the linked PRs should provide useful background.


f97ada87:
a line-too-long codacy issue was fixed

src/std_io.py
```diff
@@ -0,0 +1,166 @@
...
+            # honour global shutdown flag
+            if state.shutdown != 0:
+                logger.info('stdInput thread shutdown.')
+                break
+
+class objectStdOut(threading.Thread):
+    """
+    The objectStdOut thread receives network objects from the receiveDataThreads.
+    """
+    def __init__(self):
```
omkar1117: Follow Pep8 styling format.

src/std_io.py
```diff
@@ -0,0 +1,166 @@
...
+        threading.Thread.__init__(self, name="objectStdOut")
+
+        # REFACTOR THIS with objectProcessor into objectProcessorQueue
+        queryreturn = sqlQuery(
+            '''SELECT objecttype, data FROM objectprocessorqueue''')
+        for row in queryreturn:
+            objectType, data = row
+            queues.objectProcessorQueue.put((objectType, data))
+        sqlExecute('''DELETE FROM objectprocessorqueue''')
+        logger.debug('Loaded %s objects from disk into the objectProcessorQueue.' % str(len(queryreturn)))
```
omkar1117: Pep8 styling missing

f97ada87:
@g1itch it stops clean on Ctrl-D (EOF) on STDIN or dirty with SIGKILL. Clean exit by signal is not available because of blocking read on STDIN. 
Do you think it's needed? If yes, I can use a non-blocking read instead.

f97ada87:
@omkar1117 human or robot? If human, we need to talk.

g1itch:
Of course, Ctrl+D. Yes, it works. It's enough for me.

src/std_io.py
```diff
@@ -0,0 +1,166 @@
...
+
+                # REFACTOR THIS with objectProcessor into objectProcessorQueue
+                numberOfObjectsInObjProcQueue = 0
+                with SqlBulkExecute() as sql:
+                    while queues.objectProcessorQueue.curSize > 0:
+                        objectType, data = queues.objectProcessorQueue.get()
+                        sql.execute('''INSERT INTO objectprocessorqueue VALUES (?,?)''',
+                                   objectType, data)
+                        numberOfObjectsInObjProcQueue += 1
+                logger.debug('Saved %s objects from the objectProcessorQueue to disk. objectProcessorThread exiting.' %
```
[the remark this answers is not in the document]
f97ada87: done

src/std_io.py
```diff
@@ -0,0 +1,166 @@
...
+    def __init__(self, inputSrc):
+        threading.Thread.__init__(self, name="stdInput")
+        self.inputSrc = inputSrc
+        logger.info('stdInput thread started.')
+
+    def run(self):
+        while True:
+            # read a line in hex encoding
+            line = self.inputSrc.readline()
+            if len(line) == 0:
```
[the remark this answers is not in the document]
f97ada87: "if not line" - can't strip an EOF :)

src/std_io.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    Standard Input thread reads objects from STDIN, posts them to Inventory
+    """
+
+    def __init__(self, inputSrc):
+        threading.Thread.__init__(self, name="stdInput")
+        self.inputSrc = inputSrc
+        logger.info('stdInput thread started.')
+
+    def run(self):
```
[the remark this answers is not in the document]
f97ada87: No. Go big or go home :)

src/std_io.py
f97ada87: done

src/std_io.py
f97ada87: done

f97ada87:
all pep8 issues are now fixed (except line length which is endemic) - thanks @omkar1117

src/std_io.py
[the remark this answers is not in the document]
PeterSurda: Over longer term, I would also prefer to have shorter methods, some parts of the old code, like the class_objectProcessor are too long, but for this PR it's fine the way it is.

g1itch:
It could be useful to have an option to dump the messages which caused an exceptions with help of this `std_io` module.

f97ada87:
I'm sorry, I'm not sure what you mean, in a number of ways :) 
1) by "messages which caused an exception", are you referring to:
- the objects received from userspace via stdin thread, which fail unhexlify or other pre-parsing operations, or
- the objects received from the network layer, which fail early sanity checks in bmproto.py and are dropped before even reaching inventory?

2) by "an option to dump", do you mean the objects should be:
- displayed in raw form, and if yes, where exactly, considering that all STDIO is busy; just send them to logger.info?, or
- dropped and not processed further, which is pretty much what we already do? :)

Asking because these are all things that I've considered. :)

f97ada87:
Yes, this makes sense. As a matter of fact, the initial commit from PR #1149 was tapping into bmproto.py upstream of the sanity checks, and was capturing broken objects along with the good ones. It was only after updating the code to tap into the ObjectProcessorQueue instead (downstream of bmproto) that we stopped capturing them. Anyway, this should be easy to add in the future if needed, with a couple of conditionals inside bmproto.py

f97ada87:
Hi guys, not trying to rush things, just double-checking if there are any open issues requiring attention from my end.

PeterSurda:
Haven't had time yet to check it out

src/std_io.py
```diff
@@ -0,0 +1,167 @@
...
+            readPosition += versionLength
+            streamNumber, streamNumberLength = decodeVarint(binObject[readPosition:readPosition + 10])
+            readPosition += streamNumberLength
+
+            # calculate inventory hash
+            inventoryHash = calculateInventoryHash(binObject)
+
+            # duplicate check on inventory hash (both netcat and airgap)
+            if inventoryHash in Inventory():
+                logger.info("STDIN: Already got object " + hexlify(inventoryHash))
```
coffeedogs: Not sure how many of these one might expect during 'normal' operation. Is 'info' the right log level? Maybe this and the next log statement could be debug level if it would otherwise lead to info being spammed. Maybe we expect people running in this mode to not care about other info level statements but be very interested in knowing that messages were or were not added, I don't know.

src/std_io.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    The objectStdOut thread receives network objects from the receiveDataThreads.
+    """
+    def __init__(self):
+        threading.Thread.__init__(self, name="objectStdOut")
+
+        # REFACTOR THIS with objectProcessor into objectProcessorQueue
+        queryreturn = sqlQuery(
+            '''SELECT objecttype, data FROM objectprocessorqueue''')
+        for row in queryreturn:
```
coffeedogs: By waiting until we have synchronously pulled all objects from the database and put all objects on the queue we are missing out on some benefits of parallelism and we'll hit a memory limit with large data sets. Maybe this is unavoidable or related to your suggested refactoring.

Also, keeping the number of places where raw SQL is used to a minimum is a good idea. Again, I assume this would be part of your suggested refactoring.

src/std_io.py
```diff
@@ -0,0 +1,167 @@
...
+            objectType, data = row
+            queues.objectProcessorQueue.put((objectType, data))
+        sqlExecute('''DELETE FROM objectprocessorqueue''')
+        logger.debug('Loaded %s objects from disk into the objectProcessorQueue.' % str(len(queryreturn)))
+        # /REFACTOR THIS
+
+    def run(self):
+        while True:
```
coffeedogs: When the queue is empty would this not cause unnecessary 100% CPU? Perhaps a small sleep is needed inside the while loop?

src/std_io.py
```diff
@@ -0,0 +1,167 @@
...
+                numberOfObjectsInObjProcQueue = 0
+                with SqlBulkExecute() as sql:
+                    while queues.objectProcessorQueue.curSize > 0:
+                        objectType, data = queues.objectProcessorQueue.get()
+                        sql.execute('''INSERT INTO objectprocessorqueue VALUES (?,?)''',
+                                    objectType, data)
+                        numberOfObjectsInObjProcQueue += 1
+                logger.debug('Saved %s objects from the objectProcessorQueue to disk. objectProcessorThread exiting.' %
+                             str(numberOfObjectsInObjProcQueue))
+                # /REFACTOR THIS
```
coffeedogs: Yes please!

src/std_io.py
```diff
@@ -0,0 +1,167 @@
...
+        sqlExecute('''DELETE FROM objectprocessorqueue''')
+        logger.debug('Loaded %s objects from disk into the objectProcessorQueue.' % str(len(queryreturn)))
+        # /REFACTOR THIS
+
+    def run(self):
+        while True:
+            objectType, data = queues.objectProcessorQueue.get()
+
+            # Output in documented format
+            print "%016x" % int(time.time()) + '\t' + hexlify(data)
```
coffeedogs: Should we use sys.stdout.write() instead of print here?

f97ada87:
@coffeedogs a general comment before I address some specific points; the answer to most your queries is along the lines of "backward compat", "minimizing the diff" and "one change at a time". 
This PR is ported from of a private fork that has several special operating modes: 
- `--mode-netcat / -n` described here,
- `--mode-airgap / -a` (no network, STDIO to ObjectProcessor) which will be ported next
- `--mode-router / -r` and `--mode-seeder / -s` which may never be ported

As much as possible, I tried to preserve compatibility with both PyBitmessage and the original fork.

src/bitmessagemain.py
```diff
@@ -205,8 +206,8 @@ def start(self):
 
         try:
             opts, args = getopt.getopt(
-                sys.argv[1:], "hcdt",
-                ["help", "curses", "daemon", "test"])
+                sys.argv[1:], "hcdtn",
```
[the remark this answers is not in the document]
f97ada87: I thought of that and decided that none of the common uses of "-n" (dry-run, no-resolve, no-output, numeric etc) are applicable in the Bitmessage context. 
A second best option would be "-c" (netCat) which is already taken.

src/bitmessagemain.py
```diff
@@ -205,8 +206,8 @@ def start(self):
...
+                ["help", "curses", "daemon", "test", "mode-netcat"])
```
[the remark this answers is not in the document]
f97ada87: See general comment, there's a naming theme.

src/std_io.py
```diff
@@ -0,0 +1,167 @@
...
+            except Exception:
+                logger.info("STDIN: Invalid input format")
+                continue
+
+            # sanity check on object size
+            if len(binObject) < 22:
+                logger.info("STDIN: Invalid object size")
+                continue
+
+            if not state.enableNetwork and state.enableGUI:
```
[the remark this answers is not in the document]
f97ada87: state.enableGUI is only false in netcat mode; std_io may have other uses outside of netcat mode, which are not currently included. Yes, the conditional makes sense from a logical perspective.

src/std_io.py
f97ada87: You're probably right. It's only useful when running in manual mode and pasting objects in console, useless in most other scenarios. I'll change it to debug. 
I guess what I really need here is not `logger.info` pollution, but the ability to switch logging levels as needed via getopt; I might actually open an issue about that.

PR #1226, opened 2018-04-25
Added API for raw objects

Kleshni:
I found out that the database has no indexes, so it makes little sense to filter objects with complex queries, because they will be execuded by linear search anyway.

Apart from that, everything is done.

`disseminateRawObject`
----------------------

Parameter: a hex-encoded object without packet headers, starting from nonce.

Tries to send the object to the network. POW must be already calculated.

Raises an error or returns the object's inventory hash. It can return the hash for invalid objects, but this problem can be avoided by not sending invalid objects to this method.

`queueRawObject`
----------------

Parameter: a hex-encoded object without packet headers and nonce, starting from expiry time.

Queues the object for POW calculation and sending to the network.

Returns a unique handle to track the objects state.

`checkQueuedRawObject`
----------------------

Parameter: a handle returned from the `queueRawObject` method.

Returns current status of the object:
- `queued` for waiting objects;
- `failed` for invalid objects;
- hex-encoded inventory hash for sent object. As with the `disseminateRawObject` there may be false positives;
- `notfound` for wrong handles and for handles that previously returned `failed` or an inventory hash.

If a queued object enters the `notfound` state, it can mean that the daemon was restarted, because the queued objects are not saved to the disk.

`getRawObject`
--------------

Parameter: an inventory hash.

Raises an error or returns a JSON object with the following fields:
- `hash` - the inventory hash;
- `objectType` - an integer;
- `streamNumber`;
- `payload` - hex-encoded object without packet headers, starting from nonce;
- `expiresTime` - a UNIX timestamp;
- `tag` - hex-encoded object tag or an empty string.

`listRawObjects`
----------------

Parameters: desired object type or -1, stream number or 0, start and end of the object slice, may be negative.

Returns a list of JSON objects with the following fields:
- `hash` - an inventory hash of a found object;
- `slice` - a hex-encoded fragment of the object, determined by the provided slice boundaries.

---

I see no reason to sing the commit with GPG, because the guidelines allow sending patches through a public Bitmessage channel. But if really need it, I can sign it with a random key.

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -22,10 +22,12 @@ import shared
 import time
 from addresses import (decodeAddress, addBMIfNotPresent, decodeVarint,
     calculateInventoryHash, varintDecodeError)
+import protocol
 from bmconfigparser import BMConfigParser
 import defaults
 import helper_inbox
 import helper_sent
+import helper_random
 import hashlib
 
 import state
@@ -63,6 +65,7 @@ class StoppableXMLRPCServer(SimpleXMLRPCServer):
         while state.shutdown == 0:
             self.handle_request()
 
+queuedRawObjects = {}
 
 # This is one of several classes that constitute the API
 # This class was written by Vaibhav Bhatia. Modified by Jonathan Warren (Atheros).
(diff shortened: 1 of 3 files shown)


src/api.py
```diff
@@ -951,6 +954,79 @@ def HandleGetMessageDataByDestinationHash(self, params):
         data += ']}'
         return data
 
+    def HandleDisseminateRawObject(self, params):
+        if len(params) != 1:
+            raise APIError(0, 'I need 1 parameter!')
+        payload = self._decode(params[0], "hex")
+
+        if protocol.checkAndShareObjectWithPeers(payload) == 0:
```
PeterSurda: This function is considered obsolete, but unfortunately there is no good replacement. Please put FIXME in front of it saying that it should be refactored in the future.

PeterSurda:
So, it looks like you know what you're doing. I still have to review it in a bit more detail. Regarding signing of the commits, that's primarily for repository integrity. If an anonymous contribution comes in, I'll commit and sign it under my name. But you can wait until I'm done with the review.

src/class_singleWorker.py
```diff
@@ -226,6 +231,18 @@ def _doPOWDefaults(self, payload, TTL,
         # inventoryHash = calculateInventoryHash(payload)
         return payload
 
+    def sendRawObject(self, data):
```
g1itch: How is this method related to sending any object? Where that objects get out of the queue?

PeterSurda: It's inserted into the Inventory(). What is missing for full functionality, the hash also needs to be inserted into the invQueue.

PeterSurda: occurrences of `broadcastToSendDataQueues` in `protocol.py`  should be replaced with `invQueue.put((stream, inventoryHash))` or something like that.

src/api.py
```diff
@@ -22,10 +22,12 @@
 import time
 from addresses import (decodeAddress, addBMIfNotPresent, decodeVarint,
     calculateInventoryHash, varintDecodeError)
```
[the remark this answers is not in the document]
omkar1117: Basic imports should be written first.

from statements should be at the end

src/api.py
```diff
@@ -1039,6 +1116,11 @@ def HandleShutdown(self, params):
     handlers['disseminatePubkey'] = HandleDissimatePubKey
     handlers['getMessageDataByDestinationHash'] = HandleGetMessageDataByDestinationHash
     handlers['getMessageDataByDestinationTag'] = HandleGetMessageDataByDestinationHash
+    handlers['disseminateRawObject'] = HandleDisseminateRawObject
+    handlers['queueRawObject'] = HandleQueueRawObject
+    handlers['checkQueuedRawObject'] = HandleCheckQueuedRawObject
```
omkar1117: Why can't we have all the dictionary additions as a single dictionary ?

src/class_singleWorker.py
```diff
@@ -137,6 +137,11 @@ def run(self):
                     self.sendBroadcast()
                 except:
                     pass
+            elif command == 'sendrawobject':
+                try:
+                    self.sendRawObject(data)
+                except:
+                    pass
```
omkar1117: Add logging with the exception.

src/api.py
```diff
@@ -951,6 +954,80 @@ def HandleGetMessageDataByDestinationHash(self, params):
         data += ']}'
         return data
 
+    def HandleDisseminateRawObject(self, params):
+        if len(params) != 1:
+            raise APIError(0, 'I need 1 parameter!')
+        payload = self._decode(params[0], "hex")
+
+        # FIXME: obsolete function
```
coffeedogs: If I understand this right, we're moving from using shared to protocol. This first use of protocol starts that migration in api.py. So a comment here saying 'obsolete function' could be confusing.

PR #1230, opened 2018-05-02
Fixed: Simple Codacy errors and warnings in src/helper_*

coffeedogs:
This PR has two purposes:

 * Fix or ignore lint errors for bandit/codacy
 * Test the PR process as a new developer

### DIFF
--- a/src/helper_msgcoding.py
+++ b/src/helper_msgcoding.py
@@ -140,7 +140,6 @@ class MsgDecode(object):
         self.body = body
 
 if __name__ == '__main__':
-    import random
     messageData = {
         "subject": ''.join(helper_random.randomchoice(string.ascii_lowercase + string.digits) for _ in range(40)),
         "body": ''.join(helper_random.randomchoice(string.ascii_lowercase + string.digits) for _ in range(10000))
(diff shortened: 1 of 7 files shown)


src/helper_msgcoding.py
```diff
@@ -140,7 +140,6 @@ def decodeSimple(self, data):
         self.body = body
 
 if __name__ == '__main__':
-    import random
     messageData = {
```
omkar1117: Pep8 validation needs Here,

As I found cosmetic changes only here, Please add the pep8 validation also here, so that we wont look back into this.

coffeedogs: Removing cruft could be considered cosmetic but I'm not about to debate semantics on my first PR. The scope of the work I was asked to do in this case was to address current Codacy issues for certain files, of which this was one. Other fixes that could be considered more than just cosmetic are in the pipeline so, brace yourself! I agree that pycodestyle and flake8 throughout would be good, and I intend to help that happen. Indeed, autopep8 (#1113) is in the 0.6.4 milestone.

PeterSurda: The PR doesn't need to fix everything. I'm happy with partial fixes as well.

PR #1233, opened 2018-05-03
Fixed: Straightforward Codacy issues in src/class_*.py

coffeedogs:
This PR addresses most straightforward Codacy issues in files at src/class_*.py. Some safe operations are marked nosec. This also updates the PR template with suggestions of improvement for the submitter's editor to avoid common style issues.

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -2,8 +2,6 @@
 
 - try to explain what the code is about
 - try to follow [PEP0008](https://www.python.org/dev/peps/pep-0008/)
-- make the pull request against the ["v0.6" branch](https://github.com/Bitmessage/PyBitmessage/tree/v0.6)
-- it should be possible to do a fast-forward merge of the pull requests
 - PGP-sign the commits included in the pull request
 - try to use a good editor that removes trailing whitespace, highlights potential python issues and uses unix line endings
 - You can get paid for merged commits if you register at [Tip4Commit](https://tip4commit.com/github/Bitmessage/PyBitmessage)
@@ -13,3 +11,4 @@ If for some reason you don't want to use github, you can submit the patch using 
 
 For helping with translations, please use [Transifex](https://www.transifex.com/bitmessage-project/pybitmessage/). There is no need to submit pull requests for translations.
 For translating technical terms it is recommended to consult the [Microsoft Language Portal](https://www.microsoft.com/Language/en-US/Default.aspx).
+
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -5,6 +5,7 @@
 - make the pull request against the ["v0.6" branch](https://github.com/Bitmessage/PyBitmessage/tree/v0.6)
 - it should be possible to do a fast-forward merge of the pull requests
 - PGP-sign the commits included in the pull request
+- try to use a good editor that removes trailing whitespace, highlights potential python issues and uses unix line endings
 - You can get paid for merged commits if you register at [Tip4Commit](https://tip4commit.com/github/Bitmessage/PyBitmessage)
 
 If for some reason you don't want to use github, you can submit the patch using Bitmessage to the "bitmessage" chan, or to one of the developers.
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -2,6 +2,7 @@
 
 - try to explain what the code is about
 - try to follow [PEP0008](https://www.python.org/dev/peps/pep-0008/)
+- make the pull request against the ["v0.6" branch](https://github.com/Bitmessage/PyBitmessage/tree/v0.6)
 - PGP-sign the commits included in the pull request
 - try to use a good editor that removes trailing whitespace, highlights potential python issues and uses unix line endings
 - You can get paid for merged commits if you register at [Tip4Commit](https://tip4commit.com/github/Bitmessage/PyBitmessage)
--- a/src/class_objectProcessor.py
+++ b/src/class_objectProcessor.py
@@ -3,19 +3,14 @@ import threading
 import shared
 import hashlib
 import random
-from struct import unpack, pack
-import sys
 import string
-from subprocess import call  # used when the API must execute an outside program
+from subprocess import call  # nosec
 import traceback
 from binascii import hexlify
 
-from pyelliptic.openssl import OpenSSL
 import highlevelcrypto
 from addresses import *
 from bmconfigparser import BMConfigParser
-import helper_generic
-from helper_generic import addDataPadding
 import helper_bitcoin
 import helper_inbox
 import helper_msgcoding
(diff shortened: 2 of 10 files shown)


PeterSurda:
@coffeedogs don't worry too much about codacy, I turned on every check available so it's complaining too much. It will be fine tuned over time.

src/class_objectProcessorQueue.py
```diff
@@ -20,7 +20,7 @@ def put(self, item, block = True, timeout = None):
     def get(self, block = True, timeout = None):
         try:
             item = Queue.Queue.get(self, block, timeout)
-        except Queue.Empty as e:
+        except Queue.Empty:
             raise Queue.Empty()
```
omkar1117: better to add logging here, because it will be easy to debug the application.

src/class_singleCleaner.py
```diff
@@ -34,6 +17,23 @@
...
+import time
+import os
+
+import tr#anslate
+from bmconfigparser import BMConfigParser
+from helper_sql import *
+from helper_threading import *
+from inventory import Inventory
+from network.connectionpool import BMConnectionPool
+from debug import logger
```
omkar1117: imports should be at the top first.
later from statements should be written to the file

src/class_objectProcessorQueue.py
coffeedogs: I agree that this would be a good spot for logging However that is outside the scope of the work (fix codacy errors) that I was asked to do. Feel free to submit a separate PR.

src/class_singleCleaner.py
coffeedogs: Autopep8 (and hopefully flake8 or similar) is in the pipeline (#1113) which will fix this and many, many other style and lint errors.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,10 +5,11 @@
 - make the pull request against the ["v0.6" branch](https://github.com/Bitmessage/PyBitmessage/tree/v0.6)
 - it should be possible to do a fast-forward merge of the pull requests
```
PeterSurda: This actually isn't necessary anymore, I found a way around it, this line can be removed.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,10 +5,11 @@
...
+- try to use a good editor that removes trailing whitespace, highlights potential python issues and uses unix line endings
 - You can get paid for merged commits if you register at [Tip4Commit](https://tip4commit.com/github/Bitmessage/PyBitmessage)
 
 If for some reason you don't want to use github, you can submit the patch using Bitmessage to the "bitmessage" chan, or to one of the developers.
 ## Translations
 
 For helping with translations, please use [Transifex](https://www.transifex.com/bitmessage-project/pybitmessage/). There is no need to submit pull requests for translations.
-For translating technical terms it is recommended to consult the [Microsoft Language Portal](https://www.microsoft.com/Language/en-US/Default.aspx).
\ No newline at end of file
+For translating technical terms it is recommended to consult the [Microsoft Language Portal](https://www.microsoft.com/Language/en-US/Default.aspx).
```
PeterSurda: I must be getting blind but can't spot the change in this line.

src/class_objectProcessor.py
```diff
@@ -539,7 +527,7 @@ def processmsg(self, data):
                 except:
                     apiNotifyPath = ''
                 if apiNotifyPath != '':
-                    call([apiNotifyPath, "newMessage"])
+                    call([apiNotifyPath, "newMessage"])  # nosec
```
PeterSurda: Ok but perphaps there could be a TODO for sanitisation? Not sure at the moment.

src/class_objectProcessor.py
```diff
@@ -792,7 +776,7 @@ def processbroadcast(self, data):
             except:
                 apiNotifyPath = ''
             if apiNotifyPath != '':
-                call([apiNotifyPath, "newBroadcast"])
+                call([apiNotifyPath, "newBroadcast"])  # nosec
```
PeterSurda: Here also perhaps a "TODO: sanitise"

src/class_objectProcessorQueue.py
PeterSurda: Logging not necessary here. This is a simple class extending Queue and this line is just a re-raising. It's up to wherever it's used to handle errors.

g1itch: Is that try/except needed here at all?

src/class_singleCleaner.py
PeterSurda: Extra caution required in importing project modules as there is some crappy initialisation which requires importing in a particular order. At the moment I don't recommend reordering project module imports as things may break strangely (system module import reordering is OK). This import order fix is a wholly separate project that would require some refactoring as well.

PeterSurda:
If you want, you can make the minor changes which I requested, I'll merge it in the evening one way or another.

src/class_objectProcessorQueue.py
PeterSurda: @g1itch I think you're right. @coffeedogs If you can, please remove line 21, 23 and 24.

PULL_REQUEST_TEMPLATE.md
coffeedogs: Toggling 'whitespace' in the github diff settings hides the line. This line came through in the email notification of your comment (but I can't find it in the github UI):

> \ No newline at end of file

src/class_objectProcessorQueue.py
coffeedogs: Will do.

src/class_objectProcessor.py
coffeedogs: Personally, I believe that any commit should be as good as we can make it; production quality. If we have TODOs and commented out code, this is a sign that we have lost our agile courage to change the code with confidence. In that vein, let's fully consider the consequences of this code and whether it should be #nosec'ed or not right now.

As far as I can see, apiNotifyPath is fetching a string from settings on the disk to use as the call() command, and the parameter to call() is a hard-coded string. If anyone had access to edit that setting to cause call() to do something bad then they have local access anyway and could just run arbitrary code if they wanted.

If anyone has any concerns that the call() command could in any way run code from the network then we should address that now rather than #nosec'ing it.

src/class_objectProcessor.py
coffeedogs: I believe that this is already sanitised (only looks to settings on local disk with fixed parameter).

PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,13 +2,13 @@
 
 - try to explain what the code is about
 - try to follow [PEP0008](https://www.python.org/dev/peps/pep-0008/)
-- make the pull request against the ["v0.6" branch](https://github.com/Bitmessage/PyBitmessage/tree/v0.6)
```
PeterSurda: Please don't remove this line regarding branches, only the one regarding fast-forward merging

PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,13 +2,13 @@
...
-- it should be possible to do a fast-forward merge of the pull requests
 - PGP-sign the commits included in the pull request
+- try to use a good editor that removes trailing whitespace, highlights potential python issues and uses unix line endings
```
PeterSurda: :+1:

src/class_singleWorker.py
```diff
@@ -1266,7 +1267,7 @@ def sendMsg(self):
                     except:
                         apiNotifyPath = ''
                     if apiNotifyPath != '':
-                        call([apiNotifyPath, "newMessage"])
+                        call([apiNotifyPath, "newMessage"])  # nosec
```
PeterSurda: can you remove nosec here please?

g1itch: So it's not a magic comment turning off some codacy check like `# noqa`?

coffeedogs: @g1itch: It is exactly that. Peter wants to wait to review the security implications of this class of warning before silencing them.

Issue #1232, opened 2018-05-03
Replace pickle for knownnodes.dat with something else


PeterSurda:

PeterSurda:
Fixed in 7878f83552a0954ff8394449a26dc7cb93bdb28e

PR #1235, opened 2018-05-04
Serialize knownnodes to json

g1itch:
Hello!

I've implemented serialization of `knownnodes` to json (#1232).


g1itch:
Yes, it seems I have no access. Hope in this case it's not a problem because this is a trivial changes. It took a couple of hours or so to write.

Can you also remind me which sentence I should write in commit comment "fixes #1232"? I'll reword.

PeterSurda:
It's not a problem in this case, it's just for the future as more people get involved it makes it more complicated for me to manage.

Regarding referencing, it's described in github documentation: https://blog.github.com/2013-01-22-closing-issues-via-commit-messages/

g1itch:
Hmm, I understand why it wont work like I thought. `master` still is default branch in this repo. The article says 

> once the commit is merged into your default branch

PeterSurda:
I think I reset the default branch to v0.6, no? It will only close the issue once I merge the PR, not when you open it.

g1itch:
OK, I didn't really tested yet. I'll try to add that sentence to last commit's comment if this PR will be considered acceptable.

src/knownnodes.py
```diff
@@ -1,49 +1,131 @@
...
 
 knownNodesTrimAmount = 2000
 
 # forget a node after rating is this low
 knownNodesForgetRating = -0.5
 
-def saveKnownNodes(dirName = None):
+
+def json_serialize_knownnodes(output):
+    _serialized = []
```
omkar1117: better to add doc string.

src/knownnodes.py
```diff
@@ -1,49 +1,131 @@
...
     with knownNodesLock:
         with open(os.path.join(dirName, 'knownnodes.dat'), 'wb') as output:
-            pickle.dump(knownNodes, output)
+            json_serialize_knownnodes(output)
+
+
+def addKnownNode(stream, peer, lastseen=None, is_self=False):
+    if lastseen is None:
+        lastseen = time.time()
+    knownNodes[stream][peer] = {
```
omkar1117: "lastseen" : lastseen if lastseen else time.time()

f97ada87:
This worked well for me (v0.6 current on Linux). Good work. 

However, if the JSON file is corrupted (due to any of a number of factors, including human error), `knownnodes.readKnownNodes()` fallbacks to `pickle_deserialize_old_knownnodes()` and crashes hard. Some better error handling might be helpful in such scenarios. 

For example, you could issue a _meaningful_ fatal error message and exit cleanly. Or even better, issue a warning, rename the corrupt/unreadable file to a bak extension, start a fresh one and continue the startup sequence seamlessly.

src/knownnodes.py
PeterSurda: :+1:

g1itch:
@f97ada87, pybitmessage can start without knownnodes.dat. So I think it will be better just to handle the exception.

f97ada87:
Agreed - it's not a condition severe enough to justify a quit, soft or hard.

PR #1251, opened 2018-05-16
Fixed: Type error, reported from the bitmessage chan

coffeedogs:
Fixed: Type error, reported from the bitmessage chan

Untested.

### DIFF
--- a/src/network/bmproto.py
+++ b/src/network/bmproto.py
@@ -334,7 +334,7 @@ class BMProto(AdvancedDispatcher, ObjectTracker):
         self.object = BMObject(nonce, expiresTime, objectType, version, streamNumber, self.payload, self.payloadOffset)
 
         if len(self.payload) - self.payloadOffset > BMProto.maxObjectPayloadSize:
-            logger.info('The payload length of this object is too large (%s bytes). Ignoring it.' % len(self.payload) - self.payloadOffset)
+            logger.info('The payload length of this object is too large (%d bytes). Ignoring it.' % (len(self.payload) - self.payloadOffset))
             raise BMProtoExcessiveDataError()
 
         try:


src/network/bmproto.py
```diff
@@ -334,7 +334,7 @@ def bm_command_object(self):
         self.object = BMObject(nonce, expiresTime, objectType, version, streamNumber, self.payload, self.payloadOffset)
 
         if len(self.payload) - self.payloadOffset > BMProto.maxObjectPayloadSize:
-            logger.info('The payload length of this object is too large (%s bytes). Ignoring it.' % len(self.payload) - self.payloadOffset)
+            logger.info('The payload length of this object is too large (%d bytes). Ignoring it.' % len(self.payload) - self.payloadOffset)
```
PeterSurda: doesn't it have to be in brackets, like `(len(self.payload) - self.payloadOffset))`?

coffeedogs: Indeed, updated.

src/network/bmproto.py
```diff
@@ -334,7 +334,7 @@ def bm_command_object(self):
         self.object = BMObject(nonce, expiresTime, objectType, version, streamNumber, self.payload, self.payloadOffset)
 
         if len(self.payload) - self.payloadOffset > BMProto.maxObjectPayloadSize:
-            logger.info('The payload length of this object is too large (%s bytes). Ignoring it.' % len(self.payload) - self.payloadOffset)
+            logger.info('The payload length of this object is too large (%d bytes). Ignoring it.' % (len(self.payload) - self.payloadOffset))
```
g1itch:
```python
payload_size = len(self.payload) - self.payloadOffset
if payload_size > BMProto.maxObjectPayloadSize:
    logger.info('The payload length of this object is too large (%s bytes). Ignoring it.', payload_size)
```

Please remove percent. `payload_size` or `len(self.payload) - self.payloadOffset` should be second argument of `logger.info()`.

g1itch: I remember landscape.io marked such mistakes.

PeterSurda: I liked landscape.io but the service didn't have a reliable uptime.

coffeedogs: While I agree, this was deliberately the smallest change I could make to make the fix. Can we start an issue to remove all % formatting in log statements throughout the project?

PR #1252, opened 2018-05-16
Fixed: Style and lint issues in src/qidenticon.py

coffeedogs:
Fixed: Style and lint issues in src/qidenticon.py

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -19,3 +19,31 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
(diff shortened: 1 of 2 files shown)


PeterSurda:
I would put `qidenticon.py` into a separate directory and put a license file there

coffeedogs:
I think we should at least mention the existence of further licence files in our LICENSE file then. Wouldn't want someone to get the impression the whole project is MIT. Not that there's anything wrong with the BSD license.

Personally, I wouldn't change our project structure to accommodate derivative works that we are absorbing, we should just make sure that credit is given somewhere. If we were depending on an upstream project directly then a separate directory makes sense.

The only other thing I can think of is to split it off into a different repo, and make it optional at install time but that harms the user experience. Maybe not so much if it was installed by default.

@PeterSurda I'll give it a little while in case my comments inspire any different approaches then move qidenticon and the license to its own directory.

@g1itch I looked at that diff but I don't see much of relevance. What am I supposed to be seeing in the diff please?

g1itch:
MIT and BSD licences is basically the same. As for diff, I mean there is more formatting issues in the `qidenticon` module, and that import string  for `PyQt*` usually is `from PyQt4 import QtCore, QtGui` which demands following changes in the code.

PR #1250, opened 2018-05-15
Fixed: Style and lint violations in src/bitmessageqt/__init__.py

coffeedogs:
Fixed: Style and lint violations in src/bitmessageqt/__init__.py

PR #1256, opened 2018-05-17
flake8: state and depends

g1itch:
Formatted `state` and `depends` according to flake8 suggestions.
From #1189.


src/depends.py
```diff
@@ -1,15 +1,17 @@
...
-    sys.stdout.write('PyBitmessage requires Python 2.7.3 or greater (but not Python 3)')
-    sys.exit()
+    sys.exit(
+        'Python version: %s\n'
+        'PyBitmessage requires Python 2.7.3 or greater (but not Python 3)'
+        % sys.version
+    )
 
 # We can now use logging so set up a simple configuration
 import logging
```
omkar1117: why can't we have this import in the starting?

src/depends.py
```diff
@@ -31,22 +33,28 @@ def check_hashlib():
...
     interface to the most popular hashing algorithms. hashlib
     implements some of the algorithms, however if  OpenSSL
     installed, hashlib is able to use this algorithms as well.
     """
     if sys.hexversion < 0x020500F0:
-        logger.error('The hashlib module is not included in this version of Python.')
+        logger.error(
+            'The hashlib module is not included in this version of Python.')
         return False
     import hashlib
```
omkar1117: Can we have this import in the starting?

src/depends.py
```diff
@@ -58,35 +66,49 @@ def check_sqlite():
...
     if sys.hexversion < 0x020500F0:
-        logger.error('The sqlite3 module is not included in this version of Python.')
+        logger.error(
+            'The sqlite3 module is not included in this version of Python.')
         if sys.platform.startswith('freebsd'):
-            logger.error('On FreeBSD, try running "pkg install py27-sqlite3" as root.')
+            logger.error(
+                'On FreeBSD, try running "pkg install py27-sqlite3" as root.')
         return False
     try:
```
omkar1117: Can we place this in the starting?
 try:
         import sqlite3
     except ImportError:
         logger.error('The sqlite3 module is not available')
         return False

src/depends.py
```diff
@@ -105,10 +128,11 @@ def check_openssl():
...
```
omkar1117: Same here..

  try:
         import ctypes
     except ImportError:
        logger.error(
            'Unable to check OpenSSL. The ctypes module is not available.')
         return False

src/depends.py
g1itch: Because this module tries to import each dependency separately and exit early. @PeterSurda, maybe you can explain how it was designed?

src/depends.py
g1itch: return where? This returns from particular function.

src/depends.py
PeterSurda: @omkar1117 this file is for verifying deployment environment so the imports happen after the checks

src/depends.py
omkar1117: @g1itch , better to place this in a function and call it where it requires

g1itch: OK, I'll try.

g1itch: I figured out that this module duplicates the code of `checkdeps`.

g1itch:
86a077c  breaks #1258.

g1itch:
b35d1af breaks #1250

g1itch:
Recent changes are quite serious. So I done it in several commits. It can be rebased after careful testing.

src/depends.py
```diff
@@ -1,52 +1,199 @@
...
     '%(levelname)s: %(message)s'
 )
 handler = logging.StreamHandler(sys.stdout)
 handler.setFormatter(formatter)
-logger = logging.getLogger(__name__)
+logger = logging.getLogger('both')
 logger.addHandler(handler)
 logger.setLevel(logging.ERROR)
 
+PACKAGE_MANAGER = {
```
omkar1117: Why can't we have all these imports in a json and import where they required.

src/depends.py
```diff
@@ -1,52 +1,199 @@
...
+and suggest how it may be installed
+"""
 
 import sys
-import os
-import pyelliptic.openssl
 
 # Only really old versions of Python don't have sys.hexversion. We don't
 # support them. The logging module was introduced in Python 2.3
 if not hasattr(sys, 'hexversion') or sys.hexversion < 0x20300F0:
```
PeterSurda: During testing android porting, it turned out that it needs at least 2.7.4. 2.7.3 has a bug in struct.unpack which breaks PyBitmessage. I tested it personally with 2.7.3 and 2.7.4 today. 

Historical references: https://bitmessage.org/forum/index.php?topic=4081.0

src/depends.py
```diff
@@ -279,33 +419,35 @@ def check_dependencies(verbose = False, optional = False):
 
     has_all_dependencies = True
 
-    #Python 2.7.3 is the required minimum. Python 3+ is not supported, but it is
-    #still useful to provide information about our other requirements.
+    # Python 2.7.3 is the required minimum. Python 3+ is not supported,
+    # but it is still useful to provide information about our other
+    # requirements.
     logger.info('Python version: %s', sys.version)
     if sys.hexversion < 0x20703F0:
```
PeterSurda: Should be 0x2074F0. See my previous comment.

src/depends.py
```diff
@@ -279,33 +419,35 @@ def check_dependencies(verbose = False, optional = False):
...
-        logger.error('PyBitmessage requires Python 2.7.3 or greater (but not Python 3+)')
```
PeterSurda: Should say "2.7.4".

src/depends.py
```diff
@@ -279,33 +419,35 @@ def check_dependencies(verbose = False, optional = False):
...
+        logger.error(
+            'PyBitmessage requires Python 2.7.3 or greater'
+            ' (but not Python 3+)')
         has_all_dependencies = False
     if sys.hexversion >= 0x3000000:
-        logger.error('PyBitmessage does not support Python 3+. Python 2.7.3 or greater is required.')
+        logger.error(
+            'PyBitmessage does not support Python 3+. Python 2.7.3'
```
PeterSurda: should say "2.7.4"

src/depends.py
g1itch: 2.3 is checked here to exit before `import logging`. I think it makes sense because 2.7.4 is checked below.

checkdeps.py
```diff
@@ -9,79 +13,8 @@
...
-        "Ubuntu": "python-setuptools",
-        "Ubuntu 12": "python-setuptools",
-        "Fedora": "python2-setuptools",
-        "openSUSE": "python-setuptools",
-        "Guix": "python2-setuptools",
-        "Gentoo": "",
-        "optional": False,
-    }
-}
+from src.depends import detectOS, PACKAGES, PACKAGE_MANAGER
```
PeterSurda: :+1:

checkdeps.py
```diff
@@ -182,27 +78,27 @@ def testCompiler():
...
 
 if not compiler:
     print "Building the bitmsghash module failed.\n" \
         "You may be missing a C++ compiler and/or the OpenSSL headers."
 
 if prereqs:
-    mandatory = list(x for x in prereqs if "optional" not in PACKAGES[x] or not PACKAGES[x]["optional"])
-    optional = list(x for x in prereqs if "optional" in PACKAGES[x] and PACKAGES[x]["optional"])
+    mandatory = [x for x in prereqs if not PACKAGES[x].get("optional")]
+    optional = [x for x in prereqs if PACKAGES[x].get("optional")]
```
PeterSurda: :+1:

src/depends.py
PeterSurda: This is a special case, this file is for verifying what modules are available in the system so that it can take compensating measures or fail gracefully while explaining to the user what he has to do.

PeterSurda:
Please resolve conflicts, other than that I'm ready to merge.

g1itch:
I pass, have no idea how to resolve it. `bitmessageqt` looks broken for me after #1250.

g1itch:
I reverted any changes to `bitmessageqt.__init__` while resolved the conflict.

PR #1257, opened 2018-05-17
flake8: bitmessagemain, class_singleCleaner, helper_generic, shutdown ...

g1itch:
Formatted `bitmessagemain`, `class_singleCleaner`, `helper_generic`, `shutdown` and `singleinstance` according to flake8 suggestions.
From #1189.

### DIFF
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python2.7
 # Copyright (c) 2012-2016 Jonathan Warren
-# Copyright (c) 2012-2016 The Bitmessage developers
+# Copyright (c) 2012-2018 The Bitmessage developers
 # Distributed under the MIT/X11 software license. See the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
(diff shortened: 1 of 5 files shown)


src/bitmessagemain.py
```diff
@@ -16,10 +16,12 @@
 os.chdir(app_dir)
 sys.path.insert(0, app_dir)
 
+
 import depends
 depends.check_dependencies()
```
omkar1117: Please write imports in a order.
standard imports first.
from statements next.

custom from statements in the end.

src/helper_generic.py
```diff
@@ -1,21 +1,22 @@
...
-Helper Generic perform generic oprations for threading.
+Helper Generic perform generic operations for threading.
 
 Also perform some conversion operations.
 """
 
+
 import socket
 import sys
-from binascii import hexlify, unhexlify
```
omkar1117: Here also place imports in a order.

src/shutdown.py
```diff
@@ -8,52 +8,67 @@
 from helper_threading import StoppableThread
 from knownnodes import saveKnownNodes
```
omkar1117: Place imports in a order

src/bitmessagemain.py
PeterSurda: This is a special situation as `depends` checks which modules are available and handles missing modules gracefully. It needs to run before other imports.

PeterSurda: Also, since this is the main program function, there are no problems with potential recursive imports and things like that.

src/helper_generic.py
PeterSurda: multiprocessing is special, importing it in wrong order could cause some initialisation to be executed in wrong order. The initialisation has to be fixed first.

src/shutdown.py
PeterSurda: importing shared and state in wrong order could trigger bugs.

PR #1258, opened 2018-05-19
Added: Support installing system dependencies of optional components

coffeedogs:
Not quite ready to merge but I'm inviting feedback at this stage regarding the approach and scope as well as requesting any first-hand, OS-specific data about prctl (or other packages).

Assume I already know the join comprehension needs turning inside out ;)


checkdeps.py
```diff
@@ -1,4 +1,13 @@
...
+Limitations:
+
+    * Does not detect whether packages are already installed. Solving this requires writing more of a configuration
+    management system. Or we could switch to an existing one.
+    * Not fully PEP508 compliant. Not slightly. It makes bold assumptions about the simplicity of the contents of
+    EXTRAS_REQUIRE. This is fine because most developers do, too.
+"""
 
 from distutils.errors import CompileError
 try:
```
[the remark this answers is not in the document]
coffeedogs: I'll catch all that when I do the code style and linting changes. Since we're progressing through doing this to the files as a parallel task to this fix it would be unwise to conflate the two very distinct types of work.

coffeedogs:
I mean, I like that slightly nested comprehension. If other people don't mind it this is ready to go. Unless we want to extend the scope of the task to check for installed packages.

As alluded to in code comments, we could share thoughts on configuration management systems before committing to extend our home-grown version.

checkdeps.py
```diff
@@ -199,13 +231,36 @@ def testCompiler():
...
+            ' ',
+            ' '.join([
+                ''. join([
+                    xx for xx in EXTRAS_REQUIRE_DEPS[x][OPSYS]
+                ])
+                for x in rhs
+                if x in EXTRAS_REQUIRE_DEPS
+            ]),
+        ])
+        print "Optional dependency `pip install .[{}]` would require `{}` to be run as root".format(lhs, rhs_cmd)
```
g1itch: I got lost in this expression ):

requirements.txt
```diff
@@ -1 +0,0 @@
-python_prctl
```
g1itch: And this will appear again if #1221 is acceptable? ...

setup.py
```diff
@@ -1,13 +1,24 @@
...
 from setuptools.command.install import install
 
 from src.version import softwareVersion
 
+
+EXTRAS_REQUIRE = {
+    'gir': ['pygobject'],
+    'notify2': ['notify2'],
+    'pyopencl': ['pyopencl'],
+    'prctl': ['python_prctl'],  # Named threads
```
g1itch: Where are you going to use this feature?

checkdeps.py
g1itch: Also I've used [print](https://docs.python.org/3/library/functions.html#print) which is available in python 2.7 because it's a slow preparation to python3.

setup.py
PeterSurda: prctl is needed so that the OS can see thread names. If you run `top -H`, it will show thread names assuming prctl is installed, otherwise all will be called `python` or `pybitmessage`.

requirements.txt
coffeedogs: Indeed, #1221 is now showing as conflicting on requirements.txt and setup.py. Could you rebase once this PR is in and use an optional dependency in setup.py rather than requirements.txt please?

checkdeps.py
coffeedogs: OK, I will unroll it in the next commit.

coffeedogs: Ah, already merged. Can you open an issue please and I'll get round to it.

g1itch: That's a matter of taste. I won't mind if @PeterSurda accepts.

requirements.txt
g1itch: The method with pip you've proposed will work only for pip and incompatible with plain setuptools. What should I write into `tests_require` in that case?

setup.py
coffeedogs: OK, I'm reading up, making sure I understand what your needs are here and considering a solution.

requirements.txt
coffeedogs: I'm not sure what you mean by not compatible with setuptools. We can still do `python setup.py install` and our package is installed. It still has no dependencies. If we want to depend on `prctl` we could list it in install_requires.

So `prctl` is an optional component. Maybe it could be in an extras_require category like 'enhancements' and that be strongly recommended when we check the dependencies? Bear in mind the task here was to extend our existing dependency checking to include OS-specific system packages dependencies of optional dependencies.

What is your suggestion for an alternative approach, maybe pipenv? I've read some more about python packaging (there seems to be an infinite amount of it!) but it all suggests that `pip install .[category]` after `python setup.py install` is the way to go. `pip` deals with installing a concrete environment rather than an abstract dependency and altering os thread naming by installing an optional package sounds environmental to me.

As for requirements for tests, I have no opinion. Either we can have `tests_require` specify the abstract dependencies so that `python setup.py test` installs the test deps and runs the tests, or we can have an extras_require named 'test' and do something like `pip install .[test]; pytest`. Or we can do both - define the list at the top of setup.py and use it in both places.

setup.py
coffeedogs: OK, I'm reading up, making sure I understand what your needs are here and considering a solution.

coffeedogs: I have opened issue#1269 to address this.

PR #1259, opened 2018-05-23
enable delayed POW checks

f97ada87:
Currently, function `isProofOfWorkSufficient` checks the object EOL against the current time of the computer running the object processor. This can lead to false positives or negatives if the computer time is significantly different to the object receive time. 

This PR adds an optional "receive time" argument to function `isProofOfWorkSufficient` to enable correct validation of such objects received from the past or future. The change is 100% backward compatible.

### DIFF
--- a/src/protocol.py
+++ b/src/protocol.py
@@ -168,13 +168,14 @@ def checkSocksIP(host):
 
 def isProofOfWorkSufficient(data,
                             nonceTrialsPerByte=0,
-                            payloadLengthExtraBytes=0):
+                            payloadLengthExtraBytes=0,
+                            recvTime=0):
     if nonceTrialsPerByte < defaults.networkDefaultProofOfWorkNonceTrialsPerByte:
         nonceTrialsPerByte = defaults.networkDefaultProofOfWorkNonceTrialsPerByte
     if payloadLengthExtraBytes < defaults.networkDefaultPayloadLengthExtraBytes:
         payloadLengthExtraBytes = defaults.networkDefaultPayloadLengthExtraBytes
     endOfLifeTime, = unpack('>Q', data[8:16])
-    TTL = endOfLifeTime - int(time.time())
+    TTL = endOfLifeTime - (recvTime if recvTime else int(time.time()))
     if TTL < 300:
         TTL = 300
     POW, = unpack('>Q', hashlib.sha512(hashlib.sha512(data[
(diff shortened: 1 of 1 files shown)


src/protocol.py
```diff
@@ -168,13 +168,14 @@ def checkSocksIP(host):
 
 def isProofOfWorkSufficient(data,
                             nonceTrialsPerByte=0,
-                            payloadLengthExtraBytes=0):
+                            payloadLengthExtraBytes=0,
+                            recvTime=0):
```
g1itch: Are you planning to write some code which will call this `isProofOfWorkSufficient()` with the `recvTime` kwarg? Now it's hard to imagine how this changes supposed to work.

f97ada87:
Certainly more code changes are needed in order to take advantage of this new feature. Some of these changes are already written and ready to go. 
I am just trying to avoid a large diff that would lead to a long review time on my PR, with the ensuing complications. From prior experience, small increments seem to work better for everyone.

src/protocol.py
f97ada87: Certainly more code changes are needed in order to take advantage of this new feature. Some of these changes are already written and ready to go.
I am just trying to avoid a large diff that would lead to a long review time on my PR, with the ensuing complications. From past experience, small increments seem to work better for everyone.

PeterSurda: this could be helpful for airgapped machines, or for running the network thread and objectProcessorThread in different processes/machines in general.

f97ada87: This is the original purpose indeed, and then I noticed that object time travel may be possible even in the current version, in rare circumstances. For example, when an objectProcessorQueue backlog is persisted to sqlite on client shutdown: 
https://github.com/Bitmessage/PyBitmessage/blob/74ddda8775156d6b4271284b6902058513c5d3d5/src/class_objectProcessor.py#L102
If the client is restarted after a clock change (or a few days break), the persisted objects will return a false status on this check: 
https://github.com/Bitmessage/PyBitmessage/blob/74ddda8775156d6b4271284b6902058513c5d3d5/src/class_objectProcessor.py#L618
Again, unlikely circumstances but not impossible.

coffeedogs: Could you add a parameter type definition in the docstring please? It's not obvious what kind of time object you want to be passed in. I believe you're expecting a float representing unix time rather than a python time/datetime type.

PeterSurda: Good idea, @f97ada87 please add docstring.

f97ada87: Docstring added; also an integer cast to match the treatment given to time.time()

PR #1264, opened 2018-05-28
Fixed: Style and lint violations for three more of the worst violatin…

coffeedogs:
Fixed: Style and lint viloations for three more of the worst violating files


src/bitmessageqt/settings.py
```diff
@@ -1,16 +1,24 @@
...
+from __future__ import absolute_import
 
-from PyQt4 import QtCore, QtGui
-from languagebox import LanguageBox
 from sys import platform
 
+from PyQt4 import QtCore, QtGui
+
+from . import bitmessage_icons_rc  # pylint: disable=unused-import
+from .languagebox import LanguageBox
```
g1itch: This imports are breaking bundled usage

PeterSurda: You mean the frozen mode?

PR #1268, opened 2018-06-11
Added: Fabric features

coffeedogs:
## default_hosts decorator

Until now it was necessary to add `-H localhost` to fab tasks. This was cumbersome. Now tasks can have a list of default hosts defined that will be used if the `-H` option is not given. Mostly that means adding `@default_hosts(['localhost'])` to the task definition.

## filelist from git diff feature

Until now the `code_quality` task operated on all files, the worst offending or a specific file. This feature adds the option to check just the files that have changed in the working directory, git's staging area, or since a particular revision.

## task to run tests

The aim of this task is to provide a simple and consistent way of running tests. As we develop more tests, and types of tests, this might mean doing a 'quick' run or a 'full' run or something more specific. Currently it runs the same test command as travis, plus runs the tests that @g1itch has written (these are pending merge so currently no tests are found when running the task).


PeterSurda:
Please resolve merge conflicts.

PR #1263, opened 2018-05-24
Added: Sphinx docs, testing readthedocs integration

coffeedogs:
* Added: Sphinx configuration
 * Added: Sphinx fabric task to auto-document the project
 * Added: Manual documentation
 * Fixed: RsT formatting in docstrings
 * Fixed: Some adjacent minor style and lint issues

~~This is not ready to merge, there are around 80 warnings, mostly 'unable to import logger' so some modules are not documented.
Also, it seems it needs re-basing.~~

OK, so I think we'll get the import errors fixed with upcoming refactoring of imports. Let's get this in now because it gives everyone exposure to the docs and the build process. They can only get better from this point onwards!

PR #1271, opened 2018-06-13
Codeq31 05

coffeedogs:
This PR obsoletes 1265. Outstanding issues (not addressed in this PR):

 * ~~import re-org may break portable use~~
 * ~~settings.py is autogenerated~~
 * ~~Docstrings TBC~~
 * ~~Trailing braces style at proofofwook:143~~
 * ~~from struct import pack (order?)~~
 * ~~Best place for copyright notices?~~


src/bitmessageqt/account.py
```diff
@@ -1,26 +1,58 @@
...
+from __future__ import absolute_import
+
+import inspect
 import re
 import sys
-import inspect
-from helper_sql import *
-from helper_ackPayload import genAckPayload
+import time
+
```
omkar1117: Can we have imports in a order

src/upnp.py
```diff
@@ -1,21 +1,33 @@
...
+"""
+A simple upnp module to forward port for BitMessage
+Reference: http://mattscodecave.com/posts/using-python-and-upnp-to-forward-a-port
+"""
+
+from __future__ import absolute_import
+
 import httplib
 from random import randint
 import socket
```
omkar1117: place imports in a order please.

coffeedogs: Not with __future__, no.

src/bitmessageqt/account.py
coffeedogs: Sure.

src/bitmessagemain.py
```diff
@@ -1,49 +1,55 @@
...
-from time import sleep
-from random import randint
-import getopt
 
 from api import MySimpleXMLRPCRequestHandler, StoppableXMLRPCServer
 from helper_startup import (
     isOurOperatingSystemLimitedToHavingVeryFewHalfOpenConnections
 )
 
 import defaults
```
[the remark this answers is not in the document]
coffeedogs: Sorry, this PR is a work in progress, as per the one it replaced. I needed a PR# to reference in the old ticket. If I close this one I'll have to open another PR since I don't have permission to re-open in such cases. Just hold off reviewing this on until I mark it as ready for review. There's three other PRs #1263, #1264 and #1268 that need review though.

src/bitmessagemain.py
```diff
@@ -72,7 +78,12 @@
 import helper_threading
 
 
+depends.check_dependencies()
+
+
 def connectToStream(streamNumber):
+    """Connect to a stream"""
```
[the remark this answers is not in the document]
omkar1117: I think rather than sphinx we can update manually when we are customizing it, it would ne more realistic I hope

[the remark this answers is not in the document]
coffeedogs: Well sphinx is coming in with #1263 and it already produces html docs from the code, including function signatures and parameter documentation. It supports `:param foo: blah` syntax as well as type hints and other comment styles. We should decide on a convention.

PeterSurda:
Please resolve merge conflicts.

src/bitmessagemain.py
```diff
@@ -1,49 +1,55 @@
...
+from subprocess import call
+from time import sleep
 
 app_dir = os.path.dirname(os.path.abspath(__file__))
 os.chdir(app_dir)
 sys.path.insert(0, app_dir)
 
-
-import depends
-depends.check_dependencies()
```
g1itch: Did you test this?

src/bitmessageqt/account.py
```diff
@@ -1,26 +1,58 @@
...
+            unicode(
+                BMConfigParser().get(
+                    x,
+                    'label'),
+                'utf-8').lower(),
+            unicode(
+                BMConfigParser().get(
+                    y,
+                    'label'),
+                'utf-8').lower()))
```
g1itch: Are you joking?

src/bitmessageqt/bitmessageui.py
```diff
@@ -1,20 +1,29 @@
...
-from networkstatus import NetworkStatus
-from blacklist import Blacklist
+from . import bitmessage_icons_rc  # pylint: disable=unused-import
+from . import settingsmixin
+from .messageview import MessageView
+from .messagecompose import MessageCompose
+from .networkstatus import NetworkStatus
+from .blacklist import Blacklist
+from .foldertree import AddressBookCompleter
+
```
g1itch: My usual reminder about dot imports here...

src/bitmessageqt/bitmessageui.py
```diff
@@ -744,9 +813,17 @@ def retranslateUi(self, MainWindow):
...
+        self.tabWidget.setTabText(
+            self.tabWidget.indexOf(
+                self.blackwhitelist),
+            _translate(
+                "blacklist", "Blacklist", None))
         self.networkstatus.retranslateUi()
-        self.tabWidget.setTabText(self.tabWidget.indexOf(self.networkstatus), _translate("networkstatus", "Network Status", None))
+        self.tabWidget.setTabText(
+            self.tabWidget.indexOf(
+                self.networkstatus),
```
g1itch: lol

src/bitmessageqt/settings.py
```diff
@@ -1,16 +1,24 @@
 # -*- coding: utf-8 -*-
```
g1itch: Why do you need a second pass for this module? Something similar was already merged.

src/bitmessageqt/settings.py
```diff
@@ -1,16 +1,24 @@
...
-#
-# Created: Thu Dec 25 23:21:20 2014
-#      by: PyQt4 UI code generator 4.10.3
-#
-# WARNING! All changes made in this file will be lost!
+Created: Thu Dec 25 23:21:20 2014
+     by: PyQt4 UI code generator 4.10.3
+
+WARNING! All changes made in this file will be lost!
+"""
```
g1itch: It's really going to be a funny documentation (`:

src/namecoin.py
```diff
@@ -1,54 +1,64 @@
-# Copyright (C) 2013 by Daniel Kraft <d@domob.eu>
```
g1itch: Again. It was merged already.

src/network/asyncore_pollchoose.py
```diff
@@ -75,13 +73,16 @@
 except (ImportError, AttributeError):
     WSAECONNRESET = ECONNRESET
 try:
-    from errno import WSAEADDRINUSE
+    # side-effects on Windows or cruft?
+    from errno import WSAEADDRINUSE  # pylint: disable=unused-import
 except (ImportError, AttributeError):
     WSAEADDRINUSE = EADDRINUSE
+import helper_random
```
g1itch: Why it was moved here?

src/network/asyncore_pollchoose.py
```diff
@@ -707,13 +782,16 @@ def __getattr__(self, attr):
...
 
     def log(self, message):
+        """Log a message to stderr"""
         sys.stderr.write('log: %s\n' % str(message))
 
     def log_info(self, message, log_type='info'):
+        """Conditionally print a message"""
         if log_type not in self.ignore_log_types:
-            print('%s: %s' % (log_type, message))
+            print '%s: %s' % log_type, message
```
g1itch: Did you tested? It seems to not work.
```
In [1]: log_type = 'foo'

In [2]: message = 'bar'

In [3]: print '%s: %s' % log_type, message
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-e799024119aa> in <module>()
----> 1 print '%s: %s' % log_type, message

TypeError: not enough arguments for format string

In [4]: 
```

src/bitmessagemain.py
coffeedogs: Yes. The import has moved up and the function call moved down.

src/bitmessagemain.py
```diff
@@ -1,49 +1,55 @@
...
+
+import ctypes
+import errno
+import getopt
 import os
+import signal
+import socket
 import sys
+import threading
+from random import randint
```
omkar1117: please give a space between import and from statements.

src/bitmessageqt/account.py
coffeedogs: Guessing you don't like the formatting? I was just following pylint's indentation. The whole BMConfigParser is a copy-pasta boiler-plate abomination that needs ripping out at some point. If you confirm that this style is your issue I'll be happy to change it, as usual.

src/bitmessagemain.py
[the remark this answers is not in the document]
omkar1117: @coffeedogs any update here?

src/bitmessageqt/settings.py
coffeedogs: As discussed elsewhere, our process involves manually editing this file which was auto-generated once, five years ago. This comment will therefore be removed since it is misleading. If you want to suggest re-generating this file as part of the build process then please either open an issue or PR.

src/bitmessageqt/settings.py
coffeedogs: Indeed these changes are already in.

src/bitmessageqt/bitmessageui.py
coffeedogs: Made imports absolute, tested both installed and portable methods.

src/network/asyncore_pollchoose.py
coffeedogs: Recall this commit was a WIP. All imports now automatically sorted. If you have any import sorting issues from now on you will have to find suitable isort settings. There will be no more manual import sorting.

src/network/asyncore_pollchoose.py
coffeedogs: Typo fixed in next commit.

src/bitmessagemain.py
coffeedogs: Personally, I'm not in favour of such a formatting option. Neither is isort by default. It may be possible to get isort to do this but we should invite more general discussion about non-violating import formatting issues and update the isort config accordingly.

g1itch: No need for space. PEP8 requires spaces between groups, e.g. system, libs, internal imports, not `import ..` and `from .. import`.

src/socks/__init__.py
```diff
@@ -51,46 +49,71 @@
...
-    "host unreachable")
+                  "invalid data",
+                  "not connected",
+                  "not available",
+                  "bad proxy type",
+                  "bad input",
+                  "timed out",
+                  "network unreachable",
+                  "connection refused",
+                  "host unreachable")
```
g1itch:
```python
_generalerrors = (
    "success",
    "invalid data",
    ...
)
```

coffeedogs: I'm also a fan of hanging indentation, though our style is not consistent throughout the project. I've started creating a style guide for summarising our discussion for style ietms like this that are beyond pep8 but before manually changing one or two instances I would rather update the `fab code_quality` task to ensure the rule is enforced by autopep8 consistently.

If you want hanging indents to be the norm for autopep8 then we should (at a suitable time and after discussion) try updating setup.cfg and running code_quality:fix=1 on all affected files.

coffeedogs:
> I disagree with that # pylint: disable= and ignore = in setup.cfg. It seems you are going to shoot our legs, again.

If you mean 'shoot ourselves in the foot' please explain how. Lint markers are necessary for the moment. They indicate places where we still need to improve the code, opportunities for refactoring. Having them in allows us to perform continuous delivery without regressing.

Adding lint markers to the code and temporarily listing violations in setup.cfg is normal, recommended practice.

src/socks/__init__.py
PeterSurda: The socks directory is mostly decommissioned, no need to look at its code quality. If someone could instead move the SOCKS DNS lookups to the asyncore thread, that should solve the issue. I did implement SOCKS DNS lookups in asyncore but never tested or integrated them. Perhaps it's not even necessary to support SOCKS DNS lookups, and instead keep the bootstrap DNS record in knownnodes, and update parts of the code which guess whether the record is an IPv4/IPv6/onion.

src/bitmessagecli.py
```diff
@@ -13,18 +13,17 @@
...
 import socket
-import time
 import sys
-import os
-
-from bmconfigparser import BMConfigParser
+import time
+import xmlrpclib
 
+from pybitmessage.bmconfigparser import BMConfigParser
```
g1itch: I'm not sure. Maybe `from .bmconfigparser import BMConfigParser` would be better.

checkdeps.py
```diff
@@ -10,8 +11,14 @@
     EXTRAS_REQUIRE. This is fine because most developers do, too.
 """
 
+from __future__ import print_function
```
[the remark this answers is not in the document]
coffeedogs: Hmm, obviously you meant `from __future__ import absolute_import` rather than `from __future__ import print_function`. Which I'm sure I saw earlier here. Maybe rebasing made git diff confused. Anyway, absolute_imports removed (not needed if we don't support Python 2.6 as you say), while the print_function remains (since you've said you're interested in starting to think about Python3).

src/bitmessagecli.py
coffeedogs: No point in relative imports any more. Setting the path and cwd has been fixed to work better running from both standalone and installed locations.

coffeedogs:
The [lanscape link](https://landscape.io/github/Bitmessage/PyBitmessage/pulls/1271) gives me a 500 error :/

coffeedogs:
Please test, final review, merge. Thanks.

checkdeps.py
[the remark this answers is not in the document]
g1itch: Same true for print_function:
https://docs.python.org/2/library/__future__.html

g1itch:
`desktop/icons` in the same time in `.gitignore` and in this PR

checkdeps.py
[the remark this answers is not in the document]
coffeedogs: Keeping this prevents regressions in style and enforces the newer syntax: https://stackoverflow.com/questions/28694380/pylint-says-unnecessary-parens-after-r-keyword

PeterSurda:
@coffeedogs landscape seems to be broken since about a year ago, that's why I switched to codacy, maybe there is some half-alive leftover glue.

src/bitmessagecli.py
g1itch: but you keep removing the parens...

checkdeps.py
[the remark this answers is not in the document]
g1itch: but you keep removing the parens...

src/bitmessagecli.py
g1itch: The form
```python
from pybitmessage.bmconfigparser import BMConfigParser
```
will not work for standalone use-case because python wont find the package `pybitmessage`.

coffeedogs:
> desktop/icons in the same time in .gitignore and in this PR

Removed that line in .gitignore, not sure how it ended up there, I was probably looking to exclude generated files.

g1itch:
> Removed that line in .gitignore, not sure how it ended up there, I was probably looking to exclude generated files.

These are indeed "generated" files: [look](../blob/v0.6/setup.py#L29). It was made for indicator_libmessaging.

checkdeps.py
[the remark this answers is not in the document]
coffeedogs: If there was only one or two instances of parens but lots of statements I removed the statements rather than adding the import and enforcing the whole file. If there were mostly parens in a file I changed the print statements to functions and enforced this at the top to avoid regression. I wasn't about to enforce print functions throughout the project. If there's any project-wide tidying up around print I would prefer to change print statements to use logging or sys.stdout.write() as appropriate. That was outside the scope of the task I was undertaking, however.

src/bitmessagecli.py
coffeedogs: Changed in recent commit to `from bmconfigparser import BMConfigParser` which works for me under all test scenarios listed in `PULL_REQUEST_TEMPLATE.md`, which I updated as well.

coffeedogs:
@g1itch could you please comment in a way I can reply to you where possible (guessing it's because it's a general comment rather than a line comment or maybe it's my rebasing, in which case I apologise), that way the convo can be followed more easily.

So you're saying I was right in the first place to exclude the generated (copied) files. I'll revert that `.gitignore` change then...

g1itch:
@coffeedogs 

> So you're saying I was right in the first place to exclude the generated (copied) files.

yes, better keep it in .gitignore though I see no problem to have some untracked files in git dir.

coffeedogs:
> yes, better keep it in .gitignore though I see no problem to have some untracked files in git dir.

The main problems I find with it are:

 * Manually, mentally removing certain ignores every time I do git status is a poor workflow
 * `git diff` is used by our fabric tasks (against cwd, staged as well as revisions) so that would need further handling, duplicating the job of .gitignore
 * Having files outstanding causes some git commands to return different exit status, ruining our ability for straightforward CI
 * This is the very point of .gitignore

src/bitmessageqt/bitmessageui.py
```diff
@@ -1,20 +1,24 @@
...
-#
-# WARNING! All changes made in this file will be lost!
+Originally generated from reading ui file 'bitmessageui.ui'. Since then maintained manually.
+"""
+
+import sys
 
 from PyQt4 import QtCore, QtGui
+
+from bitmessageqt import bitmessage_icons_rc  # pylint: disable=unused-import
```
omkar1117: from bitmessageqt import settingsmixin, bitmessage_icons_rc  # pylint: disable=unused-import

src/proofofwork.py
```diff
@@ -96,41 +107,56 @@ def _doFastPoW(target, initialHash):
...
     nonce = openclpow.do_opencl_pow(initialHash.encode("hex"), target)
-    trialValue, = unpack('>Q',hashlib.sha512(hashlib.sha512(pack('>Q',nonce) + initialHash).digest()).digest()[0:8])
-    #print "{} - value {} < {}".format(nonce, trialValue, target)
+    trialValue, = unpack('>Q', hashlib.sha512(hashlib.sha512(pack('>Q', nonce) + initialHash).digest()).digest()[0:8])
     if trialValue > target:
         deviceNames = ", ".join(gpu.name for gpu in openclpow.enabledGpus)
-        queues.UISignalQueue.put(('updateStatusBar', (tr._translate("MainWindow",'Your GPU(s) did not calculate correctly, disabling OpenCL. Please report to the developers.'), 1)))
-        logger.error("Your GPUs (%s) did not calculate correctly, disabling OpenCL. Please report to the developers.", deviceNames)
+        queues.UISignalQueue.put((
+            'updateStatusBar', (
```
omkar1117: PEP8 validtion

src/bitmessageqt/bitmessageui.py
coffeedogs: I was concerned there might have been side-effects from this import. After a closer look I don't think there is. Removed.

src/proofofwork.py
coffeedogs: Not sure what issue this may have been but it passes all the checks now

PeterSurda:
@g1itch are you happy with the way it is now or you recommend more changes?

omkar1117:
I have verified the functionality on Two Ubuntu machines with PR-1271 build and working as expected. Please find the below steps i have done for the same.

1. Upgraded Two Ubuntu machines with PR-1271 build
2. Sent Bitmessage from machine1 to machine2 and vice-versa.
3. verified the received Bitmessage on both the machines. It is workingas expected.
4. this testing was done on  two machines  with same network not on braodcost machines(more than two machines).
5. Verified with Chans and with out Chans, it is working as expected.
6. Verified the Network status page, showing all the required information in that page.
7. Bitmessage was sent by two ways on both the machines. 1. send ordinary machine and 2. Send message to your subscribers.
8. Selected "Use a random number generator to make an address" radio button for Bitmessage address.

@coffeedogs , @PeterSurda  please let me know if I miss any functional flow here.

g1itch:
@PeterSurda No. I'll repeat my objections:

1. Strings in the beginning, "Copyright (c) 2012-2016 Jonathan Warren ..." are not a documentation, - it's copyright strings. They not contain any useful information to be included into documentation, on the contrary, they misinform, there is no software version in shared.py. Docstring in setup.py is complete nonsense.
1. Expressions like `from __future__ import absolute_import` are for python 2.6 which is not supported by PyBitmessage.
1. print functions was added intentionally and I see no sense to revert it to print op.
1. `depends.check_dependencies()` should be called in the very beginning, otherwise the `depends` module makes no sense.
1. You can simply turn off any pylint check instead of adding that `pylint: disable=...` in the beginning of each module.
1. Changes like `except: -> except BaseException:` resolve lint issues just formally and make future debugging harder.
1. This changeset contains  `desktop/icons/` made by setup.py which is also in `.gitignore`
1. The changeset also includes many changes made before this PR because of multiple merges with v0.6.

PeterSurda:
Thank you @g1itch 

1. I agree that copyright and docstring should be separated. I don't know the best way to do this, I'll leave it up to @coffeedogs. The documentation should also be updated to reflect that the version is in `version.py` rather than `shared.py`

2. I don't know, but looks like @coffeedogs agrees that it's only for python 2.6. Python 2.7.4 is the lowest supported version (or you have to use a patched 2.7.3). So remove.

3. I don't care

4. I concur

5. I am ok with the suppressing pylint warnings in particular lines where the issue either can't be fixed or makes no sense (e.g. unused-argument). They aren't needed but if they're in, I'm ok. However the per-class / per-file suppressions like too-many-locals or too-many statements should be removed, it's better if they warnings remain active. These are better addressed through refactoring at a later stage

6. I concur

7. I concur

8. I concur

@coffedogs please make the changes requested

coffeedogs:
>Strings in the beginning, "Copyright (c) 2012-2016 Jonathan Warren ..." are not a documentation, - it's copyright strings. They not contain any useful information to be included into documentation, on the contrary, they misinform, there is no software version in shared.py. Docstring in setup.py is complete nonsense.
>>I agree that copyright and docstring should be separated. I don't know the best way to do this, I'll leave it up to @coffeedogs. The documentation should also be updated to reflect that the version is in version.py rather than shared.py
I've removed the comment about version as it is not used in this file. I did not introduce it, merely moved it to the module docstring. I will create docs/licenses.rst, move all copyright mentions to there and mention its existence in LICENCE. The docstring in setup.py might be a good place to document our plugin system, alternative invocations of setup.py, alternative installation / standalone mode. I don't see anything that is 'complete nonsence' in there, besides the version line that I have removed.

>Expressions like from __future__ import absolute_import are for python 2.6 which is not supported by PyBitmessage.
>>I don't know, but looks like @coffeedogs agrees that it's only for python 2.6. Python 2.7.4 is the lowest supported version (or you have to use a patched 2.7.3). So remove.
I agree python 2.7 doesn't require it, though it does permit it. By importing `print_function` on a per-file basis we are able to avoid regressions (people using the older print statements) since doing so becomes invalid syntax. But if there's a better way to migrate to print_function I'm open to hearing about it.

>print functions was added intentionally and I see no sense to revert it to print op.
>>I don't care
I suspect this was a file where there was a couple of `print <tuple>` statements in a file with way more print <string> statements. Pylint recognises that [you think you're using the print function](https://docs.python.org/2/library/functions.html#print) but you haven't imported it from __future__ at the top. So pylint asks you to commit to a print statement or a print function.

>depends.check_dependencies() should be called in the very beginning, otherwise the depends module makes no sense.
>>I concur
The very idea of a runtime dependency check here (as opposed to during install, build or exe bundling) is unpleasant to me. I have updated setup.cfg to instruct isort to keep it in place.

>You can simply turn off any pylint check instead of adding that pylint: disable=... in the beginning of each module.
>>I am ok with the suppressing pylint warnings in particular lines where the issue either can't be fixed or makes no sense (e.g. unused-argument). They aren't needed but if they're in, I'm ok. However the per-class / per-file suppressions like too-many-locals or too-many statements should be removed, it's better if they warnings remain active. These are better addressed through refactoring at a later stage
Interesting, Dmitri is suggesting for the lint exceptions to be more general and quiet and Peter is suggesting more specific and noisy, :) I am removing such file and class-level lint markers.

>Changes like except: -> except BaseException: resolve lint issues just formally and make future debugging harder.
>>I concur
I've not done that consciously. I've removed `as e` where it wasn't used. Ah, autopep8 replaces a bare except with except BaseException. There's discussion [here](https://github.com/hhatto/autopep8/issues/297) about this this. I can't see how this would affect debugging; if any BaseException is raised it will be there in the stack trace. If you're sure I can undo those changes.

>This changeset contains desktop/icons/ made by setup.py which is also in .gitignore
>>I concur
Removed; I clearly swung the `git add` hammer too wildly.

>The changeset also includes many changes made before this PR because of multiple merges with v0.6.
>>I concur
Rebasing...

PR #1272, opened 2018-06-14
Gitignore : updated gitignore

omkar1117:
Pycharm idea ignore files added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ build/lib.*
 build/temp.*
 dist
 *.egg-info
+
+# Pycharm idea files excluded
+.idea/*


omkar1117:
@coffeedogs  this PR is for the developers who are suppose to use Pycharm IDE, this will reject developers .idea/ folder and their xml files.

coffeedogs:
I understand what it it for, but it is not the responsibility of the project to dictate the developer's IDE, nor to support exclusions of cruft for all IDEs. If you use pycharm, adding an exclusion at the user level for pycharm files is the best place to configure it. It then applies to all your projects without any projects needing to add it specifically. The only files that should be specified in the project's `.gitignore` are files that are generated by running the project, installing it, testing it, building docs etc.

PeterSurda:
I agree with @g1itch and @coffeedogs . You should exclude these locally, not in the project.

PR #1278, opened 2018-06-15
Fixed: Revert changes from boolean operators to any()/all().

coffeedogs:
Bug: The actual bug was introduced in the former line 2585 and fixed in the later 2579: `elif treeWidget == all(` instead of `elif all( treeWidget ==`

Style change: I'm agnostic to any preference on style, though I had to adjust indention to satisfy the linters and move the operators to the end of the previous line.

Performance: Boolean operators short-circuit more often and so indeed are better performing. We could also look towards reducing the number of repeated calls in this code to better improve performance.

### DIFF
--- a/src/bitmessageqt/__init__.py
+++ b/src/bitmessageqt/__init__.py
@@ -2509,18 +2509,14 @@ class MyForm(settingsmixin.SMainWindow):  # pylint: disable=too-many-public-meth
             treeWidget = self.widgetConvert(sent)
             if self.getCurrentFolder(treeWidget) != "sent":
                 continue
-            if all(
-                    [
-                        treeWidget == self.ui.treeWidgetYourIdentities,
-                        self.getCurrentAccount(treeWidget) not in (fromAddress, None, False),
-                    ]
+            if (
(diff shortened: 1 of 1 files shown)


src/bitmessageqt/__init__.py
```diff
@@ -4644,12 +4614,10 @@ def treeWidgetItemChanged(self, item, column):
...
-        if any(
-                [
-                    not isinstance(item, Ui_AddressWidget),
-                    not self.getCurrentTreeWidget(),
-                    self.getCurrentTreeWidget().currentItem() is None,
-                ]
+        if (
+                not isinstance(item, Ui_AddressWidget) or
+                not self.getCurrentTreeWidget() or
+                self.getCurrentTreeWidget().currentItem() is None
```
PeterSurda: My C background is causing me to code like this.

PeterSurda:
Please rebase, then I'll merge.

src/bitmessageqt/__init__.py
coffeedogs: Pylint likes the boolean operators at the end of the line, too.

PR #1279, opened 2018-06-19
Fixed: Revert bug introduced in playSound

coffeedogs:
During a linting exercise, while attempting to correct an 'inconsistent-return' violation, logic was changed, causing a bug. This reverts the bug.


PeterSurda:
Empty after rebasing.

PR #1280, opened 2018-06-19
Returning bitmessageqt package back to normal

g1itch:
I  reverted all changes to __init__ in this commit.

### DIFF
--- a/src/bitmessageqt/__init__.py
+++ b/src/bitmessageqt/__init__.py
@@ -1,9 +1,5 @@
-# pylint: disable=too-many-lines,broad-except,too-many-instance-attributes,global-statement,too-few-public-methods
-# pylint: disable=too-many-statements,too-many-branches,attribute-defined-outside-init,too-many-arguments,no-member
-# pylint: disable=unused-argument,no-self-use,too-many-locals,unused-variable,too-many-nested-blocks
-# pylint: disable=too-many-return-statements,protected-access,super-init-not-called,non-parent-init-called
 """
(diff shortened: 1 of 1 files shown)


src/bitmessageqt/__init__.py
```diff
@@ -1,78 +1,60 @@
...
-Initialise the QT interface
+PyQt based UI for bitmessage, the main module
 """
 
-import hashlib
+import sys
+from PyQt4 import QtCore, QtGui
+from PyQt4.QtNetwork import QLocalSocket, QLocalServer
+
+from debug import logger
```
omkar1117: write imports first and then from statements later.
Please maintain all the imports in a alphabetic order

g1itch: There is still a problems with import order in this package. Changed import statements is one of the reasons why the previous commit broke everything.

omkar1117: ok. but may I know how come these order break the workflow, so that I can ignore this for future purpose.

g1itch: It seems still nobody knows it exactly. Changing the order sometimes can lead to circular import issue, especially in modules like `bitmessagemain`. That's why for example `depends` imports standard python `logging` instead of using `debug`.

g1itch:
@omkar1117, nevertheless I sorted the imports from standard library. Now I should test it locally.

g1itch:
Squashed to one commit.

PR #1283, opened 2018-06-21
Avoid importing big libs before decision to use them

g1itch:
Came in bitmessage chan:
[Patch](https://beamstat.com/chan/bitmessage/30b1357084631806d8abf97e35452a80b65cda8b317b3446f8ee39b484cc94dc): try to load `pyopencl` before `numpy`
[Patch2](https://beamstat.com/chan/bitmessage/83d50cdaafd9919431a59093611d71281644ae93a6252b9c0e0428600ac15f15): do not load modules which use `email` if not configured

It makes sense  for me. I looked into top -Hp `cat .config/PyBitmessage/singleton.lock` and got these values for VIRT and RES just after pybitmessage started:

```
no changes:   1572.2m 117.2m
no email:     1570.5m 115.3m
and no numpy: 1529.9m 106.6m
```

### DIFF
--- a/src/openclpow.py
+++ b/src/openclpow.py
@@ -20,10 +20,11 @@ vendors = []
 hash_dt = None
 
 try:
-    import numpy
     import pyopencl as cl
-except:
+    import numpy
+except ImportError:
     libAvailable = False
+
 
 def initCL():
     global ctx, queue, program, hash_dt, libAvailable
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -51,8 +51,6 @@ from class_singleCleaner import singleCleaner
 from class_objectProcessor import objectProcessor
 from class_singleWorker import singleWorker
 from class_addressGenerator import addressGenerator
-from class_smtpDeliver import smtpDeliver
-from class_smtpServer import smtpServer
 from bmconfigparser import BMConfigParser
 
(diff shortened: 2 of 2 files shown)


coffeedogs:
Personally, I would think that `class_smtpDeliver` and `class_smtpServer` should be imported at the top of `src/bitmessagemain.py`. Then, within each module they decide how to implement the features based on the modules that are installed. This better encapsulates the concerns and makes for more reusable code.

src/openclpow.py
```diff
@@ -20,11 +20,12 @@
 hash_dt = None
 
 try:
-    import numpy
     import pyopencl as cl
-except:
+    import numpy
+except ImportError:
     libAvailable = False
```
omkar1117: better log it or display it with help command.
print ("Please import numpy module : pip install numpy")

g1itch: @omkar1117 maybe something like "Please install numpy package if you want to use opencl PoW:\n pip install numpy"? It's useful only with AMD GPU, not every user needs opencl.

g1itch: Though it's already done in `bitmessagemain` by `depends.check_dependencies()` call.

PeterSurda:
Please rebase, I'm ready to merge.

Issue #1227, opened 2018-04-28
POW improvements


Kleshni:
I'm going to work on some improvements to the current POW subsystem:

- Continuously update expiry time while doing the work to hide your computing power from peers. Short living messages would also get advatage from this, because POW calculation can take notable part of their TTL.

- Calculate several POWs in parallel in the same thread. Previous change will require to periodically interrupt calculation, so we can interleave several POWs creating something like a simple task scheduler.

- Add ability to stop POW and cancel or edit a message in the UI.

- Add an expected time indicator to the UI, analogous to Vanitygen's.

- Make starting nonce random, solving this issue #1213.

What more?

PeterSurda:
Good idea

PR #1284, opened 2018-06-23
New POW calculation module

Kleshni:
For #1227. It's currently not connected to the PyBitmessage code, I'm going to get it done sooner or later.

It provides the `WorkProver` thread class which manages tasks and schedules them to solvers. There are 4 possible solvers:

- Single-threaded `DumbSolver` in Python. I optimised it so it now works ~2 times faster. In particular, I replaced `hashlib.sha512` with `SHA512` from OpenSSL. Since you need OpenSSL to assemble a message anyway, I think it's safe to use it in a fallback solver.

- Multiprocess based `ForkingSolver` in Python. It works faster too, because it relies on `DumbSolver`.

- Multithreaded `FastSolver` in C. I tried to utilize SIMD, but it gave very little speed up of 7 %, so it still calls OpenSSL.

- And `GPUSolver` in OpenCL.

The library was tested on a 4-core AMD64 Linux with integrated GPU and 45-CPU virtual machines with hyperthreading enabled:
- X86 Linux;
- AMD64 FreeBSD;
- AMD64 OpenBSD with the `bsd.mp` kernel;
- AMD64 NetBSD;
- AMD64 Windows 7.

I failed to install Hackintosh on QEMU, so somebody else has to test it there.

### DIFF
--- a/src/workprover/Readme.md
+++ b/src/workprover/Readme.md
@@ -1,0 +1,42 @@
+Please keep this module independent from the outside code, so that it can be reused in other applications.
+
+If you are going to use it, you should wrap your program's main file in this:
+
+```python
+import workprover.dumbsolver
+
+workprover.dumbsolver.libcrypto = ...
+
+if __name__ == "__main__":
+    import multiprocessing
+
+    multiprocessing.freeze_support()
+
+    ...
+```
+
+See the `multiprocessing` module documentation for explaination.
+
+Build fast solver
+-----------------
+
+On Linux, BSDs or MacOS: `make -C fastsolver`.
+
+On Windows:
+
+- Install OpenSSL. Build it yourself or install [third-party](https://wiki.openssl.org/index.php/Binaries) prebuilt binaries.
+
+- Install MSVC as part of Visual Studio or standalone. Official offline installer: https://aka.ms/vcpython27.
+
+- Open its command line and go to the `fastsolver` directory.
+
+- Add OpenSSL paths to environment variables:
+
+```bat
+set INCLUDE=C:\OpenSSL-Win64\include;%INCLUDE%
+set LIB=C:\OpenSSL-Win64\lib;%LIB%
+```
+
+- Do `cl @options.txt`.
+
+- Append the `-32` or `-64` suffix to the DLL file name.
--- a/src/workprover/__init__.py
+++ b/src/workprover/__init__.py
@@ -1,0 +1,238 @@
+import sys
+import multiprocessing
+import time
+import struct
+import threading
+import Queue
+import collections
+
+import utils
+import dumbsolver
+import forkingsolver
+import fastsolver
+import gpusolver
+
+timeout = .5
+
+class Stop(Exception):
+    pass
+
+class Task(object):
+    previous = None
+    next = None
+
+    def __init__(self, headlessPayload, TTL, expiryTime, target):
+        self.headlessPayload = headlessPayload
+        self.TTL = TTL
+        self.expiryTime = expiryTime
+        self.target = target
+
+class WorkProver(threading.Thread):
+    def __init__(self, codePath, GPUVendors, seed, statusUpdated):
+        super(self.__class__, self).__init__()
+
+        self.availableSolvers = {
+            "dumb": dumbsolver.DumbSolver(codePath)
+        }
+
+        # Comment from the previous version:
+
+        # on my (Peter Surda) Windows 10, Windows Defender
+        # does not like this and fights with PyBitmessage
+        # over CPU, resulting in very slow PoW
+        # added on 2015-11-29: multiprocesing.freeze_support() doesn't help
+
+        if not hasattr(sys, "frozen") or sys.frozen == "macosx_app":
+            self.availableSolvers["forking"] = forkingsolver.ForkingSolver(codePath)
+
+        try:
+            self.availableSolvers["fast"] = fastsolver.FastSolver(codePath)
+        except:
+            pass
+
+        try:
+            self.availableSolvers["gpu"] = gpusolver.GPUSolver(codePath, GPUVendors)
+        except:
+            pass
+
+        try:
+            self.defaultParallelism = multiprocessing.cpu_count()
+        except NotImplementedError:
+            self.defaultParallelism = 1
+
+        self.seed = seed
+        self.roundsCounter = 0
+        self.statusUpdated = statusUpdated
+
+        self.commandsQueue = Queue.Queue()
+        self.resultsQueue = Queue.Queue()
+
+        self.solverName = None
+        self.solver = None
+
+        self.lastTime = utils.getTimePoint()
+        self.timedIntervals = collections.deque()
+        self.speed = 0
+
+        self.tasks = {}
+        self.currentTaskID = None
+
+    def notifyStatus(self):
+        if self.statusUpdated is None:
+            return
+
+        if self.solver is None:
+            parallelism = 0
+        else:
+            parallelism = self.solver.parallelism
+
+        self.statusUpdated((self.solverName, parallelism, self.speed))
+
+    def setSolver(self, name, parallelism):
+        if name is None and self.solverName is None:
+            pass
+        elif name == self.solverName:
+            if self.solver.parallelism != parallelism:
+                self.solver.setParallelism(parallelism)
+        else:
+            if self.solver is not None:
+                self.solver.setParallelism(0)
+                self.solverName = None
+                self.solver = None
+
+            if name is not None:
+                if name not in self.availableSolvers:
+                    name, parallelism = "dumb", 1
+
+                self.solverName = name
+                self.solver = self.availableSolvers[name]
+                self.solver.setParallelism(parallelism)
+
+        self.notifyStatus()
+
+    def updateSpeed(self, iterationsCount):
+        currentTime = utils.getTimePoint()
+        duration = currentTime - self.lastTime
+        self.lastTime = currentTime
+
+        self.timedIntervals.append((currentTime, iterationsCount, duration))
+
+        for i in xrange(len(self.timedIntervals)):
+            time, iterationsCount, duration = self.timedIntervals[0]
+
+            if time + duration < currentTime - 3:
+                self.timedIntervals.popleft()
+
+        totalDuration = 0
+        totalIterationsCount = 0
+
+        for time, iterationsCount, duration in self.timedIntervals:
+            totalIterationsCount += iterationsCount
+            totalDuration += duration
+
+        if totalDuration < .25:
+            self.speed = 0
+        else:
+            self.speed = totalIterationsCount / totalDuration
+
+        self.notifyStatus()
+
+    def addTask(self, ID, headlessPayload, TTL, expiryTime, byteDifficulty, lengthExtension):
+        target = utils.calculateTarget(8 + 8 + len(headlessPayload), TTL, byteDifficulty, lengthExtension)
+
+        task = Task(headlessPayload, TTL, expiryTime, target)
+
+        self.tasks[ID] = task
+
+        if self.currentTaskID is None:
+            task.previous = ID
+            task.next = ID
+
+            self.currentTaskID = ID
+        else:
+            task.previous = self.currentTaskID
+            task.next = self.tasks[self.currentTaskID].next
+
+            self.tasks[task.previous].next = ID
+            self.tasks[task.next].previous = ID
+
+    def cancelTask(self, ID):
+        if ID not in self.tasks:
+            return
+
+        task = self.tasks.pop(ID)
+
+        if len(self.tasks) == 0:
+            self.currentTaskID = None
+        else:
+            self.tasks[task.previous].next = task.next
+            self.tasks[task.next].previous = task.previous
+
+            if self.currentTaskID == ID:
+                self.currentTaskID = task.next
+
+    def nextTask(self):
+        self.currentTaskID = self.tasks[self.currentTaskID].next
+
+    def shutdown(self):
+        self.setSolver(None, 0)
+
+        for i in self.tasks.keys():
+            self.cancelTask(i)
+
+        raise Stop()
+
+    def processCommand(self, command, *arguments):
+        getattr(self, command)(*arguments)
+
+    def round(self):
+        while True:
+            try:
+                self.processCommand(*self.commandsQueue.get_nowait())
+            except Queue.Empty:
+                break
+
+        while self.solver is None or self.currentTaskID is None:
+            try:
+                self.processCommand(*self.commandsQueue.get(True, timeout))
+            except Queue.Empty:
+                self.updateSpeed(0)
+
+        task = self.tasks[self.currentTaskID]
+
+        if task.expiryTime is None:
+            expiryTime = int(time.time() + task.TTL)
+        else:
+            expiryTime = task.expiryTime
+
+        initialPayload = struct.pack(">Q", expiryTime) + task.headlessPayload
+        initialHash = utils.calculateInitialHash(initialPayload)
+
+        appendedSeed = self.seed + struct.pack(">Q", self.roundsCounter)
+        self.roundsCounter += 1
+
+        try:
+            nonce, iterationsCount = self.solver.search(initialHash, task.target, appendedSeed, timeout)
+        except gpusolver.GPUSolverError:
+            self.setSolver("dumb", 1)
+            self.availableSolvers.pop("gpu")
+
+            nonce, iterationsCount = None, 0
+
+        self.updateSpeed(iterationsCount)
+
+        if nonce is None:
+            self.nextTask()
+        else:
+            self.resultsQueue.put(("taskDone", self.currentTaskID, nonce, expiryTime))
+
+            self.cancelTask(self.currentTaskID)
+
+    def run(self):
+        try:
+            while True:
+                self.round()
+        except Stop:
+            return
+        except Exception as exception:
+            self.resultsQueue.put(exception)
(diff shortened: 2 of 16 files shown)


PeterSurda:
Have you tried frozen mode on Windows?

Kleshni:
No, but `ForkingSolver` is disabled in this case complying to the current behaviour. All file paths (to the C library and OpenCL kernel) are relative to the code directory, whose path must be provided by the calling code.

src/workprover/forkingsolver.py
```diff
@@ -0,0 +1,106 @@
+import os
```
omkar1117: pep8 validation for the file

src/workprover/forkingsolver.py
```diff
@@ -0,0 +1,106 @@
...
+import multiprocessing
+import struct
+
+import dumbsolver
+
+def setIdle():
+    try:
+        import psutil
+
```
omkar1117: remove un necessary spaces

src/workprover/Readme.md
```diff
@@ -0,0 +1,42 @@
+Please keep this module independent from the outside code, so that it can be reused in other applications.
```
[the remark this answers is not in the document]
Kleshni: It's a Python package with the `__init__.py` file, and it's intended to be imported like `import workprover`. It could be moved to a separate repository, but I think, it's easier and safer to keep it in the same file tree.

src/workprover/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+        self.speed = 0
+
+        self.tasks = {}
+        self.currentTaskID = None
+
+    def notifyStatus(self):
+        if self.statusUpdated is None:
+            return
+
+        if self.solver is None:
```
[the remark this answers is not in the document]
Kleshni: It's less readable.

src/workprover/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+
+        self.lastTime = utils.getTimePoint()
+        self.timedIntervals = collections.deque()
+        self.speed = 0
+
+        self.tasks = {}
+        self.currentTaskID = None
+
+    def notifyStatus(self):
+        if self.statusUpdated is None:
```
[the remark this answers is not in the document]
Kleshni: It's less acurate.

`self.statusUpdated` must be either a function or a `None`. If it's 0 or an empty string, it would be a programming error and further call would rise an exception making the error noticable.

src/workprover/__init__.py
```diff
@@ -0,0 +1,238 @@
...
+
+        if self.solver is None:
+            parallelism = 0
+        else:
+            parallelism = self.solver.parallelism
+
+        self.statusUpdated((self.solverName, parallelism, self.speed))
+
+    def setSolver(self, name, parallelism):
+        if name is None and self.solverName is None:
```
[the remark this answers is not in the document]
Kleshni: It calls the `self.statusUpdated` callback and the outside code should log this and display in the GUI.

g1itch:
Some tests [are failing](https://travis-ci.org/g1itch/PyBitmessage/builds/395814851). It seems to be expected for `TestGPUSolver`, what about the rest?

g1itch:
[Strange thing](https://travis-ci.org/g1itch/PyBitmessage/builds/395846736):

> OSError: /home/travis/build/g1itch/PyBitmessage/src/workprover/fastsolver/libfastsolver.so: undefined symbol: SHA512_Update

It seems to be specific to ubuntu:trusty or Travis-CI.

Nevertheless could you please add `pybitmessage.workprover` to `packages` and decorate `TestGPUSolver` like I did?

Kleshni:
I have registered in Travis-CI and debugged the issue. It was a very stupid mistake, shared libraries like `-lpthread` and `-lcrypto` should be listed after object files in the linker command line.

Travis-CI also gives an OS X environment, so I tested the module there.

PeterSurda:
`packages/pyinstaller/bitmessagemain.spec` needs updating.

src/workprover/gpusolver.py
```diff
@@ -0,0 +1,106 @@
...
+class GPUSolverError(Exception):
+    pass
+
+class GPUSolver(object):
+    def __init__(self, codePath, vendors = None):
+        global pyopencl, numpy
+
+        try:
+            import numpy
+            import pyopencl
```
g1itch: Could you please comment #1283 and maybe change the import order too if you agree?

g1itch:
Could you please also add the extension and a workaround for dh_python in Debian: a9955e5?

Kleshni:
Maybe just invoke the makefile from `setup.py`?

g1itch:
The extension approach is more pythonic and there is no need for makefile. You can see it even in the [tests log](https://travis-ci.org/g1itch/PyBitmessage/builds/397318626).

Kleshni:
But there rises up a need for a workaround, while the makefile is needed anyway. And a pythonic extention is not a `ctypes` library, it must use Python C API to be directly importable:
>A C extension for CPython is a shared library (e.g. a .so file on Linux, .pyd on Windows), which exports an *initialization function*.

src/api.py
```diff
@@ -10,8 +10,11 @@
 import base64
 import hashlib
 import json
+import socket
+import threading
 import time
 from binascii import hexlify, unhexlify
```
omkar1117: Please place all the imports in a order

src/api.py
```diff
@@ -1297,3 +1301,49 @@ def _dispatch(self, method, params):
...
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        try:
+            s.connect((
+                BMConfigParser().get('bitmessagesettings', 'apiinterface'),
+                BMConfigParser().getint('bitmessagesettings', 'apiport')
+            ))
+            s.shutdown(socket.SHUT_RDWR)
+            s.close()
+        except:
+            pass
```
omkar1117: better to place log instead of exception pass.
I hope it will be dangerous if we have pass statement.

src/bitmessagemain.py
```diff
@@ -499,10 +381,87 @@ def main():
     mainprogram = Main()
     mainprogram.start()
 
+# See "workprover/Readme.md"
+
+import pyelliptic.openssl
```
omkar1117: Place this imports in the starting of the script.

g1itch:
Are you going to connect new POW code soon? This PR still not includes the actual use of this code for doing work.

Kleshni:
I almost finished rewriting `class_singleWorker`, but I have a problem with the function for requesting pubkeys. It uses the `retrynumber` column in the database for detemining the TTL and rising it exponentially for every next attempt.

But proper implementation of this scheme gets very complicated if two or more messages need the same pubkey and have different retry numbers. It would be wise to not send two simultaneous getpubkey requests in this case, and to not send it at all if it's already present in the inventory. So there emerges a question on how to handle the `retrynumber` column in such situations, and what to do with it when some other user sends a getpubkey request for a key we also need.

I think we can get rid of this column, I mean use it only for resending the message itself but ignore it for getpubkeys. This would simplify the algorithm: just check there is no requests in the inventory, and send it with a fixed TTL.

The comment near the [line](https://github.com/Bitmessage/PyBitmessage/blob/529559d06a3cd1c89838482acaf0fd16a018a057/src/class_singleWorker.py#L1324) calculating the TTL says, that the 2,5-day period was chosen "fairly arbitrarily". Maybe redefine it to the maximum 28 days? Getpubkey requests are negligibly small and are sent with the default network difficulty, they don't need too much work.

There are currently 13661 getpubkeys in the inventory, but only 84 of them have different tags! The others are duplicates, which I propose to avoid by simplifying the sending algorithm. And there are 40939 different pubkeys, so I conclude that the need for getpubkeys is very small, and there is no big reason to save on short TTLs.

Less frequent resending and not repeating already existing requests can also affect privacy in the positive way.

(Another question is that the current requesting function specifies the object type as 0 in one place and 1 in another, but I think it's certainly a bug.)

omkar1117:
@Kleshni can I have the functional flow also along with the PR, it will be easy to test the internal things please.

Kleshni:
Sorry, I don't know what the "functional flow" is. I tried to google, but it seems, that I have to learn much new material to satisfy your request.

>with some snapshots
Snapshots of what?

Maybe I should add doc strings to functions or something like that?

Kleshni:
I don't completely understand why the TTL randomization is needed:

```python
TTL += random.randrange(-300, 300)
expiryTime = int(time.time() + TTL)
```

Possibly, it's supposed to hide the origin node of the message because it makes neighbour nodes unable to tell whether the message was generated right now or its retranslation took some time.

But it seems very weak:

1. Maybe it could hide singular messages, but if a node generates messages continuously (like the time service broadcast), the expected value of the random addition approaches -0,5 - a constant.

2. An attacker can guess the original TTL, which is usually a multiple of 3600 seconds. If the expiry time of a received object is `currentTime + originalTTL + 299`, then it's obvious, that the random addition was 299 seconds, it can't be more. In this case it hides nothing.

3. TTL affects POW difficulty, POW difficulty affects the nonce value, the nonce value is known to everybody. It's another way to leak the real sending time.

I don't know how to solve the first two problems, but the the last can be solved by randomizing the expiry time instead:

```
TTL += 599
expiryTime = int(time.time() + TTL - random.randrange(600))
```

So I propose to change the code this way.

Kleshni:
I deleted the file `bitmessageqt/settings.py`. It was originally generated by the `pyuic4` utility from `bitmessageqt/settings.ui` and was not actually needed, because `*.ui` files can be used directly without conversion to `*.py`.

The generated files must remain untouched because they need a regeneration on every change to the source files, but `settings.py` was manually modified one day. This prevented easy modification of the settings window, so I solved this by moving the changes out from this file and deleting it.

Kleshni:
The same for the main window.

src/bitmessageqt/__init__.py
```diff
@@ -584,9 +582,51 @@ def rerenderTabTree(self, tab):
...
+        self.ui.treeWidgetSubscriptions.resize(200, self.ui.treeWidgetSubscriptions.height())
+        self.ui.treeWidgetYourIdentities.resize(200, self.ui.treeWidgetYourIdentities.height())
+
+        # TODO: why we need splitters here if they are disabled?
+
+        self.ui.horizontalSplitter_2.handle(1).setEnabled(False)
+        self.ui.horizontalSplitter_6.handle(1).setEnabled(False)
+        self.ui.horizontalSplitterSearch.handle(1).setEnabled(False)
+        self.ui.verticalSplitter.handle(1).setEnabled(False)
+        self.ui.verticalSplitter_2.handle(1).setEnabled(False)
```
omkar1117: can't we make it dynamic or iterable?

src/api.py
```diff
@@ -1297,3 +1301,49 @@ def _dispatch(self, method, params):
             logger.exception(e)
 
             return "API Error 0021: Unexpected API Failure - %s" % e
+
+# This thread, of which there is only one, runs the API.
+class singleAPI(threading.Thread, helper_threading.StoppableThread):
+    def __init__(self):
```
omkar1117: please add doc string to the class.

src/api.py
```diff
@@ -1297,3 +1301,49 @@ def _dispatch(self, method, params):
...
+        threading.Thread.__init__(self, name="singleAPI")
+        self.initStop()
+
+    def stopThread(self):
+        super(singleAPI, self).stopThread()
```
omkar1117: please add doc string to the class

src/bitmessagemain.py
```diff
@@ -502,10 +384,85 @@ def main():
     mainprogram = Main()
     mainprogram.start()
 
+# See "workprover/Readme.md"
+
+import os
```
omkar1117: please write all the imports on the starting of the file.

src/bitmessagemain.py
```diff
@@ -502,10 +384,85 @@ def main():
...
+sys.path.insert(0, app_dir)
+
+import pyelliptic.openssl
+import workprover.dumbsolver
+
+workprover.dumbsolver.libcrypto = pyelliptic.openssl.OpenSSL._lib
 
 if __name__ == "__main__":
-    main()
+    import multiprocessing
```
omkar1117: Please place all the imports in the starting the file...

src/class_singleCleaner.py
```diff
@@ -35,6 +35,45 @@
 import queues
 import state
 
+def resendStaleMessages():
+    staleMessages = sqlQuery("""
+        SELECT "toaddress", "ackdata", "status" FROM "sent"
```
omkar1117: Pep8 validation here.

src/class_sqlThread.py
```diff
@@ -148,16 +150,55 @@ def run(self):
...
+            UPDATE "sent" SET "status" = 'msgqueued'
+            WHERE "status" IN (
+                'doingmsgpow', 'doingpubkeypow', 'awaitingpubkey',
+                'doingpow', 'findingpubkey' -- Old names
+            );
+        """)
+
+        # Update status values inherited from old versions
+
+        self.cur.execute("""UPDATE "sent" SET "status" = 'msgsent' WHERE "status" == 'sentmessage';""")
```
omkar1117: pep8 validation here.

src/workprover/__init__.py
```diff
@@ -0,0 +1,253 @@
+import Queue
```
omkar1117: please place all the imports in a alphabetic order.

src/workprover/__init__.py
```diff
@@ -0,0 +1,253 @@
...
+
+        status = None
+
+        if self.solver is not None:
+            status = self.solver.status
+
+        self.statusUpdated(Status(self.solverName, status, self.speed, len(self.tasks), self.totalDifficulty))
+
+    def setSolver(self, name, configuration):
+        if name is None and self.solverName is None:
```
omkar1117: if not name and not self.solverName:
    pass

src/workprover/__init__.py
```diff
@@ -0,0 +1,253 @@
...
+            pass
+        elif name == self.solverName:
+            self.solver.setConfiguration(configuration)
+        else:
+            if self.solver is not None:
```
omkar1117: if self.solver:
   ….

src/workprover/__init__.py
```diff
@@ -0,0 +1,253 @@
...
+                self.solver.setConfiguration(None)
+                self.solverName = None
+                self.solver = None
+
+            if name is not None:
```
omkar1117: if name:
 ….

src/bitmessagemain.py
Kleshni: These imports must be conditional.

Kleshni:
I deleted the `getMessageDataByDestination{Hash,Tag}` methods because they were undocumented and didn't work at all, so I concluded that no-one ever used them. They are now replaced by `getRawObject` and `listRawObjects`.

The same for `disseminatePreEncryptedMsg` and `disseminatePubkey`.

src/workprover/__init__.py
```diff
@@ -0,0 +1,262 @@
...
+                    self.solver = None
+
+                if name is not None:
+                    if name not in self.availableSolvers:
+                        name, configuration = "dumb", None
+
+                    self.solverName = name
+                    self.solver = self.availableSolvers[name]
+                    self.solver.setConfiguration(configuration)
+        except GPUSolverError:
```
g1itch: undefined name 'GPUSolverError'

g1itch:
I got 3 additional threads (`workprover.WorkProver` instances) which don't stop when I run `src/pybitmessage.py -t`:
https://travis-ci.org/g1itch/PyBitmessage/builds/411231409

This patch helped:
```patch
diff --git a/src/singleworker.py b/src/singleworker.py
index 9fa1391c..e1eeda3f 100644
--- a/src/singleworker.py
+++ b/src/singleworker.py
@@ -270,17 +270,19 @@ def setBestSolver():
 
 setBestSolver()
 
+
 class singleWorker(threading.Thread, helper_threading.StoppableThread):
     name = "singleWorker"
 
     def __init__(self):
-        super(self.__class__, self).__init__()
+        super(self.__class__, self).__init__(name="singleWorker")
 
         self.initStop()
 
     def stopThread(self):
         queues.workerQueue.put(("stopThread", "data"))
 
+        workProver.commandsQueue.put(("shutdown", ))
         super(self.__class__, self).stopThread()
 
     def run(self):
```

Kleshni:
Any plans on merging?

PeterSurda:
@Kleshni I haven't looked at it yet, but it first needs to pass all the code quality checks and all commits need to be signed (you can squash commits if it helps).

However, others gave good feedback so chances are good.

PeterSurda:
Also, please split this into multiple patches. The PoW, the API and the other changes should be separate. You'll have better chances of getting it merged that way too.

PeterSurda:
@Kleshni would it be ok if I assigned someone else to the task to help you with cleaning it up? You'd have to give them write access to your repo.

Kleshni:
OK.

PeterSurda:
@Kleshni please give @omkar1117 write access to the POW branch. this way your commits will be preserved and you can still get paid through tip4commit.

Kleshni:
I have sent him an invite yesterday but forgot to notify you :eyes:

omkar1117:
@Kleshni could you please resolve the conflicsts please.

PeterSurda:
@Kleshni Omkar says he doesn't have write access. He could fork it into his own repo, but since this PR is already open I would prefer to continue this way. Can you check and let me know?

Kleshni:
This is what I see in the repository settings:
Screenshot

PR #1295, opened 2018-06-30
Fix an issue from #1153 for setup without plugins package

g1itch:
A fix for an issue came from chan.


omkar1117:
It's ok for me in the point of code review

PR #1296, opened 2018-07-01
Moved all config checks to helper_startup

g1itch:
Hello.

I rewrote all config checks with `.safeGet..` and moved to `helper_startup` module, although not touched `bitmessagemain` in this changeset.

### DIFF
--- a/src/helper_startup.py
+++ b/src/helper_startup.py
@@ -30,19 +30,20 @@ def _loadTrustedPeer():
 
 
 def loadConfig():
+    config = BMConfigParser()
     if state.appdata:
-        BMConfigParser().read(state.appdata + 'keys.dat')
+        config.read(state.appdata + 'keys.dat')
         # state.appdata must have been specified as a startup option.
-        needToCreateKeysFile = BMConfigParser().safeGet(
+        needToCreateKeysFile = config.safeGet(
             'bitmessagesettings', 'settingsversion') is None
         if not needToCreateKeysFile:
             print(
                 'Loading config files from directory specified'
                 ' on startup: %s' % state.appdata)
     else:
-        BMConfigParser().read(paths.lookupExeFolder() + 'keys.dat')
+        config.read(paths.lookupExeFolder() + 'keys.dat')
         try:
-            BMConfigParser().get('bitmessagesettings', 'settingsversion')
+            config.get('bitmessagesettings', 'settingsversion')
             print 'Loading config files from same directory as program.'
             needToCreateKeysFile = False
             state.appdata = paths.lookupExeFolder()
@@ -50,8 +51,8 @@ def loadConfig():
             # Could not load the keys.dat file in the program directory.
             # Perhaps it is in the appdata directory.
             state.appdata = paths.lookupAppdataFolder()
-            BMConfigParser().read(state.appdata + 'keys.dat')
-            needToCreateKeysFile = BMConfigParser().safeGet(
+            config.read(state.appdata + 'keys.dat')
+            needToCreateKeysFile = config.safeGet(
                 'bitmessagesettings', 'settingsversion') is None
             if not needToCreateKeysFile:
                 print 'Loading existing config files from', state.appdata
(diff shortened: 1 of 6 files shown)


src/helper_startup.py
```diff
@@ -30,103 +30,98 @@ def _loadTrustedPeer():
...
         # no config file (or it cannot be accessed). Create config file.
-        BMConfigParser().add_section('bitmessagesettings')
-        BMConfigParser().set('bitmessagesettings', 'settingsversion', '10')
-        BMConfigParser().set('bitmessagesettings', 'port', '8444')
-        BMConfigParser().set(
-            'bitmessagesettings', 'timeformat', '%%c')
-        BMConfigParser().set('bitmessagesettings', 'blackwhitelist', 'black')
-        BMConfigParser().set('bitmessagesettings', 'startonlogon', 'false')
+        config.add_section('bitmessagesettings')
+        config.set('bitmessagesettings', 'settingsversion', '10')
```
omkar1117: Why can't we have all this constants some where and loop them ?

src/helper_startup.py
```diff
@@ -30,103 +30,98 @@ def _loadTrustedPeer():
...
-        #start:UI setting to stop trying to send messages after X days/months
-        BMConfigParser().set(
-            'bitmessagesettings', 'stopresendingafterxdays', '')
-        BMConfigParser().set(
-            'bitmessagesettings', 'stopresendingafterxmonths', '')
-        #BMConfigParser().set(
-        #    'bitmessagesettings', 'timeperiod', '-1')
-        #end
+            config.set('bitmessagesettings', 'minimizetotray', 'true')
+        config.set('bitmessagesettings', 'showtraynotifications', 'true')
```
omkar1117: same here, can we have this constants in a constant file and call them here?

src/helper_startup.py
```diff
@@ -139,16 +134,187 @@ def loadConfig():
...
+        updateConfig()
 
     _loadTrustedPeer()
 
 
+def updateConfig():
+    config = BMConfigParser()
+    settingsversion = config.getint('bitmessagesettings', 'settingsversion')
+    if settingsversion == 1:
+        config.set('bitmessagesettings', 'socksproxytype', 'none')
```
omkar1117: same here also

src/helper_startup.py
```diff
@@ -139,16 +134,187 @@ def loadConfig():
...
+        settingsversion = 5
+
+    if settingsversion == 5:
+        config.set(
+            'bitmessagesettings', 'maxacceptablenoncetrialsperbyte', '0')
+        config.set(
+            'bitmessagesettings', 'maxacceptablepayloadlengthextrabytes', '0')
+        settingsversion = 7
+
+    # Raise the default required difficulty from 1 to 2
```
omkar1117: Remove this commented code if not required.

src/helper_startup.py
```diff
@@ -139,16 +134,187 @@ def loadConfig():
...
+    # a certain period of time even if we never get an ack
+    if settingsversion == 7:
+        config.set('bitmessagesettings', 'stopresendingafterxdays', '')
+        config.set('bitmessagesettings', 'stopresendingafterxmonths', '')
+        settingsversion = 8
+
+    # With the change to protocol version 3, reset the user-settable
+    # difficulties to 1
+    if settingsversion == 8:
+        config.set(
```
omkar1117: Better to maintain the code in functions instead of this length

src/helper_startup.py
g1itch: It's a compatibility layer for previous versions. To turn it into "constants" I would be forced to define a dictionary of dictionaries. It would look awful I think.

src/helper_startup.py
g1itch: I retained commented out code in order to facilitate a review, so you can see how I interpret logic of the code. I can remove any commented code by final commit if all reviewers agree that changes are correct and the code is indeed not required.

omkar1117: ok, I think If we write a function with in 15 - 20 lines its better to have a good review.

coffeedogs: I'd be more in favour of using the PR description, self-comments on the PR or a gist, referred to in the PR description rather than removing comments about the change in a final, post-review commit. Comments should be about the code in a particular revision, not the change/diff.

So, rather than saying `# increase frobulation from 1 to 3` there should be a `# frobulation should be 1 due to the small size of the userbase` comment in the earlier revision and `# frobulation should be 3 due to the large size of the userbase` in the later revision. Viewing the diff should provide the transitional view that the frobulation has changed from 1 to 3.

But I don't have particularly strong feelings about this.

Issue #1298, opened 2018-07-05
After user chooses "Quit" and allows Bitmesasge to wait till sync is done, app can only be stopped via "kill"


stayen:
When I press "Quit", PyBitmessage offers to wait till all the sync is done, then quit.

If I need to quit it anyway and click "Quit" again, nothing happens. The only means to kill the app at that moment is by using "kill" command (in my case, Ubuntu 16.04.4, only "kill -9" worked).

g1itch:
What if I often have a couple of dozen objects to be synced? This expression will wait forever: [`__init__.py`#L2764](../blob/v0.6/src/bitmessageqt/__init__.py#L2764).

g1itch:
We can use something like
```patch
diff --git a/src/bitmessageqt/__init__.py b/src/bitmessageqt/__init__.py
index 5e4c18fd..6c7b2e7e 100644
--- a/src/bitmessageqt/__init__.py
+++ b/src/bitmessageqt/__init__.py
@@ -627,8 +627,8 @@ class MyForm(settingsmixin.SMainWindow):
         self.replyFromTab = None
 
         # so that quit won't loop
-        self.quitAccepted = False
-        
+        self.quitAccepted = self.wait = False
+
         self.init_file_menu()
         self.init_inbox_popup_menu()
         self.init_identities_popup_menu()
@@ -2703,7 +2703,7 @@ class MyForm(settingsmixin.SMainWindow):
             return
         '''
 
-        if self.quitAccepted:
+        if self.quitAccepted and not self.wait:
             return
 
         self.show()
@@ -2731,7 +2731,7 @@ class MyForm(settingsmixin.SMainWindow):
                     _translate("MainWindow", "Bitmessage hasn't synchronised with the network, %n object(s) to be downloaded. If you quit now, it may cause delivery delays. Wait until the synchronisation finishes?", None, QtCore.QCoreApplication.CodecForTr, pendingDownload()),
                     QtGui.QMessageBox.Yes|QtGui.QMessageBox.No|QtGui.QMessageBox.Cancel, QtGui.QMessageBox.Cancel)
             if reply == QtGui.QMessageBox.Yes:
-                waitForSync = True
+                self.wait = waitForSync = True
             elif reply == QtGui.QMessageBox.Cancel:
                 return
 
@@ -2741,7 +2741,7 @@ class MyForm(settingsmixin.SMainWindow):
[...]

PR #1300, opened 2018-07-06
 Allow re-call of MyForm.quit() if user selected to wait at first

g1itch:
Hello.

This is related to #1298. The "Wait until ... ?" message box will be shown again if user activate "File -> Quit" or "Quit" context menu on tray icon after selecting to wait. However this not works with window close button - `MyForm.closeEvent()` not called.

### DIFF
--- a/src/bitmessageqt/__init__.py
+++ b/src/bitmessageqt/__init__.py
@@ -627,8 +627,8 @@ class MyForm(settingsmixin.SMainWindow):
         self.replyFromTab = None
 
         # so that quit won't loop
-        self.quitAccepted = False
-        
+        self.wait = self.quitAccepted = False
+
         self.init_file_menu()
         self.init_inbox_popup_menu()
         self.init_identities_popup_menu()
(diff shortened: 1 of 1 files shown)


src/bitmessageqt/__init__.py
```diff
@@ -2708,15 +2708,8 @@ def network_switch(self):
...
-        '''quit_msg = "Are you sure you want to exit Bitmessage?"
-        reply = QtGui.QMessageBox.question(self, 'Message',
-                         quit_msg, QtGui.QMessageBox.Yes, QtGui.QMessageBox.No)
-
-        if reply is QtGui.QMessageBox.No:
-            return
-        '''
-
-        if self.quitAccepted:
+        """Quit the bitmessageqt application"""
```
omkar1117: write comment rather than triple quotes

g1itch: Look precisely: I removed triple quotes and added docstring.

PeterSurda:
@g1itch Can you look at my comment? Do you think it's easy to fix?

PeterSurda:
@g1itch ok then I'll approve and update the issue to reflect there is still work to do.

PR #1302, opened 2018-07-06
Moved class singleAPI to api module and `import api` below config check

g1itch:
This allows to further reduce early imports in `bitmessagemain` and potentially to not import `api` at all.

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -8,8 +8,13 @@ This is not what you run to run the Bitmessage API. Instead, enable the API
 """
 
 import base64
+import errno
 import hashlib
 import json
+import random  # nosec
+import socket
+import subprocess
+import threading
 import time
 from binascii import hexlify, unhexlify
 from SimpleXMLRPCServer import SimpleXMLRPCRequestHandler, SimpleXMLRPCServer
@@ -23,6 +28,7 @@ from bmconfigparser import BMConfigParser
 import defaults
 import helper_inbox
 import helper_sent
+import helper_threading
 
 import state
 import queues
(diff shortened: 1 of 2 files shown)


src/api.py
```diff
@@ -60,6 +66,69 @@ def serve_forever(self):
             self.handle_request()
 
 
+# This thread, of which there is only one, runs the API.
+class singleAPI(threading.Thread, helper_threading.StoppableThread):
+    def __init__(self):
```
omkar1117: doc strings please

src/api.py
```diff
@@ -60,6 +66,69 @@ def serve_forever(self):
...
+                BMConfigParser().get('bitmessagesettings', 'apiinterface'),
+                BMConfigParser().getint('bitmessagesettings', 'apiport')
+            ))
+            s.shutdown(socket.SHUT_RDWR)
+            s.close()
+        except:
+            pass
+
+    def run(self):
+        port = BMConfigParser().getint('bitmessagesettings', 'apiport')
```
omkar1117: doc strings here also.

src/api.py
```diff
@@ -23,6 +28,7 @@
 import defaults
```
omkar1117: Please place all the imports in the first order.

src/bitmessagemain.py
```diff
@@ -338,19 +287,9 @@ def start(self):
...
-                except:
-                    apiNotifyPath = ''
-                if apiNotifyPath != '':
-                    with shared.printLock:
-                        print('Trying to call', apiNotifyPath)
-
-                    call([apiNotifyPath, "startingUp"])
-                singleAPIThread = singleAPI()
+            if BMConfigParser().safeGetBoolean('bitmessagesettings', 'apienabled'):
+                import api
```
PeterSurda: Just add a disable comment for pylint, then the check will succeed and you can merge (I already approved).

Issue #1292, opened 2018-06-30
Option to connect/disconnect should require Confirmation


AyrA:
As of now, clicking the Menu option to connect/disconnect will not show a message but immediately perform that action.

I recommend asking the user if he wants that action performed because the menu option is close to others and might be clicked by accident

g1itch:
Excuse me, I don't understand how it might be clicked by accident. It close to "Quit" and "Regenerate deterministic addresses". If you clicked it by accident, just click again.

AyrA:
> If you clicked it by accident, just click again.

And wait for ages for the client to reconnect? No thanks. Every Application that has a TCP listener I know of will show a confirmation dialog before disconnecting. Reconnecting Bitmessage occasionally takes a long time. I personally would also move this menu item to the settings menu

g1itch:
> And wait for ages for the client to reconnect? No thanks.

~25 seconds

So you want to force the people who know what they're doing to do an extra click because you often regenerate your addresses and misclick to "Go offline"?

AyrA:
> ~25 seconds

If you are really lucky. Bitmessage only tries 8 simultaneous addresses and waits for an entire 20 seconds before continuing (by the way, this should probably be reduced too).

20 seconds is still much longer than confirming that you actually want to disconnect. I doubt that people are using this anyway. You can just exit the client if you want to so I am not sure what the point of this option is at all. Neither any browser nor Mail client nor Torrent application I am using has a disconnect option that prominent.

g1itch:
It's probably specific to PyBitmessage. I've often seen complaints about `dontconnect` config setting. Users want easy knob to go offline.

PeterSurda:
I would add the confirmation, this is not really a technical but a usability question.

g1itch:
OK, let's try

g1itch:
@AyrA did #1303 solve this issue?

AyrA:
It did, but instead of OK/Cancel it should probably be Yes/No

posta-store:
many users hat those confirmation with a vengeance. introducing another pointless "YOU REALLY REALLY WANNA DO IT" without posibility of clicking "NEVER AGAIN GET ON MY NERVES" to disable all repitition of it is a complete  ANTI-feature and hated by most users.

AyrA:
You know you can confirm these dialogs by pressing Y/N? do you?

Next question, why are you upvoting your own comment?

Kleshni:
[...]

PR #1303, opened 2018-07-07
Confirmation for network switch

g1itch:
A simple confirmation msgbox for #1292. Perhaps you'll suggest a better text.

### DIFF
--- a/src/bitmessageqt/__init__.py
+++ b/src/bitmessageqt/__init__.py
@@ -2684,6 +2684,19 @@ class MyForm(settingsmixin.SMainWindow):
     def network_switch(self):
         dontconnect_option = not BMConfigParser().safeGetBoolean(
             'bitmessagesettings', 'dontconnect')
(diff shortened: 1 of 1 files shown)


src/bitmessageqt/__init__.py
```diff
@@ -2684,6 +2684,19 @@ def click_NewAddressDialog(self):
     def network_switch(self):
         dontconnect_option = not BMConfigParser().safeGetBoolean(
             'bitmessagesettings', 'dontconnect')
+        reply = QtGui.QMessageBox.question(
+            self, _translate("MainWindow", "Disconnecting")
+            if dontconnect_option else _translate("MainWindow", "Disconnecting"),
```
PeterSurda: This should be "Connecting" rather than "Disconnecting".

g1itch: oops (:

PR #1309, opened 2018-07-17
make ## ./bitmessagemain.py --curses ##  work in Konsole and xterm

KM-200:
subprocess.check_call('which dialog') # no longer works

the "which" is superfluous

### DIFF
--- a/src/depends.py
+++ b/src/depends.py
@@ -363,7 +363,7 @@ def check_curses():
     import subprocess
 
     try:
-        subprocess.check_call(       "dialog")
+        subprocess.check_call("dialog")
         #subprocess.check_call('which dialog') # no longer works
     except subprocess.CalledProcessError:
         logger.error(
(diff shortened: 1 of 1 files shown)


src/depends.py
```diff
@@ -363,7 +363,8 @@ def check_curses():
     import subprocess
 
     try:
-        subprocess.check_call('which dialog')
+        subprocess.check_call(       "dialog")
```
g1itch: Why so many spaces? Also did you try '`which dialog`'? @coffeedogs

KM-200: "which dialog"  was a no workie here, but "dialog" alone worked

g1itch: Oh, markdown, I meant
```
"`which dialog`"
```

g1itch:
@KM-200, it will change nothing. Better wait for author. I'll approve when my review requested.

KM-200:
well OK then. but its helpful against breakage using  --curses  UI

KM-200:
#1309 is a dupe of #1310 
thank you very much, github

PR #1310, opened 2018-07-17
"which" is superfluous and it  breaks     "bitmessagemain --curses"

KM-200:
the word "which" must be removed

### DIFF
--- a/src/depends.py
+++ b/src/depends.py
@@ -363,7 +363,7 @@ def check_curses():
     import subprocess
 
     try:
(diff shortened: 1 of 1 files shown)


KM-200:
all checks passed ! finally we got ....   **--curses** working again !   both in Konsole and xterm !

PeterSurda:
It looks like this has been fixed in a different PR.

KM-200:
OK then

PR #1317, opened 2018-07-19
Fix exception 'KeyError: None' in checkdeps introduced in e92a85e

g1itch:
Hello.

#1316 is my bug (: https://travis-ci.org/g1itch/PyBitmessage/builds/405760919
I didn't take into account a quotation marks in /etc/os-release

PR #1319, opened 2018-07-21
this way the check will succeed if requirements met

t88yop:
else it will fail with keyError exception, even though everything is satisfied.

### DIFF
--- a/checkdeps.py
+++ b/checkdeps.py
@@ -134,27 +134,19 @@ if prereqs:
 
 # Install the system dependencies of optional extras_require components
 OPSYS = detectOS()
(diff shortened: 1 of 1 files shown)


PeterSurda:
Duplicate of #1317

PeterSurda:
I don't think it's PEP8. However, it looks like it does fix the bug.

g1itch:
I think this patch just masks any potential future problems with that evil expression with nested `any([`.

PeterSurda:
@g1itch After reviewing the patch, I agree. This needs to be fixed properly.

PR #1327, opened 2018-07-23
Create README_RU.md

AntonMatveyenka:
Broadening borders. Created the Russian version of README.md.


g1itch:
What's the sense of adding such localized README? When I open the page of your PyBitmessage fork with russian locale I still see the English description. Furthermore, PyBitmessage has ~ 16 localizations: https://www.transifex.com/bitmessage-project/pybitmessage/languages/ so do you think it should be 15 other README's in the project root?

dimyme:
we should not clutter the root dir .maybe find a subdir for that.

PR #1337, opened 2018-08-15
Added: Fabric task to deploy current project to remote host

coffeedogs:
Fabric task to deploy current project to remote host


g1itch:
Another stale nonsense. I leave open to remember about removing all that trash together with `Makefile`, `configure` (#843) and other nonsense.

PR #1341, opened 2018-08-27
Add src to path when running pylint

PeterSurda:
- this will allow pylint to run in the main directory without complaining about
  imports


PeterSurda:
I tested it locally as well as against the code quality checking app and it works.

PR #1344, opened 2018-09-18
better give positive advice, not negative advice

dimyme:
better give positive advice, not negative advice

victims of cybercrime / ransomware don't want to be told what NOT to do.
they need to be told WHAT to do. we should not appear to be cynical here.

https://www.interpol.int/Crime-areas/Cybercrime

CAT USA     : https://www.fbi.gov/news/stories/the-cyber-action-team
ZAC Germany : https://www.polizei.de/Polizei/DE/Einrichtungen/ZAC/zac_node.html

https://www.nomoreransom.org/en/index.html  supported by Interpol: https://www.interpol.int/Crime-areas/Cybercrime/Online-safety/Ransomware


dimyme:
ft

Issue #1335, opened 2018-08-09
 list index out of range traceback when no connects found or file corrupt.


surbhicis:
random choice need a list to Generates a random sample from a list, but in connection pool it find self.stream empty and through  list index out of range traceback:

Exception in thread Asyncore:
Traceback (most recent call last):
PyBitmessage/src/helper_random.py", line 65, in randomchoice
    return random.choice(population)  # nosec
  File "/usr/lib/python2.7/random.py", line 275, in choice
    return seq[int(self.random() * len(seq))]  # raises IndexError if seq is empty
IndexError: list index out of range

2018-08-09 13:28:22,901 - WARNING - No notification.sound plugin found

PeterSurda:
Looks like when `knownnodes` is empty, an exception is thrown.

milahu:
> knownnodes is empty

yes. 
this happens, when knownnodes.dat is empty or corrupted*,
but parsing dont raise an exception
to trigger createDefaultKnownNodes.

patch:
in knownnodes.py
after createDefaultKnownNodes()
insert:
```
    node_count = 0
    for stream in knownNodes.keys():
        for node in knownNodes[stream].keys():
            node_count += 1
    if node_count == 0:
        createDefaultKnownNodes()
    else:
        print("restored %i nodes" % node_count)

```

*corrupted?
blame the file, or blame the parser?
see knownnodes.py
```
def pickle_deserialize_old_knownnodes(source):
    knownNodes = pickle.load(source)
    for stream in knownNodes.keys():
        for node, params in knownNodes[stream].items():
            if isinstance(params, (float, int)):
                addKnownNode(stream, node, params)
```
the if condition is never satisfied,
so old nodes are not added.

somehow, old pickle and new json format got mixed up.

in my pickle-coded knownnodes.dat
params is a dict, and looks like this
```{'rating': 0, 'self': False, 'lastseen': 1538113792}```

i suggest:
1. check for duplicate nodes
2. add elif condition
```
def pickle_deserialize_old_knownnodes(source):
    knownNodes = pickle.load(source)
    for stream in knownNodes.keys():
        for node, params in knownNodes[stream].items():
            # find duplicate nodes
            found_node = False
            for old_stream in knownNodes.keys():
                for old_node in knownNodes[old_stream].keys():
                    if old_node == node:
                        found_node = True
                        break
                if found_node:
                    break
            if found_node:
                continue

            if isinstance(params, (float, int)):
                # add old format
[...]

PR #1355, opened 2018-10-02
Fix import of pickled knownnodes.dat

g1itch:
Hello.

This is a fix for bug in compatibility code for pickled `knownnodes.dat` discovered when I tried to parse @milahu comments to #1335.

### DIFF
--- a/src/knownnodes.py
+++ b/src/knownnodes.py
@@ -60,9 +60,10 @@ def pickle_deserialize_old_knownnodes(source):
     the old format was {Peer:lastseen, ...}
     the new format is {Peer:{"lastseen":i, "rating":f}}
     """
+    global knownNodes  # pylint: disable=global-statement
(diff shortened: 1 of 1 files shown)


g1itch:
Maybe I'll create another PR for second commit? I'm also going to add tests soon.

src/knownnodes.py
```diff
@@ -60,9 +60,10 @@ def pickle_deserialize_old_knownnodes(source):
     the old format was {Peer:lastseen, ...}
     the new format is {Peer:{"lastseen":i, "rating":f}}
     """
+    global knownNodes
```
PeterSurda: Please use a comment to suppress pylint warning, like

    global knownNodes # pylint: disable=global-statement

g1itch:
If `global` statement is indeed a problem, I think `knownNodes` can be turned into dictionary like object.

PR #1359, opened 2018-10-08
Fix knownnodes starvation issue

g1itch:
#1335 steps to reproduce:
 1. run BM with [existing knownnodes.dat](../blob/v0.6/src/network/connectionpool.py#L142) containing only nodes older then 28 days (or run offline long enough to outdate all nodes),
 1. wait 5 min for [cleanup](../blob/v0.6/src/class_singleCleaner.py#L127),
 1. get `IndexError` in [connectionchooser](../blob/v0.6/src/network/connectionchooser.py#L37) because `knownnodes.knownNodes[stream].keys()` returns empty list.

[Corresponding test](../../../g1itch/PyBitmessage/blob/test-mode/src/tests/core.py#L97)
[Test result](https://travis-ci.org/Bitmessage/PyBitmessage/builds/438549302#L1155)
[After this patch](https://travis-ci.org/g1itch/PyBitmessage/builds/438581399#L1143)

### DIFF
--- a/src/class_singleCleaner.py
+++ b/src/class_singleCleaner.py
@@ -55,7 +55,7 @@ class singleCleaner(threading.Thread, StoppableThread):
             ) + (
                 float(BMConfigParser().get(
                     'bitmessagesettings', 'stopresendingafterxmonths')) *
-                (60 * 60 * 24 * 365)/12)
+                (60 * 60 * 24 * 365) / 12)
         except:
             # Either the user hasn't set stopresendingafterxdays and
             # stopresendingafterxmonths yet or the options are missing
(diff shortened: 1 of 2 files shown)


src/class_singleCleaner.py
```diff
@@ -115,54 +114,28 @@ def run(self):
...
+                        (tr._translate("MainWindow", "Disk full"),
+                         tr._translate(
+                             "MainWindow",
+                             'Alert: Your disk or data storage volume'
+                             ' is full. Bitmessage will now exit.'),
+                            True)
+                    ))
+                    # FIXME redundant?
+                    if shared.daemon or not state.enableGUI:
+                        sys.exit(1)
```
PeterSurda: The reason why `os._exit()` rather than `sys.exit()` was used here is to avoid calling `shutdown.doCleanShutdown()`, which may screw up things even more if there is low diskspace.

PR #1362, opened 2018-10-10
Changes based on style and lint checks. (final_code_quality_4)

coffeedogs:
Changes based on style and lint checks. (final_code_quality_4)


coffeedogs:
Looks like pycodestyle check fails if the file doesn't exist. Not sure where the check is defined but I'm guessing a quick shell check before might work (if the shell is bash: `if [ -e $filename ] ; then pycodestyle $filename; fi`)

PeterSurda:
@coffeedogs Ok, will check. It's all written in python BTW, no shell. The algorithm for calculating new annotations did have a bug triggerd by removing a file, I thought I fixed it or maybe there's something that I missed.

PeterSurda:
Updated pylint checks to ignore E902 (IOError), requeued check.

PeterSurda:
Improved handling of removed files.

src/network/bmproto.py
```diff
@@ -320,21 +350,24 @@ def _command_inv(self, dandelion=False):
...
 
     def bm_command_dinv(self):
-        """
-        Dandelion stem announce
-        """
+        """Dandelion stem announce"""
         return self._command_inv(True)
 
     def bm_command_object(self):
+        """TBC"""
```
PeterSurda: Incoming object

src/network/bmproto.py
```diff
@@ -375,9 +413,10 @@ def _decode_addr(self):
         return self.decode_payload_content("LQIQ16sH")
 
     def bm_command_addr(self):
-        addresses = self._decode_addr()
-        for i in addresses:
-            seenTime, stream, services, ip, port = i
+        """TBC"""
```
PeterSurda:
```suggestion
        """Incoming address"""
```

src/network/bmproto.py
```diff
@@ -406,18 +445,22 @@ def bm_command_addr(self):
...
         return True
 
     def bm_command_pong(self):
-        # nothing really
+        """noop"""
+        # pylint: disable=no-self-use
         return True
 
     def bm_command_verack(self):
+        """Return True if a verack has been sent, False otherwise"""
```
PeterSurda:
```suggestion
        """Incoming verack. If already send my own verack, handshake is complete (except potentially waiting for buffers to flush), so we can continue to the main connection phase. If not sent verack yet, continue processing."""
```

src/network/bmproto.py
```diff
@@ -406,18 +445,22 @@ def bm_command_addr(self):
         return True
 
     def bm_command_portcheck(self):
+        """Add a job port of a peer"""
         portCheckerQueue.put(state.Peer(self.destination, self.peerNode.port))
         return True
 
     def bm_command_ping(self):
+        """Respond to a ping"""
```
PeterSurda:
```suggestion
        """Incoming ping, respond to it."""
```

src/network/bmproto.py
```diff
@@ -406,18 +445,22 @@ def bm_command_addr(self):
         return True
 
     def bm_command_portcheck(self):
+        """Add a job port of a peer"""
```
PeterSurda:
```suggestion
        """Incoming port check request, queue it."""
```

src/network/bmproto.py
```diff
@@ -269,16 +296,19 @@ def decode_simple(self, char="v"):
...
 
     def bm_command_error(self):
+        """Decode an error message and log it"""
+        # pylint: disable=unused-variable
         fatalStatus, banTime, inventoryVector, errorText = self.decode_payload_content("vvlsls")
         logger.error("%s:%i error: %i, %s", self.destination.host, self.destination.port, fatalStatus, errorText)
         return True
 
     def bm_command_getdata(self):
+        """Decode object data, conditionally append a newly created object to the write buffer"""
```
PeterSurda:
```suggestion
        """Incoming request for object(s), if we have them and some other conditions are fulfilled, append them to the write queue."""
```

PS. the wire protocol works very similarly to Bitcoin, so even if the command isn't documented in Bitmessage Wiki, just look at how it works in Bitcoin.

src/network/bmproto.py
```diff
@@ -269,16 +296,19 @@ def decode_simple(self, char="v"):
...
         items = self.decode_payload_content("l32s")
-        # skip?
+        # .. todo:: skip?
```
PeterSurda: I don' think this is a todo. It's just an explanation of what the `if`is doing.

src/network/bmproto.py
```diff
@@ -269,16 +296,19 @@ def decode_simple(self, char="v"):
...
         if time.time() < self.skipUntil:
             return True
-        #TODO make this more asynchronous
+        # .. todo:: make this more asynchronous
```
PeterSurda: This however IS a todo. This loop may take a while, during which time a lot of other stuff is locked. Should be refactored in the future.

src/network/bmproto.py
```diff
@@ -406,18 +445,22 @@ def bm_command_addr(self):
...
         return True
 
     def bm_command_ping(self):
+        """Respond to a ping"""
         self.append_write_buf(protocol.CreatePacket('pong'))
         return True
 
     def bm_command_pong(self):
-        # nothing really
+        """noop"""
```
PeterSurda:
```suggestion
        """Incoming pong, ignore it. PyBitmessage pings connections after about 5 minutes of inactivity, and leaves it to the TCP stack to handle actual timeouts. So there is no need to do anything when a pong arrives."""
```

src/network/bmproto.py
```diff
@@ -428,6 +471,8 @@ def bm_command_verack(self):
         return True
 
     def bm_command_version(self):
+        """Determine and log protocol version and other details"""
```
PeterSurda:
```suggestion
        """Incoming version. Parse and log, remember important things, like streams, bitfields, etc."""
```

src/network/bmproto.py
```diff
@@ -438,17 +483,20 @@ def bm_command_version(self):
         logger.debug("my external IP: %s", self.sockNode.host)
         logger.debug("remote node incoming address: %s:%i", self.destination.host, self.peerNode.port)
         logger.debug("user agent: %s", self.userAgent)
-        logger.debug("streams: [%s]", ",".join(map(str,self.streams)))
+        logger.debug("streams: [%s]", ",".join(map(str, self.streams)))
         if not self.peerValidityChecks():
-            # TODO ABORT
+            # .. todo:: ABORT
```
PeterSurda: I think this isn't necessary. Returning here will prevent verack being sent. Without that the handshake is considered incomplete, other commands are ignored and the connection will timeout after 20 seconds. The error in the write buffer has this time to make it through. Probably a leftover from early stages, please remove this "todo".

src/network/bmproto.py
```diff
@@ -461,69 +509,76 @@ def bm_command_version(self):
         return True
 
     def peerValidityChecks(self):
+        """Check the validity of peers"""
```
PeterSurda:
```suggestion
        """Check the validity of the peer"""
```

src/network/bmproto.py
```diff
@@ -461,69 +509,76 @@ def bm_command_version(self):
...
+                    errorText="I'm connected to myself. Closing connection."))
+            logger.debug("Closed connection to %s because I'm connected to myself.",
+                         str(self.destination))
             return False
 
         return True
 
     @staticmethod
     def assembleAddr(peerList):
+        """iBuild up a packed address"""
```
PeterSurda: Typo?

src/socks/__init__.py
```diff
@@ -1,4 +1,9 @@
-"""SocksiPy - Python SOCKS module.
+# pylint: disable=too-many-arguments,global-statement,too-many-branches
```
PeterSurda: This file will be removed once asyncore Socks resolver is finished, so I don't care about it. For P2P connections over Socks we're already using my port of this to asyncore.

src/network/bmproto.py
coffeedogs: BTW, the rST syntax so that the sphinx documentation can both format and collect a list of all todo items.

src/network/bmproto.py
```diff
@@ -70,7 +88,10 @@ def bm_proto_reset(self):
         self.object = None
 
     def state_bm_header(self):
-        self.magic, self.command, self.payloadLength, self.checksum = protocol.Header.unpack(self.read_buf[:protocol.Header.size])
+        """Predicate (with logging) to indicate the prescence of a header"""
```
PeterSurda: It doesn't indicate a header presence, it expects it, according to the protocol specification.

src/network/bmproto.py
```diff
@@ -85,8 +106,9 @@ def state_bm_header(self):
             self.invalid = True
         self.set_state("bm_command", length=protocol.Header.size, expectBytes=self.payloadLength)
         return True
-        
+
     def state_bm_command(self):
+        """Predicate (with logging) to indicate the presence of a command"""
```
PeterSurda: Again, doesn't indicate, but expects, the presence of a command.

src/class_sqlThread.py
```diff
@@ -461,11 +563,27 @@ def run(self):
...
                         os._exit(0)
                     else:
-                        logger.fatal('Major error occurred when trying to execute a SQL statement within the sqlThread. Please tell Atheros about this error message or post it in the forum! Error occurred while trying to execute statement: "%s"  Here are the parameters; you might want to censor this data with asterisks (***) as it can contain private information: %s. Here is the actual error message thrown by the sqlThread: %s', str(item), str(repr(parameters)),  str(err))
+                        logger.fatal(
+                            'Major error occurred when trying to execute a SQL statement within the sqlThread. Please'
+                            ' tell Atheros about this error message or post it in the forum! Error occurred while'
+                            ' trying to execute statement: "%s"  Here are the parameters; you might want to censor'
+                            ' this data with asterisks (***) as it can contain private information: %s. Here is the'
+                            ' actual error message thrown by the sqlThread: %s',
+                            str(item),
```
g1itch: `%s` causes `str()` itself - no need to convert args.
use `%r` instead of str(repr(parameters))

PR #1363, opened 2018-10-10
Changes based on style and lint checks. (final_code_quality_5)

coffeedogs:
Changes based on style and lint checks. (final_code_quality_5)

### DIFF
--- a/src/network/asyncore_pollchoose.py
+++ b/src/network/asyncore_pollchoose.py
@@ -1,6 +1,11 @@
 # -*- Mode: Python -*-
 #   Id: asyncore.py,v 2.51 2000/09/07 22:29:26 rushing Exp
 #   Author: Sam Rushing <rushing@nightmare.com>
+# pylint: disable=too-many-statements,too-many-branches,no-self-use,too-many-lines,attribute-defined-outside-init
+# pylint: disable=global-statement
+"""
+src/network/asyncore_pollchoose.py
+==================================
 
 # ======================================================================
 # Copyright 1996 by Sam Rushing
@@ -25,7 +30,7 @@
 # CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 # ======================================================================
 
-"""Basic infrastructure for asynchronous socket service clients and servers.
+Basic infrastructure for asynchronous socket service clients and servers.
 
 There are only two ways to have a program on a single processor do "more
 than one thing at a time".  Multi-threaded programming is the simplest and
@@ -46,22 +51,20 @@ many of the difficult problems for you, making the task of building
 sophisticated high-performance network servers and clients a snap.
 """
 
-# randomise object order for bandwidth balancing
-import random
+import os
 import select
 import socket
 import sys
 import time
-from threading import current_thread
 import warnings
+from errno import (
+    EADDRINUSE, EAGAIN, EALREADY, EBADF, ECONNABORTED, ECONNREFUSED, ECONNRESET, EHOSTUNREACH, EINPROGRESS, EINTR,
+    EINVAL, EISCONN, ENETUNREACH, ENOTCONN, ENOTSOCK, EPIPE, ESHUTDOWN, ETIMEDOUT, EWOULDBLOCK, errorcode
+)
+from threading import current_thread
 
-import os
 import helper_random
-from errno import EALREADY, EINPROGRESS, EWOULDBLOCK, ECONNRESET, EINVAL, \
-     ENOTCONN, ESHUTDOWN, EISCONN, EBADF, ECONNABORTED, EPIPE, EAGAIN, \
-     ECONNREFUSED, EHOSTUNREACH, ENETUNREACH, ENOTSOCK, EINTR, ETIMEDOUT, \
-     EADDRINUSE, \
-     errorcode
+
 try:
     from errno import WSAEWOULDBLOCK
 except (ImportError, AttributeError):
@@ -75,13 +78,15 @@ try:
 except (ImportError, AttributeError):
     WSAECONNRESET = ECONNRESET
 try:
-    from errno import WSAEADDRINUSE
+    # Desirable side-effects on Windows; imports winsock error numbers
+    from errno import WSAEADDRINUSE  # pylint: disable=unused-import
 except (ImportError, AttributeError):
     WSAEADDRINUSE = EADDRINUSE
 
-_DISCONNECTED = frozenset((ECONNRESET, ENOTCONN, ESHUTDOWN, ECONNABORTED, EPIPE,
-                           EBADF, ECONNREFUSED, EHOSTUNREACH, ENETUNREACH, ETIMEDOUT,
-                           WSAECONNRESET))
+
+_DISCONNECTED = frozenset((
+    ECONNRESET, ENOTCONN, ESHUTDOWN, ECONNABORTED, EPIPE, EBADF, ECONNREFUSED,
+    EHOSTUNREACH, ENETUNREACH, ETIMEDOUT, WSAECONNRESET))
 
 OP_READ = 1
 OP_WRITE = 2
@@ -91,16 +96,20 @@ try:
 except NameError:
     socket_map = {}
 
+
 def _strerror(err):
     try:
         return os.strerror(err)
     except (ValueError, OverflowError, NameError):
         if err in errorcode:
             return errorcode[err]
-        return "Unknown error %s" %err
+        return "Unknown error %s" % err
+
 
 class ExitNow(Exception):
+    """We don't use directly but may be necessary as we replace asyncore due to some library raising or expecting it"""
     pass
+
 
 _reraised_exceptions = (ExitNow, KeyboardInterrupt, SystemExit)
 
@@ -113,28 +122,38 @@ uploadTimestamp = 0
 uploadBucket = 0
 sentBytes = 0
 
+
 def read(obj):
+    """Event to read from the object, i.e. its network socket."""
+
     if not can_receive():
         return
     try:
         obj.handle_read_event()
     except _reraised_exceptions:
         raise
-    except:
+    except BaseException:
         obj.handle_error()
 
+
 def write(obj):
+    """Event to write to the object, i.e. its network socket."""
+
     if not can_send():
         return
     try:
         obj.handle_write_event()
     except _reraised_exceptions:
         raise
-    except:
+    except BaseException:
         obj.handle_error()
 
+
 def set_rates(download, upload):
+    """Set throttling rates"""
+
     global maxDownloadRate, maxUploadRate, downloadBucket, uploadBucket, downloadTimestamp, uploadTimestamp
+
     maxDownloadRate = float(download) * 1024
     maxUploadRate = float(upload) * 1024
     downloadBucket = maxDownloadRate
@@ -142,14 +161,24 @@ def set_rates(download, upload):
     downloadTimestamp = time.time()
     uploadTimestamp = time.time()
 
+
 def can_receive():
+    """Predicate indicating whether the download throttle is in effect"""
+
     return maxDownloadRate == 0 or downloadBucket > 0
 
+
 def can_send():
+    """Predicate indicating whether the upload throttle is in effect"""
+
     return maxUploadRate == 0 or uploadBucket > 0
 
+
 def update_received(download=0):
+    """Update the receiving throttle"""
+
     global receivedBytes, downloadBucket, downloadTimestamp
+
     currentTimestamp = time.time()
     receivedBytes += download
     if maxDownloadRate > 0:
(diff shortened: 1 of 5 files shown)


coffeedogs:
`src/defaults.py` is made use of by `src/bitmessagemain.py`, though  I've bumped that to a later PR since otherwise it was going to start pulling in the kivy files in that I haven't finished yet.

src/network/asyncore_pollchoose.py
```diff
@@ -75,13 +78,15 @@
 except (ImportError, AttributeError):
     WSAECONNRESET = ECONNRESET
 try:
-    from errno import WSAEADDRINUSE
+    # side-effects on Windows or cruft?
```
PeterSurda: Yea this is workaround for some Windows crap, it uses winsock which is kind of similar to unix' socket, but it can have its own error numbers, and these aren't defined when running python on other OSes. Maybe there is a nicer workaround but for the time being let's keep it as it is.

src/network/asyncore_pollchoose.py
```diff
@@ -631,7 +704,11 @@ def connect(self, address):
             raise socket.error(err, errorcode[err])
 
     def accept(self):
-        # XXX can return either an address pair or None
+        """
+        Accept incoming connections
+
+        .. todo:: FIXME: can return either an address pair or None
```
PeterSurda: I'm not sure it's a FIXME, it could also be just a comment. However, it's not a big deal so I'm accepting this hunk.

src/network/tcp.py
```diff
@@ -67,18 +68,19 @@ def __init__(self, address=None, sock=None):
...
+            protocol.checkIPAddress(encodedAddr, True),
+            not protocol.checkSocksIP(self.destination.host)
+        ])
+        ObjectTracker.__init__(self)  # pylint: disable=non-parent-init-called
         self.bm_proto_reset()
         self.set_state("bm_header", expectBytes=protocol.Header.size)
 
-    def antiIntersectionDelay(self, initial = False):
+    def antiIntersectionDelay(self, initial=False):
+        """TBC"""
```
PeterSurda: This is a defense against the so called intersection attacks. It is called when you notice peer is requesting non-existing objects, or right after the connection is established. It will estimate how long an object will take to propagate across the network, and skip processing "getdata" requests until then. This means an attacker only has one shot per IP to perform the attack.

src/network/tcp.py
```diff
@@ -93,12 +95,14 @@ def antiIntersectionDelay(self, initial = False):
                 self.skipUntil = time.time() + delay
 
     def state_connection_fully_established(self):
+        """TBC"""
```
PeterSurda: State after the bitmessage protocol handshake is completed (version/verack exchange, and if both side support TLS, the TLS handshake as well).

src/network/tcp.py
```diff
@@ -93,12 +95,14 @@ def antiIntersectionDelay(self, initial = False):
...
         self.set_connection_fully_established()
         self.set_state("bm_header")
         self.bm_proto_reset()
         return True
 
     def set_connection_fully_established(self):
+        """TBC"""
```
PeterSurda: Initiate inventory synchronisation.

src/network/tcp.py
```diff
@@ -113,50 +117,50 @@ def set_connection_fully_established(self):
         self.sendBigInv()
 
     def sendAddr(self):
+        """TBC"""
```
PeterSurda: Send a partial list of known addresses to peer.

src/network/tcp.py
```diff
@@ -113,50 +117,50 @@ def set_connection_fully_established(self):
...
-        for substream in addrs.keys():
+        for substream in addrs:
             for peer, params in addrs[substream]:
                 templist.append((substream, peer, params["lastseen"]))
-        if len(templist) > 0:
+        if templist:
             self.append_write_buf(BMProto.assembleAddr(templist))
 
     def sendBigInv(self):
+        """TBC"""
```
PeterSurda: Send hashes of all inventory objects, chunked as the protocol has a per-command limit.

src/network/tcp.py
```diff
@@ -113,50 +117,50 @@ def set_connection_fully_established(self):
...
         def sendChunk():
+            """TBC"""
```
PeterSurda: Send one chunk of inv entries in one command

src/network/tcp.py
```diff
@@ -193,20 +196,26 @@ def sendChunk():
         sendChunk()
 
     def handle_connect(self):
+        """TBC"""
```
PeterSurda: Callback for TCP connection being established.

src/network/tcp.py
```diff
@@ -193,20 +196,26 @@ def sendChunk():
...
+                self.destination.host,
+                self.destination.port,
+                network.connectionpool.BMConnectionPool().streams,
+                False,
+                nodeid=self.nodeid))
         self.connectedAt = time.time()
         receiveDataQueue.put(self.destination)
 
     def handle_read(self):
+        """TBC"""
```
PeterSurda: Callback for reading from a socket

src/network/tcp.py
```diff
@@ -218,9 +227,11 @@ def handle_read(self):
         receiveDataQueue.put(self.destination)
 
     def handle_write(self):
+        """TBC"""
```
PeterSurda: Callback for writing to a socket

src/network/tcp.py
```diff
@@ -218,9 +227,11 @@ def handle_read(self):
...
         TLSDispatcher.handle_write(self)
 
     def handle_close(self):
+        """TBC"""
```
PeterSurda: Callback for connection being closed.

src/network/tcp.py
```diff
@@ -231,37 +242,55 @@ def handle_close(self):
 
 
 class Socks5BMConnection(Socks5Connection, TCPConnection):
+    """TBC"""
```
PeterSurda: SOCKS5 wrapper for TCP connections

src/network/tcp.py
```diff
@@ -231,37 +242,55 @@ def handle_close(self):
...
+
     def __init__(self, address):
         Socks5Connection.__init__(self, address=address)
         TCPConnection.__init__(self, address=address, sock=self.socket)
         self.set_state("init")
 
     def state_proxy_handshake_done(self):
+        """TBC"""
```
PeterSurda: State when SOCKS5 connection succeeds, we need to send a Bitmessage handshake to peer.

src/network/tcp.py
```diff
@@ -231,37 +242,55 @@ def handle_close(self):
...
+                self.destination.port,
+                network.connectionpool.BMConnectionPool().streams,
+                False,
+                nodeid=self.nodeid))
         self.set_state("bm_header", expectBytes=protocol.Header.size)
         return True
 
 
 class Socks4aBMConnection(Socks4aConnection, TCPConnection):
+    """TBC"""
```
PeterSurda: SOCKS4a wrapper for TCP connections

src/network/tcp.py
```diff
@@ -231,37 +242,55 @@ def handle_close(self):
...
+
     def __init__(self, address):
         Socks4aConnection.__init__(self, address=address)
         TCPConnection.__init__(self, address=address, sock=self.socket)
         self.set_state("init")
 
     def state_proxy_handshake_done(self):
+        """TBC"""
```
PeterSurda: State when SOCKS4a connection succeeds, we need to send a Bitmessage handshake to peer.

src/network/tcp.py
```diff
@@ -231,37 +242,55 @@ def handle_close(self):
...
+                network.connectionpool.BMConnectionPool().streams,
+                False,
+                nodeid=self.nodeid))
         self.set_state("bm_header", expectBytes=protocol.Header.size)
         return True
 
 
 class TCPServer(AdvancedDispatcher):
-    def __init__(self, host='127.0.0.1', port=8444):
+    """TBC"""
```
PeterSurda: TCP connection server for Bitmessage protocol

src/network/tcp.py
```diff
@@ -284,20 +313,22 @@ def __init__(self, host='127.0.0.1', port=8444):
         self.listen(5)
 
     def is_bound(self):
+        """TBC"""
```
PeterSurda: Is the socket bound?

src/network/tcp.py
```diff
@@ -284,20 +313,22 @@ def __init__(self, host='127.0.0.1', port=8444):
...
         try:
             return self.bound
         except AttributeError:
             return False
 
     def handle_accept(self):
+        """TBC"""
```
PeterSurda: Incoming connection callback

src/network/asyncore_pollchoose.py
coffeedogs: The todo was added to consider the fixme. I consider it now well considered. I've formatted it as part of the docstring.

src/network/asyncore_pollchoose.py
coffeedogs: Updated in next commit...

src/upnp.py
```diff
@@ -283,19 +321,24 @@ def sendSearchRouter(self):
             logger.exception("UPnP send query failed")
 
     def createPortMapping(self, router):
-        from debug import logger
+        """Add a port mapping"""
 
         for i in range(50):
             try:
-                routerIP, = unpack('>I', socket.inet_aton(router.address))
+                unpack('>I', socket.inet_aton(router.address))
```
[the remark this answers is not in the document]
coffeedogs: OK, removed. Where I found unused variables involving unpack I left in the unpacks in case there was some state to do with unpacking lying around.

src/network/asyncore_pollchoose.py
g1itch: see 533df80

src/upnp.py
```diff
@@ -230,14 +263,16 @@ def run(self):
                         logger.debug("Found UPnP router at %s", ip)
                         self.routers.append(newRouter)
                         self.createPortMapping(newRouter)
-                        queues.UISignalQueue.put(('updateStatusBar', tr._translate("MainWindow",'UPnP port mapping established on port %1').arg(str(self.extPort))))
+                        queues.UISignalQueue.put(('updateStatusBar', tr._translate(
+                            "MainWindow", 'UPnP port mapping established on port %1'
+                        ).arg(str(self.extPort))))
                         # retry connections so that the submitted port is refreshed
                         with shared.alreadyAttemptedConnectionsListLock:
```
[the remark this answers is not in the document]
coffeedogs: OK, removed that context.

src/network/asyncore_pollchoose.py
```diff
@@ -113,43 +122,63 @@ class ExitNow(Exception):
 uploadBucket = 0
 sentBytes = 0
 
+
 def read(obj):
+    """Read an event from the object"""
```
PeterSurda:
```suggestion
    """Event to read from the object, i.e. its network socket."""
```

src/network/asyncore_pollchoose.py
```diff
@@ -113,43 +122,63 @@ class ExitNow(Exception):
...
         obj.handle_read_event()
     except _reraised_exceptions:
         raise
-    except:
+    except BaseException:
         obj.handle_error()
 
+
 def write(obj):
+    """Write an event to the object"""
```
PeterSurda:
```suggestion
    """Event to write to the object, i.e. its network socket."""
```

src/network/asyncore_pollchoose.py
```diff
@@ -425,13 +479,15 @@ def kqueue_poller(timeout=0.0, map=None):
         current_thread().stop.wait(timeout)
 
 
-def loop(timeout=30.0, use_poll=False, map=None, count=None, 
-         poller=None):
+def loop(timeout=30.0, use_poll=False, map=None, count=None, poller=None):
+    """Poll in a loop, forever if count is None"""
```
PeterSurda:
```suggestion
    """Poll in a loop, until count or timeout is reached"""
```

src/network/tcp.py
```diff
@@ -67,18 +68,25 @@ def __init__(self, address=None, sock=None):
...
-            self.local = True
-        else:
-            self.local = False
-        #shared.connectedHostsList[self.destination] = 0
-        ObjectTracker.__init__(self)
+        self.local = all([
+            protocol.checkIPAddress(encodedAddr, True),
+            not protocol.checkSocksIP(self.destination.host)
+        ])
+        ObjectTracker.__init__(self)  # pylint: disable=non-parent-init-called
```
PeterSurda: It looks like `ObjectTracker` is used as a mixin, but it has an `__init__`. Other than suppressing the pylint warning, I'm not sure there's anything else to do. We could rename the `__init__` to `_fake_init`, but then we would have to suppress `attribute-defined-outside-init`. I would leave it as it is unless there's something I'm missing.

coffeedogs: I would expect a mixin class to be defined as a parent: `class TCPConnection(BMProto, TLSDispatcher, ObjectTracker):` and for the `__init__`s to be called in an appropriate order, each seeing `*args` and `**kwargs` and each updating self as they go, passing off initialisation to the next parent class using `super(ThisClass, self).__init__(*args, **kwargs)`.

However, if this is working as intended, perhaps re-initialisation is both happening and required? Either way, fixing this bad code smell will require some consideration, refactoring and testing.

g1itch: `super()` wont work without refactoring, I tried.

PR #1364, opened 2018-10-10
Changes based on style and lint checks. (final_code_quality_6)

coffeedogs:
Changes based on style and lint checks. (final_code_quality_6)

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -1,11 +1,18 @@
+# pylint: disable=too-many-locals,too-many-lines,no-self-use,too-many-public-methods,too-many-branches
+# pylint: disable=too-many-statements
+"""
+src/api.py
+==========
+
 # Copyright (c) 2012-2016 Jonathan Warren
 # Copyright (c) 2012-2018 The Bitmessage developers
 
-"""
 This is not what you run to run the Bitmessage API. Instead, enable the API
 ( https://bitmessage.org/wiki/API ) and optionally enable daemon mode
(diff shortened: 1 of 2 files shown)


src/api.py
```diff
@@ -1044,18 +1118,30 @@ def HandleDisseminatePreEncryptedMsg(self, params):
...
         target = 2**64 / (
-            (len(encryptedPayload) + requiredPayloadLengthExtraBytes + 8)
-            * requiredAverageProofOfWorkNonceTrialsPerByte)
+            (
+                len(encryptedPayload) + requiredPayloadLengthExtraBytes + 8
+            ) * requiredAverageProofOfWorkNonceTrialsPerByte
+        )
         with shared.printLock:
-            print '(For msg message via API) Doing proof of work. Total required difficulty:', float(requiredAverageProofOfWorkNonceTrialsPerByte) / defaults.networkDefaultProofOfWorkNonceTrialsPerByte, 'Required small message difficulty:', float(requiredPayloadLengthExtraBytes) / defaults.networkDefaultPayloadLengthExtraBytes
+            print(
```
PeterSurda: @coffeedogs This print should be removed, however this is a separate task so I created a new issue and assigned it to you.

PeterSurda:
I think the commit isn't signed, other than that it's ready to go.

coffeedogs:
Guessing that wasn't signed since I only had a project-level enforce set in git config and was using a separate directory to do this rebasing. I've now set git's enforced signing to be global, thanks for the vigilance!

Issue #1374, opened 2018-10-12
Unused code in shared and protocol modules


g1itch:
Hello.

The functions `checkAndShareObjectWithPeers`, `_checkAndShareUndefinedObjectWithPeers`, `_checkAndShareMsgWithPeers`, `_checkAndShareGetpubkeyWithPeers`, `_checkAndSharePubkeyWithPeers`, `_checkAndShareBroadcastWithPeers` are same in `protocol` and `shared` modules and used from latter as I can see.

Remove it from `protocol`?

g1itch:
The only place where `checkAndShareObjectWithPeers()` appears is [class_objectProcessor.py#L751](../blob/v0.6/src/class_objectProcessor.py#L751) but `broadcastToSendDataQueues` seems to be useless, because I cannot find where `sendDataQueues` are handled.

PeterSurda:
Yea it probably does need extra work as the `BMObject` constructor requires decoded arguments, and the decoder only works on buffers. So some redesign is necessary if we're to avoid duplication of code.

PeterSurda:
Maybe create a fake `BMProto` object `o`, copy the buffer from the ACK buffer (or just reference it), and call `o.bm_command_object()` to process it. Probably some additional tests are necessary.

g1itch:
I found that I do not understand where objects are broadcasted to the network ):

g1itch:
I have a rather controversial solution: 03ee14e9.

PR #1377, opened 2018-10-19
Removed obsolete code from protocol, shared

g1itch:
Hello!

This changes is for #1374. I removed unused code and rewritten ACK sending like dirty hack. Maybe you suggest a better solution?


PeterSurda:
It isn't so bad, maybe create a new child for `BMProto`, say `BMStringParser`, that can do this instead of creating a fake `UDPSocket`?

g1itch:
@PeterSurda that's it. Still testing it locally...

src/network/bmproto.py
```diff
@@ -575,3 +576,32 @@ def handle_close(self):
...
+        self.bm_proto_reset()
+        self.payload = data
+        try:
+            self.bm_command_object()
+        except BMObjectAlreadyHaveError:
+            pass  # maybe the same msg received on different nodes
+        except BMObjectExpiredError:
+            logger.debug(
+                'Sending ACK failure (expired): %s', hexlify(data))
+        except Exception as e:
```
PeterSurda: `BMObjectUnwantedStreamError` probably needs additional coding if the creator of the ACK isn't in of of my streams. Which could happen in the future. But this should probably be addressed in `bm_command_object` and for the time being doesn't matter. It will be sufficiently tested once streams are implemented.

src/network/udp.py
```diff
@@ -60,48 +52,45 @@ def set_socket_reuse(self):
...
+        # return BMProto.bm_command_inv(self)
         return True
-#        return BMProto.bm_command_inv(self)
-
-    def bm_command_object(self):
-        return BMProto.bm_command_object(self)
 
     def bm_command_addr(self):
-#        BMProto.bm_command_object(self)
+        # BMProto.bm_command_object(self)
```
PeterSurda: this line can be removed

src/network/udp.py
```diff
@@ -126,51 +115,54 @@ def writable(self):
...
+            retval = self.socket.sendto(
+                self.write_buf, ('<broadcast>', self.port))
         except socket.error as e:
-            logger.error("socket error on sendato: %s", str(e))
+            logger.error("socket error on sendato: %s", e)
             retval = 0
         self.slice_write_buf(retval)
 
 
 if __name__ == "__main__":
```
PeterSurda: this whole section can be removed, there is no UDP support in Socks.

PeterSurda:
You can remove the lines I pointed out or merge right away.

Issue #1378, opened 2018-10-24
Questions about API


g1itch:
Hello

I used API to test ACK sending so I finally read the `api` module completely. And decided to [refactor it](../compare/v0.6...g1itch:api). Now I have several questions:

1. If `MySimpleXMLRPCRequestHandler` is XML-RPC server why it's methods return json strings?
1. If it's desirable to return JSON, maybe it should be JSON-RPC server?
1. Why so inconsistent return values? [success](../blob/v0.6/src/api.py#L624), [Added subscription](../blob/v0.6/src/api.py#L1145), [done](../blob/v0.6/src/api.py#L1386) seems to have same meaning. [sentMessage](../blob/v0.6/src/api.py#L842) and [sentMessages](../blob/v0.6/src/api.py#L873) counter intuitively both are lists. [getInboxMessagesByReceiver](../blob/v0.6/src/api.py#L820) doesn't return read status. Also some methods may return `None`.
1. Wouldn't it be better to return error status like `{"error": "Number and descriptive text"}` instead of simple string?

JSON string vs XML object:
```xml
<?xml version='1.0'?>
<methodResponse>
<params>
<param>
<value><string>{
    "numberOfBroadcastsProcessed": 7,
    "networkStatus": "notConnected",
    "softwareName": "PyBitmessage",
    "softwareVersion": "0.6.3.2",
    "networkConnections": 3,
    "numberOfMessagesProcessed": 69,
    "numberOfPubkeysProcessed": 1
}</string></value>
</param>
</params>
</methodResponse>
```
```python
'{\n    "numberOfBroadcastsProcessed": 7,\n    "networkStatus": "notConnected",\n    "softwareName": "PyBitmessage",\n    "softwareVersion": "0.6.3.2",\n    "networkConnections": 3,\n    "numberOfMessagesProcessed": 69,\n    "numberOfPubkeysProcessed": 1\n}'
```

```xml
<?xml version='1.0'?>
<methodResponse>
<params>
<param>
<value><struct>
<member>
<name>numberOfBroadcastsProcessed</name>
<value><int>0</int></value>
</member>
<member>
<name>networkStatus</name>
<value><string>connectedButHaveNotReceivedIncomingConnections</string></value>
</member>
<member>
<name>softwareName</name>
<value><string>PyBitmessage</string></value>
</member>
<member>
<name>softwareVersion</name>
<value><string>0.6.3.2</string></value>
</member>
<member>
<name>networkConnections</name>
[...]

PR #1382, opened 2018-10-31
API refactoring

g1itch:
Hello!

This is for you review. I'm not finished yet. At least I'm going to cover all methods by tests, write docstrings and deal with error codes.

Maybe we can continue the discussion started in #1378 here.

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -3,9 +3,58 @@
 # pylint: disable=too-many-lines,no-self-use,unused-variable,unused-argument
 
 """
-This is not what you run to run the Bitmessage API. Instead, enable the API
-( https://bitmessage.org/wiki/API ) and optionally enable daemon mode
-( https://bitmessage.org/wiki/Daemon ) then run bitmessagemain.py.
+This is not what you run to start the Bitmessage API.
+Instead, `enable the API <https://bitmessage.org/wiki/API>`_
+and optionally `enable daemon mode <https://bitmessage.org/wiki/Daemon>`_
+then run the PyBitmessage.
+
+The PyBitmessage API is provided either as
+`XML-RPC <http://xmlrpc.scripting.com/spec.html>`_ or
+`JSON-RPC <https://www.jsonrpc.org/specification>`_ like in bitcoin.
+It's selected according to 'apivariant' setting in config file.
+
+Special value ``apivariant=legacy`` is to mimic the old pre 0.6.3
+behaviour when any results are returned as strings of json.
+
+.. list-table:: All config settings related to API:
+  :header-rows: 0
+
+  * - apienabled = true
+    - if 'false' the `singleAPI` wont start
+  * - apiinterface = 127.0.0.1
+    - this is the recommended default
+  * - apiport = 8442
+    - the API listens apiinterface:apiport if apiport is not used,
+      random in range (32767, 65535) otherwice
+  * - apivariant = xml
+    - current default for backward compatibility, 'json' is recommended
+  * - apiusername = username
+    - set the username
+  * - apipassword = password
+    - and the password
+  * - apinotifypath =
+    - not really the API setting, this sets a path for the executable to be ran
+      when certain internal event happens
+
+To use the API concider such simple example:
+
+.. code-block:: python
+
+    import jsonrpclib
+
+    from pybitmessage import bmconfigparser, helper_startup
+
+    helper_startup.loadConfig()  # find and load local config file
+    conf = bmconfigparser.BMConfigParser()
+    api_uri = "http://%s:%s@127.0.0.1:8442/" % (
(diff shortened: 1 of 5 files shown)


src/api.py
```diff
@@ -1335,148 +1172,96 @@ def HandleClientStatus(self, params):  # pylint: disable=unused-argument
...
+            _fault = APIError(
+                26, 'Data contains a malformed varint. Some details: %s' %
+                e)
         except Exception as e:
             logger.exception(e)
+            _fault = APIError(21, 'Unexpected API Failure - %s' % e)
+
+        if _fault:
+            if self.config.safeGet(
+                    'bitmessagesettings', 'apivariant') == 'legacy':
```
PeterSurda: I wouldn't use `legacy` but `xmlrpc` and `jsonrpc` or just `xml`/`json`

g1itch: The current implementation is neither XML-RPC nor JSON-RPC. It returns any result or error as json strings inside of XML. So I used word 'legacy' for the variant which mimics current API behaviour with two exceptions: it returns response 401 if user authentication fails and elements of the result of `getInboxMessagesByReceiver` have "read" field (e456707, 86ab0d3).

g1itch:
`no-member` probably is the [pylint bug](../../../PyCQA/pylint/issues/378).

src/bmconfigparser.py
```diff
@@ -7,7 +7,6 @@
 import os
```
omkar1117: I have seen only copyfile method from shutil.

@g1itch can you import the copyfile method  from shutil
```
from shutil import copyfile
```
instead of shutil library.

PeterSurda:
Any plans on finishing this, should I review it now?

g1itch:
> Any plans on finishing this, should I review it now?

Yes. Give me a couple of days.

PeterSurda:
Do you still want to have this merged?

g1itch:
> Do you still want to have this merged?

Off course, but I should probably cover all API commands by tests to be sure that it doesn't break anything.

PeterSurda:
Let me know when it's ready to be merged so that I can do a final review.

g1itch:
> Let me know when it's ready to be merged so that I can do a final review.

I think it's ready. Though some of the questions from #1378 are still actual.

g1itch:
Apparently the tests from this branch fail in several places on branch v0.6. So the branch introduces backward incompatibilities. More of them are commented with "FIXME" in the code.

Initially I planned to test all apivariants but stuck on changing settings. So all api tests currently are for 'legacy' apivariant.

src/tests/test_api.py
```diff
@@ -157,23 +188,136 @@ def test_addressbook(self):
...
+            json.loads(self.api.listSubscriptions())['subscriptions'], [])
+
+    def test_send(self):
+        """Test message sending"""
+        # self.api.createDeterministicAddresses(self._seed, 1, 4)
+        addr = self._add_random_address('random_2')
+        msg = base64.encodestring('test message')
+        msg_subject = base64.encodestring('test_subject')
+        ackdata = self.api.sendMessage(
+            'BM-2cWzSnwjJ7yRP3nLEWUV5LisTZyREWSzUK', addr, msg_subject, msg)
```
PeterSurda, line 218: You can use `BM-2cVWtdUzPwF7UNGDrZftWuHWiJ6xxBpiSP`. There is an autoresponder running there, it will send all messages back.

PeterSurda (approved):
I haven't tested it but it looks ok. There are a couple of typos in the docstrings, and I would like to add a couple of more checks and UISignaller messages but these aren't important, and cone be done in the future and I don't want to review such a huge chunk in one go again. To me the most important thing is it doesn't appear to break anything that was working. Very well done, this really needed refactoring.

src/tests/test_api.py
g1itch: I don't think current code is ready for such complex test. To achieve it we should wait for full sync (not sure how to determine it), send a message and wait up to 2 min for reply. And of course also turn off this: https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/bitmessagemain.py#L231

g1itch:
> I haven't tested it but it looks ok. There are a couple of typos in the docstrings, and I would like to add a couple of more checks and UISignaller messages but these aren't important, and can be done in the future and I don't want to review such a huge chunk in one go again. To me the most important thing is it doesn't appear to break anything that was working. Very well done, this really needed refactoring.

`UISignaller` is the separate pain. I think it needs refactoring too. Qt signals shouldn't go through queue if they sent from one widget to another - only for communication between core and UI. Messages like `setStatusIcon` should be changed to something like `connected`, `disconnected`. And it's the place where we can receive signal about the synchronization end.

g1itch:
OK. So I merge it.

PR #1384, opened 2018-11-01
Fix #1383 - formatting artifact in 7ed459b

### LINKED ISSUE #1383: API clientStatus bug
**came over BM. I haven't tested it, just crosspasting**

Hmm well for me it's been working fine until just recently.

------------------------------------------------------
in --curses it never worked even be4

------------------------------------------------------
I've been running a headless BM instance for a while and I use the api.clientStatus call to monitor it.  I noticed that the networkStatus field says "notConnected" when it's connected and receiving incoming connections.
I don't know if this was today's pull or the one before that introduced it.

g1itch:
Fixed logic and reduced operations count.


g1itch:
Surprisingly the test works well.

PR #1388, opened 2018-11-05
Fix: Shift handling when deleting from Sent

PeterSurda:
Due to the way Qt.KeyboardModifiers works, treating it as an integer results in
unexpected behaviour. As a result, When deleting from Sent, it always deleted
rather than moved an item to trash, when selecting "Move to Trash" from the
context menu. This fixes it.

### DIFF
--- a/src/bitmessageqt/__init__.py
+++ b/src/bitmessageqt/__init__.py
@@ -3258,7 +3258,7 @@ class MyForm(settingsmixin.SMainWindow):
         if not tableWidget:
             return
         folder = self.getCurrentFolder()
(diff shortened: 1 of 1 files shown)


g1itch:
Yeah , looks like it. But I still not seeing deleted sent messages in trash folder, even after such change. Maybe because the sent messages have no "received" field?

g1itch:
Or because the `trash` folder selects its items from `inbox` table, not `sent`.

PeterSurda:
Yea sent trash has no UI at the moment, you have to access the database e.g. through command line.

PR #1391, opened 2018-11-11
Fix issue #1298

che2116:


PeterSurda:
Sadly, it doesn't work. It looks like the `closeEvent` is suppressed while the main qt thread is stuck in `self.quit`. It probably would have to be redone in a way that `self.quit` executes in a separate thread.

PeterSurda:
From the discussion in the bitmessage chan:

Well, not really. The signal is still blocked even if you call quit indirectly through trigger. It probably has to be redone in an actual thread, self.quit needs to be the thread's run function that is passed the myapp parameter. Upon first time the exit is triggered, the thread would start, and upon subsequent ones it would process the signal and adjust accordingly.

One approach that is slightly easier and may work as well is to create a custom event and use pushEvent to put it into the event queue.

Peter Surda
Bitmessage core developer

> It doesn't work out of the box, but I think the approach is still
> salvageable.
> 
> Peter Surda
> Bitmessage core developer
> 
> > This looks like a good idea, let me try that.
> > 
> > Peter Surda
> > Bitmessage core developer
> > 
> > > for one issue anyway
> > > 
> > > ```
>>> diff --git a/src/bitmessageqt/__init__.py
>>> b/src/bitmessageqt/__init__.py
>>> index be26a0d..7d62368 100644
>>> --- a/src/bitmessageqt/__init__.py
>>> +++ b/src/bitmessageqt/__init__.py
>>> @@ -2895,7 +2895,7 @@ class MyForm(settingsmixin.SMainWindow):
>>>          event.ignore()
>>>          if not trayonclose:
>>>              # quit the application
>>> -            self.quit()
>>> +            self.ui.actionExit.trigger()
>>> 
>>>      def on_action_InboxMessageForceHtml(self):
>>>          msgid = self.getCurrentMessageId()
>>> 
```

PeterSurda:
I tried a lot of things that were recommended in the chan, as well as some based on my own research. I tried two different asynchronous trigger approaches. It doesn't appear to change anything, it really looks like the shutdown procedure needs to run in a separate thread, i.e. some refactoring necessary.

g1itch:
20a42b1

PR #1396, opened 2018-11-26
Changes based on style and lint checks. (final_code_quality_14)

coffeedogs:
Changes based on style and lint checks. (final_code_quality_14)

### DIFF
--- a/src/bitmessageqt/utils.py
+++ b/src/bitmessageqt/utils.py
@@ -1,14 +1,25 @@
-from PyQt4 import QtGui
+"""
+src/bitmessageqt/utils.py
+=========================
+
+"""
+
 import hashlib
 import os
+
+from PyQt4 import QtGui
+
+import state
 from addresses import addBMIfNotPresent
 from bmconfigparser import BMConfigParser
-import state
 
 str_broadcast_subscribers = '[Broadcast subscribers]'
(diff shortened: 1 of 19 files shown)


src/bitmessageqt/utils.py
```diff
@@ -28,60 +39,66 @@ def identiconize(address):
...
     """
-        loads a supported image for the given address' hash form 'avatars' folder
-        falls back to default avatar if 'default.*' file exists
-        falls back to identiconize(address)
+    loads a supported image for the given address' hash form 'avatars' folder
+    falls back to default avatar if 'default.*' file exists
+    falls back to identiconize(address)
     """
+    # pylint: disable=global-statement
+    global str_broadcast_subscribers
```
PeterSurda: Are you sure you need global here at all since you fixed the assignment below?

src/messagetypes/__init__.py
```diff
@@ -7,7 +7,7 @@
 import paths
 
 class MsgBase(object):
-    def encode(self):
+    def encode(self, data=None):
         self.data = {"": lower(type(self).__name__)}
```
PeterSurda: what you could do is to make this a default (`if data is None:`), otherwise assign `data` to `self.data`

PR #1397, opened 2018-11-26
Changes based on style and lint checks. (final_code_quality_15)

coffeedogs:
Changes based on style and lint checks. (final_code_quality_15)

### DIFF
--- a/src/plugins/menu_qrcode.py
+++ b/src/plugins/menu_qrcode.py
@@ -1,18 +1,25 @@
 # -*- coding: utf-8 -*-
 """
+src/plugins/menu_qrcode.py
+==========================
+
 A menu plugin showing QR-Code for bitmessage address in modal dialog.
 """
-
-from PyQt4 import QtGui, QtCore
-import qrcode
+# pylint: disable=import-error
 
 from pybitmessage.tr import _translate
+from PyQt4 import QtCore, QtGui
+
+import qrcode
 
 
 # http://stackoverflow.com/questions/20452486
 class Image(qrcode.image.base.BaseImage):
     """Image output class for qrcode using QPainter"""
+    # pylint: disable=abstract-method
+
     def __init__(self, border, width, box_size):
+        super(Image, self).__init__()
         self.border = border
         self.width = width
         self.box_size = box_size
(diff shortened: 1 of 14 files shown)


src/shutdown.py
```diff
@@ -1,22 +1,25 @@
...
+src/shutdown.py
+===============
+"""
+
 import os
 import Queue
 import threading
 import time
 
+import shared
```
PeterSurda: reordering `shared` can cause problem, needs to be tested

src/protocol.py
```diff
@@ -1,34 +1,32 @@
...
 import hashlib
 import os
 import random
 import socket
 import ssl
-from struct import pack, unpack, Struct
 import sys
 import time
 import traceback
+from binascii import hexlify
```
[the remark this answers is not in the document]
g1itch: He sorts imports like @omkar1117 wrote: `import ...` first, then `from ... import ...`.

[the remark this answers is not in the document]
PeterSurda: I see.

src/shutdown.py
PeterSurda: Tested, works ok.

src/plugins/notification_notify2.py
```diff
@@ -1,15 +1,25 @@
...
+===================================
+"""
+# pylint: disable=import-error
 
 import gi
-gi.require_version('Notify', '0.7')
 from gi.repository import Notify
 
+gi.require_version('Notify', '0.7')
+
```
g1itch: This not works in such way. `gi.require_version()` should be called before import.

src/plugins/menu_qrcode.py
```diff
@@ -1,18 +1,25 @@
...
 """
-
-from PyQt4 import QtGui, QtCore
-import qrcode
+# pylint: disable=import-error
 
 from pybitmessage.tr import _translate
+from PyQt4 import QtCore, QtGui
+
+import qrcode
```
g1itch: Wrong order: PyQt4 and qrcode are [third party imports](https://www.python.org/dev/peps/pep-0008/#imports).

PR #1398, opened 2018-11-26
Changes based on style and lint checks. (final_code_quality_16)

coffeedogs:
Changes based on style and lint checks. (final_code_quality_16)

### DIFF
--- a/src/bitmessageqt/uisignaler.py
+++ b/src/bitmessageqt/uisignaler.py
@@ -1,11 +1,18 @@
+"""
+src/bitmessageqt/uisignaler.py
+==============================
+"""
 
-from PyQt4.QtCore import QThread, SIGNAL
 import sys
+
+from PyQt4.QtCore import SIGNAL, QThread
 
 import queues
 
 
 class UISignaler(QThread):
+    """A QT thread for managing UI status updates"""
+    # pylint: disable=too-many-locals,too-many-statements,too-many-branches
     _instance = None
 
     def __init__(self, parent=None):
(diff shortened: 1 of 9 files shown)


src/bitmessageqt/uisignaler.py
```diff
@@ -1,29 +1,39 @@
...
+    """A QT thread for managing UI status updates"""
+    # pylint: disable=too-many-locals,too-many-statements,too-many-branches
     _instance = None
 
     def __init__(self, parent=None):
         QThread.__init__(self, parent)
 
     @classmethod
     def get(cls):
+        """Ensure the same instance is always used"""
```
PeterSurda: i.e. singleton decorator

src/bitmessageqt/uisignaler.py
```diff
@@ -1,29 +1,39 @@
...
-from PyQt4.QtCore import QThread, SIGNAL
 import sys
 
+from PyQt4.QtCore import SIGNAL, QThread
+
 import queues
 
 
 class UISignaler(QThread):
+    """A QT thread for managing UI status updates"""
```
PeterSurda: This isn't really accurate. This thread processes incoming "signals" (technically queue entries) that other threads send to the UI.

src/bitmessageqt/uisignaler.py
```diff
@@ -1,29 +1,39 @@
...
 
     @classmethod
     def get(cls):
+        """Ensure the same instance is always used"""
         if not cls._instance:
             cls._instance = UISignaler()
         return cls._instance
 
     def run(self):
+        """Loop to update the UI statusbar. Never returns."""
```
PeterSurda: Loop to handle incoming signals.

src/helper_threading.py
```diff
@@ -23,20 +30,26 @@ def _thread_name_hack(self):
 
 
 class StoppableThread(object):
+    """A base class for threads giving them a custom stopping mechanism"""
     def initStop(self):
+        """Start the process of stopping the thread"""
```
PeterSurda: Perhaps I should have used a different name. This is just an initialiser, it does not stop anything. I avoided using `__init__` as this often used as a mixin, or maybe because of execution order, I don't remember the details. Perhaps I was overly cautious or simple there was something I haven't figured out how to do. If it can be done with a normal `__init__`, that would be preferable.

src/helper_threading.py
```diff
@@ -23,20 +30,26 @@ def _thread_name_hack(self):
...
         self.stop = threading.Event()
         self._stopped = False
 
     def stopThread(self):
+        """Complete the process of stopping a thread"""
```
PeterSurda: Perhaps "Stop a thread asynchronously".

src/network/receivequeuethread.py
```diff
@@ -1,25 +1,25 @@
...
-from network.bmproto import BMProto
 from network.advanceddispatcher import UnknownStateError
+from network.connectionpool import BMConnectionPool
 from queues import receiveDataQueue
-import protocol
-import state
+
 
 class ReceiveQueueThread(threading.Thread, StoppableThread):
+    """A thread for receiving items"""
```
PeterSurda: This thread processes data received from the network.

PR #1405, opened 2018-12-09
Fix spelling errors for the connection warning dialog.

GONQ:
Quick fix: changed 'connectins' to correct spelling of 'connections' in init and lang files. This mispelled word shows up in the connection warning dialog when going online/offline.


g1itch:
You don't need to change those ts files. It's done by pylupdate.

g1itch:
> Does the ts generator pull from **init**.py?

Yes, see [bitmessage.pro](../blob/v0.6/src/translations/bitmessage.pro#L19)

PeterSurda:
"This user has not uploaded their public key yet."

You need to go to your account settings on github, go to the keys section and upload your PGP pubkey.

PR #1406, opened 2018-12-10
Moved missingObjects dict from state to network.objectracker

g1itch:
Hello

I noticed `missingObjects` dict is used only inside of `network` package. So maybe it's reasonable to define it in `network.objectracker`?


PeterSurda:
That makes sense, I'll try to review it today but it looks straightforward.

PeterSurda:
Code looks ok, will run some tests.

Issue #1408, opened 2018-12-13
Duplicate connections to some onion peers


g1itch:
The description is based on messages from `[chan] bitmessage` but I've seen it myself a couple of weeks ago (I thought it's related to my #1394 - wrongly).

![image](https://user-images.githubusercontent.com/4012700/49939372-56add980-fee5-11e8-8f6c-de2f83123ebd.png)

Changes proposed today:
```patch
diff --git a/src/network/connectionpool.py b/src/network/connectionpool.py
index e599cdf..c5ba701 100644
--- a/src/network/connectionpool.py
+++ b/src/network/connectionpool.py
@@ -93,7 +93,7 @@ class BMConnectionPool(object):
                     del self.inboundConnections[connection.destination.host]
                 except KeyError:
                     pass
-        connection.close()
+        connection.handle_close()
 
     def getListeningIP(self):
         if BMConfigParser().safeGet("bitmessagesettings", "onionhostname").endswith(".onion"):

```

stman:
Are you saying this node was not an attack on bitmessage to sandbox users ?

Envoyé de mon iPhone

g1itch:
My own screenshot with the same peer:
![image](https://user-images.githubusercontent.com/4012700/49946100-ed36c680-fef6-11e8-9bd6-c020fbf311c1.png)

And with other peer from 11.28:
![image](https://user-images.githubusercontent.com/4012700/49946162-12c3d000-fef7-11e8-8b16-2fd7d880cfbb.png)

stman:
Have you implemented a mecanism to select the best noted peers to connect preferably to them first ?

Because it is the case, I think this could be a sandboxing attack. 



Envoyé de mon iPhone

PeterSurda:
@stman The rating adjust dynamically and influences the likelihood the peer will be retried in the future. The client isn't supposed to connect to the same peer multiple times, it's a bug. It also could be a UI bug, erroneously forgetting to remove an entry from the list upon disconnect. Looking at @g1itch 's patch, I think it's probably the UI. I haven't had time to investigate it yet but I also saw something similar happening on my local client.

stman:
Yep.

I think it must be seriously investigated, and this mecanism to preferably connect to the best peers, even randomly, is at the root, according to me, to a way for the trick exploited to sandbox me and several users, to sabotage silently bitmessage’ peer ring.

Envoyé de mon iPhone

stman:
[...]

PR #1413, opened 2018-12-18
Quick fix for #1408

g1itch:
The patch from `[chan] bitmessage`

```python
class BMProto(AdvancedDispatcher, ObjectTracker):
...
    def handle_close(self):
        self.set_state("close")
        if not (self.accepting or self.connecting or self.connected):
            # already disconnected
            return
...
```

### DIFF
--- a/src/network/connectionpool.py
+++ b/src/network/connectionpool.py
@@ -93,7 +93,7 @@ class BMConnectionPool(object):
                     del self.inboundConnections[connection.destination.host]
                 except KeyError:
                     pass
-        connection.close()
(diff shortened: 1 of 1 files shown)


PeterSurda:
check process got stuck or something, restarting

g1itch:
> check process got stuck or something, restarting

Thanks

stman:
Can I clone a whole new version of the source from github then ?

PeterSurda:
yes just pull `v0.6` branch (which should be the default if you're cloning it fresh)

PR #1415, opened 2018-12-18
Put uploads into a separate thread

### LINKED ISSUE #1414: Memory leak when remote peer requests too many objects
I noticed huge memory usage on some of my nodes. I managed to add memory usage tracking to PyBM, and it looks like some clients request too many objects (possibly even duplicate), and PyBitmessage then fills the socket write buffers with the object data without any length constraints. This is due to design of uploading. In most cases not checking the write buffer size isn't a problem, I don't think that a remote node has other methods of causing too much data to be send other than requesting a lot of objects.

On one of my machines I got 11GB RAM consumed by this, even though there weren't that many connections.

There are several approaches about how to address this, which can be combined.

- limit number of processed upload requests (very easy but reduces transfer performance)
- limit write buffer size (very easy but may break smooth network transfers, it may actually affect other types of commands)
- suppress duplicate uploading for some time (easy but incomplete)
- upload asynchronously, in a separate thread, similarly to how the download thread works (complicated but a proper fix)

PeterSurda:
- instead of being processed in the ReceiveQueue thread, uploads are now done
  in a dedicated thread. Only the parsing is done in ReceiveQueue thread.
- the UploadThread is modelled based on the DownloadThred, but simpler.
- it checks for intersection attack, eliminates duplicates and restricts the
  write buffer size to 2MB (may still grow slightly higher if too many big
  objects are requested, but the absolute limit appears to be about 4.5MB in the
  worst case scenario).
- the restriction of the write buffer may cause some upload throttling (to
  about 2MB per second per connection), but can be optimised later
- fixes #1414
- I haven't actually tested it yet, I'll test it as the PR is being checked and reviewed

### DIFF
--- a/src/network/bmproto.py
+++ b/src/network/bmproto.py
@@ -18,6 +18,8 @@ from network.node import Node
 from network.objectracker import ObjectTracker
 from network.proxy import ProxyError
 from objectracker import missingObjects
+from randomtrackingdict import RandomTrackingDict
+
 
 import addresses
 from queues import objectProcessorQueue, portCheckerQueue, invQueue, addrQueue
@@ -57,6 +59,7 @@ class BMProto(AdvancedDispatcher, ObjectTracker):
         self.isOutbound = False
         # packet/connection from a local IP
         self.local = False
+        self.pendingUpload = RandomTrackingDict()
 
     def bm_proto_reset(self):
         self.magic = None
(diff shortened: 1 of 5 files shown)


PeterSurda:
I changed the checks to ignore import position errors in bitmessagemain.py, as I don't know how else to work around this. I can revert it once this is merged.

PeterSurda:
Testing now, typo causing exception.

src/network/bmproto.py
```diff
@@ -277,25 +279,11 @@ def bm_command_error(self):
     def bm_command_getdata(self):
         items = self.decode_payload_content("l32s")
         # skip?
-        if time.time() < self.skipUntil:
+        now = time.time()
+        if now < self.skipUntil:
             return True
-        #TODO make this more asynchronous
-        helper_random.randomshuffle(items)
         for i in map(str, items):
```
g1itch: Why not
```python
for i in items:
    self.pendingUpload[str(i)] = now
```
?

PeterSurda: Leftover from old code, you're right, it can be rewritten for maybe a bit of performance increase.

PR #1419, opened 2019-01-01
Fix: Mac build

jramos:
* Use `/usr/local/opt/openssl` symlink for `bitmsghash` flags:
The current version from homebrew is `1.0.2q`. Using the symlink eliminates the need for updating the version in `CCFLAGS` and `LDFLAGS`.

* Install `pyqt@4`:
`pyqt` 5.x is installed by default, so the `@4` is needed to install a compatible version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ src/.settings/
 src/**/.dll
 src/**/*.o
 src/**/*.so
+src/**/a.out
 build/lib.*
 build/temp.*
 dist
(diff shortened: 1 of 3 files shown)


PeterSurda:
Looks like pyqt@4 isn't in the repo anymore, you have to build it from source. But the patch looks ok.

jramos:
Added `cartr/qt4` to `pyqt@4` install instructions.

PeterSurda:
That one looks like it's using python3.

PR #1422, opened 2019-01-02
Logic error in download reset

PeterSurda:
- fix requesting the same object over and over again, now it continues to
  iterate through the other objects and only resets the queue after a timeout
  after the last received object
- tested on Debian

### DIFF
--- a/src/randomtrackingdict.py
+++ b/src/randomtrackingdict.py
@@ -108,8 +108,8 @@ class RandomTrackingDict(object):
         # pylint: disable=redefined-outer-name
         with self.lock:
             # reset if we've requested all
-            # or if last object received too long time ago
-            if self.pendingLen == self.len or self.lastObject + self.pendingTimeout > time():
+            # and if last object received too long time ago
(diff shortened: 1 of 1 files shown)


src/randomtrackingdict.py
```diff
@@ -108,8 +108,8 @@ def randomKeys(self, count=1):
         # pylint: disable=redefined-outer-name
         with self.lock:
             # reset if we've requested all
-            # or if last object received too long time ago
-            if self.pendingLen == self.len or self.lastObject + self.pendingTimeout > time():
+            # and if last object received too long time ago
+            if self.pendingLen == self.len and self.lastObject + self.pendingTimeout > time():
```
g1itch: I don't understand it, `self.lastObject + self.pendingTimeout > time()` means that timeout isn't over? The comment says the opposite.

PeterSurda: I think you're correct. I'll change it and retest.

PeterSurda: Retest looks ok.

PR #1424, opened 2019-01-05
Hyperlink

benhc123:
Hyperlink vs raw link.


PeterSurda:
Using "here" as a hyperlink label is bad.

benhc123:
Ah ok. Fair enough.

PR #1425, opened 2019-01-11
Better nodes filtering in TCPConnection.sendAddr()

g1itch:
Hello

I changed `TCPConnection.sendAddr()` to send only nodes with positive or neutral rating.

### DIFF
--- a/src/network/tcp.py
+++ b/src/network/tcp.py
@@ -130,36 +130,30 @@ class TCPConnection(BMProto, TLSDispatcher):  # pylint: disable=too-many-instanc
         # We are going to share a maximum number of 1000 addrs (per overlapping
         # stream) with our peer. 500 from overlapping streams, 250 from the
         # left child stream, and 250 from the right child stream.
-        maxAddrCount = BMConfigParser().safeGetInt("bitmessagesettings", "maxaddrperstreamsend", 500)
+        maxAddrCount = BMConfigParser().safeGetInt(
+            "bitmessagesettings", "maxaddrperstreamsend", 500)
 
-        # init
         templist = []
         addrs = {}
         for stream in self.streams:
             with knownnodes.knownNodesLock:
-                if knownnodes.knownNodes[stream]:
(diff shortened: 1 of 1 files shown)


src/network/tcp.py
```diff
@@ -130,36 +130,30 @@ def sendAddr(self):
...
         for stream in self.streams:
             with knownnodes.knownNodesLock:
-                if knownnodes.knownNodes[stream]:
-                    filtered = {k: v for k, v in knownnodes.knownNodes[stream].items()
-                                if v["lastseen"] > (int(time.time()) - shared.maximumAgeOfNodesThatIAdvertiseToOthers)}
-                    elemCount = len(filtered)
-                    if elemCount > maxAddrCount:
-                        elemCount = maxAddrCount
+                for n, s in enumerate((stream, stream * 2, stream * 2 + 1)):
+                    nodes = knownnodes.knownNodes[s]
```
PeterSurda: Shouldn't this have an exception handler (KeyError?)

src/network/tcp.py
```diff
@@ -130,36 +130,30 @@ def sendAddr(self):
...
         addrs = {}
         for stream in self.streams:
             with knownnodes.knownNodesLock:
-                if knownnodes.knownNodes[stream]:
-                    filtered = {k: v for k, v in knownnodes.knownNodes[stream].items()
-                                if v["lastseen"] > (int(time.time()) - shared.maximumAgeOfNodesThatIAdvertiseToOthers)}
-                    elemCount = len(filtered)
-                    if elemCount > maxAddrCount:
-                        elemCount = maxAddrCount
+                for n, s in enumerate((stream, stream * 2, stream * 2 + 1)):
```
PeterSurda: This is interesting, but seems to work.

src/network/tcp.py
g1itch: Reasonable question. Replaced by `.get(s)`

PR #1429, opened 2019-01-30
More obsolete code removed

g1itch:
Hello!

I removed more unused code.

`protocol.networkType` is also unused.

Also I suggest to move `sslProtocolVersion` and `sslProtocolCiphers` from `protocol` to `network.tls`; `assembleErrorMessage` and `haveSSL` to `network.bmproto`

### DIFF
--- a/src/network/bmproto.py
+++ b/src/network/bmproto.py
@@ -431,7 +431,6 @@ class BMProto(AdvancedDispatcher, ObjectTracker):
         if not self.peerValidityChecks():
             # TODO ABORT
             return True
-        #shared.connectedHostsList[self.destination] = self.streams[0]
         self.append_write_buf(protocol.CreatePacket('verack'))
         self.verackSent = True
         if not self.isOutbound:
--- a/src/network/bmproto.py
+++ b/src/network/bmproto.py
@@ -23,9 +23,9 @@ from randomtrackingdict import RandomTrackingDict
 
 import addresses
 from queues import objectProcessorQueue, portCheckerQueue, invQueue, addrQueue
-import shared
 import state
 import protocol
+
 
 class BMProtoError(ProxyError):
     errorCodes = ("Protocol error")
@@ -52,6 +52,7 @@ class BMProto(AdvancedDispatcher, ObjectTracker):
     addressAlive = 10800
     # maximum time offset
     maxTimeOffset = 3600
+    timeOffsetWrongCount = 0
 
     def __init__(self, address=None, sock=None):
         AdvancedDispatcher.__init__(self, sock)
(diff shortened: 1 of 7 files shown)


src/protocol.py
```diff
@@ -316,125 +311,6 @@ def assembleErrorMessage(fatal=0, banTime=0, inventoryVector='', errorText=''):
     return CreatePacket('error', payload)
 
 
-# Packet decoding
-
-
-def decryptAndCheckPubkeyPayload(data, address):
```
PeterSurda: this should be the other way around, remove from shared (after updating class_objectProcessor.py and class_singleWorker.py).

g1itch: OK, I removed it from `protocol` because that copy is broken:
https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/protocol.py#L418

PeterSurda: I see

g1itch:
Some of that variable are also seems obsolete, e.g. `alreadyAttemptedConnectionsList*`.
Should I move all functions from `shared` in this PR?

g1itch:
Also noticed:
`successfullyDecryptMessageTimings` is only used in `class_objectProcessor`,
`needToWriteKnownNodesToDisk` - in `knownnodes`,
`timeOffsetWrongCount` - in `network.bmproto`

src/shared.py
```diff
@@ -6,10 +6,8 @@
 import stat
 import time
```
PeterSurda: Please remove this line.

g1itch:
I applied all changes except for `protocol.networkType`, `protocol.assembleErrorMessage` and `protocol.haveSSL`. Please check if you are OK with that. Also maybe should I squash commits?

src/network/bmproto.py
```diff
@@ -460,17 +461,17 @@ def peerValidityChecks(self):
                 errorText="Your time is too far in the future compared to mine. Closing connection."))
             logger.info("%s's time is too far in the future (%s seconds). Closing connection to it.",
                 self.destination, self.timeOffset)
-            shared.timeOffsetWrongCount += 1
+            self.timeOffsetWrongCount += 1
```
PeterSurda: This breaks things because it's now per connection, whereas it should be global. I suggest you use a class variable, i.e.
`BMProto.timeOffsetWrongCount` instead of `self.timeOffsetWRongCount`. Based on googling (although I haven't tested it) it should work. Besides, I'm not sure I fully tested this variable after moving to asyncore.

g1itch: I didn't found where this variable used at all.

g1itch: Yes setting `BMProto.timeOffsetWrongCount` works as global variable.
```
In [31]: class TestVar(object):
    ...:     var = 0
    ...:     def inc(self):
    ...:         TestVar.var += 1
    ...:         

In [32]: t1 = TestVar()

In [33]: t2 = TestVar()

In [34]: t1.var
Out[34]: 0

In [35]: t1.inc()

In [36]: t1.var
Out[36]: 1

In [37]: t2.var
Out[37]: 1
```

PeterSurda: So I probably forgot finishing implementing this. Just leave it as a class variable then and I'll open an issue about finishing this.

PR #1430, opened 2019-01-30
Dockerfile

g1itch:
Dockerfile for bitmessage daemon.


PeterSurda:
Very good, my only objection is that it's setup to run as root.

PR #1439, opened 2019-02-28
Fix bug in bitmessageqt.propagateUnreadCount()

g1itch:
Hello!

I fixed unread count calculation when there are unread messages in trash and handled `folder` and `widget` args. I still doubt, should I handle also `address` and `address_type` or remove them completely?

The bug:
bm_unread_trash
bm_unread_trash_selected


PeterSurda:
I have to look at this in more detail, I remember I spent a lot of time profiling this so don't want to break things.

g1itch:
If this changes are still OK, I'm going to squash inner commits and merge.

PR #1454, opened 2019-05-03
Removed pluralization artifact introduced in a69daa1

g1itch:
The string "Display the %n recent broadcast(s) from this address." in `bitmessageqt.address_dialogs` had `%1` substitution mark before pluralization in a69daa1. Somehow it remained in the translations and thus shown as is in that locales.


PeterSurda:
These things should be done on transifex. The webhook may be broken though so I may have to create a PR manually.

g1itch:
I'm not sure if there are active translators for that languages.

PeterSurda:
Traditionally I was translating both, maybe I should hire someone for that now.

PR #1455, opened 2019-05-06
 Minimal implementation of onionpeer object

g1itch:
Hello!

In this changes each node having onion hidden service sends an object with objectType 4 and TTL 1 week.

PR #1462, opened 2019-05-22
Do not insert encoded onion v3 address into version message

g1itch:
Hi!

This is a followup for #1455. As I can see the `remoteHost` in version command isn't used by receiving party so it may be also replaced with `127.0.0.1` for onion v3 service addresses. Without this change the fields `user_agent` and `stream_numbers` got mangled and interpreted as '' and 0.

### DIFF
--- a/src/protocol.py
+++ b/src/protocol.py
@@ -262,7 +262,9 @@ def assembleVersionMessage(remoteHost, remotePort, participatingStreams, server=
         payload += encodeHost('127.0.0.1')
         payload += pack('>H', 8444)
     else:
-        payload += encodeHost(remoteHost)
+        # use first 16 bytes if host data is longer
+        # for example in case of onion v3 service
+        payload += encodeHost(remoteHost)[:16]
         payload += pack('>H', remotePort)  # remote IPv6 and port
 
     # bitflags of the services I offer.


src/protocol.py
```diff
@@ -262,7 +262,8 @@ def assembleVersionMessage(remoteHost, remotePort, participatingStreams, server=
         payload += encodeHost('127.0.0.1')
         payload += pack('>H', 8444)
     else:
-        payload += encodeHost(remoteHost)
+        # 127.0.0.1 for onion v3 services
+        payload += encodeHost(remoteHost if len(remoteHost) <= 22 else '127.0.0.1')
```
PeterSurda: remoteHost is helpful in the logs so that you can see your external IP address. Maybe it's not currently used other than in the logs but it could in the future.

g1itch: So what to do? Place there first 16 bytes of onion v3 pubkey?

PeterSurda: I think that would be better than 127.0.0.1

src/protocol.py
```diff
@@ -262,7 +262,9 @@ def assembleVersionMessage(remoteHost, remotePort, participatingStreams, server=
         payload += encodeHost('127.0.0.1')
         payload += pack('>H', 8444)
     else:
-        payload += encodeHost(remoteHost)
+        # use first 16 bytes if host data is longer
+        # for example in case of onion v3 service
+        payload += encodeHost(remoteHost)[:17]
```
PeterSurda: I think this is an off-by-one bug, it produces a string that is up to 17 bytes long (position 0 to 16). Or maybe I'm missing something.

g1itch: oops! There probably should be `[:16]`.

PR #1466, opened 2019-06-06
Fixing my bug in 453e045

g1itch:
Hello!

I commited a bug in 453e045 when changed `protocol.OBJECT_ONIONPEER` from 4 to 0x746f72:

https://github.com/Bitmessage/PyBitmessage/commit/453e045ae5c846b00be10fb397ec33907782cb0d#diff-433d5e3077dae305aaac72fe4ecedad1R485


g1itch:
Maybe I'm missing something else.

PR #1471, opened 2019-06-21
Finalizing qrcode plugin

### LINKED ISSUE #819: idea: QR Code export of identities

g1itch:
Hello!

This changeset closes #819.

### DIFF
--- a/src/bitmessageqt/__init__.py
+++ b/src/bitmessageqt/__init__.py
@@ -817,6 +817,17 @@ class MyForm(settingsmixin.SMainWindow):
         self.initSettings()
         self.resetNamecoinConnection()
 
+        self._contact_selected = None
+
+    def getContactSelected(self):
+        """Returns last selected contact once"""
+        try:
+            return self._contact_selected
+        except AttributeError:
+            pass
+        finally:
+            self._contact_selected = None
+
     def updateTTL(self, sliderPosition):
         TTL = int(sliderPosition ** 3.199 + 3600)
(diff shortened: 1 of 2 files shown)


src/plugins/menu_qrcode.py
```diff
@@ -77,7 +81,20 @@ def on_action_ShowQR():
             dialog = form.qrcode_dialog
         except AttributeError:
             form.qrcode_dialog = dialog = QRCodeDialog(form)
-        dialog.render('bitmessage:' + str(form.getCurrentAccount()))
+        account = form._contact_selected
```
PeterSurda: Maybe there can be a getContactSelected method to prevent this?

src/bitmessageqt/__init__.py
```diff
@@ -4004,6 +4015,11 @@ def on_context_menuInbox(self, point):
                     _translate("MainWindow", "Copy address to clipboard"),
                     self.on_action_ClipboardMessagelist)
                 self.popMenuInbox.addAction(self.actionClipboardMessagelist)
+                self._contact_selected = accountClass(
+                    tableWidget.item(currentRow, 1).data(QtCore.Qt.UserRole))
```
PeterSurda: You can probably suppress the warning, this looks like pylint doesn't understand we check that it's not boolean before.

PR #1474, opened 2019-06-27
Basic code clean up

omkar1117:
Basic code cleanup


src/depends.py
```diff
@@ -4,6 +4,10 @@
 """
 
 import sys
+import logging
```
g1itch: This module intentionally violates pep8 because of it's purpose: it should work in any (even unsupported) python version, so it imports minimum needed modules for each check. There will be `ImportError` in python 2.3 because it has no `logging` as the following comment says.

PR #1475, opened 2019-06-28
fix windows header files api issue

surbhicis:
fix windows header files api issue.


PeterSurda:
This patch is needed for crosscompiling bitmsghash on linux for windows target, because linux filesystem is case sensitive.

PR #1480, opened 2019-07-08
More style and formatting changes

g1itch:
Hello!

Here is a bunch of style and formatting changes. I'm going to incorporate relevant changes from @coffeedogs PRs in this branch.

### DIFF
--- a/src/network/bmproto.py
+++ b/src/network/bmproto.py
@@ -27,18 +27,22 @@ from randomtrackingdict import RandomTrackingDict
 
 
 class BMProtoError(ProxyError):
+    """A Bitmessage Protocol Base Error"""
     errorCodes = ("Protocol error")
 
 
 class BMProtoInsufficientDataError(BMProtoError):
+    """A Bitmessage Protocol Insufficient Data Error"""
     errorCodes = ("Insufficient data")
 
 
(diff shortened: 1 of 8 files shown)


src/network/bmproto.py
```diff
@@ -201,51 +212,59 @@ def decode_simple(self, char="v"):
         # retval (array)
         parserStack = [[1, 1, False, pattern, 0, []]]
 
-        #try:
+        # try:
```
PeterSurda: This section can be removed, it's from when I was debugging for the "maximum stack exceeded" exception, in the end it was in the garbage collector so it's completely unrelated)

src/storage/filesystem.py
```diff
@@ -1,173 +1,215 @@
...
         return self._inventory.keys()
 
     def object_list(self):
-        return [unhexlify(x) for x in listdir(path.join(self.baseDir, FilesystemInventory.objectDir))]
+        """Return inventory vectors (hashes) from a directory"""
+        return [
+            unhexlify(x) for x in os.listdir(
+                os.path.join(self.baseDir, self.objectDir))]
 
     def getData(self, hashId):
```
g1itch: Am I understand right that `hashId` has the same meaning as `hash` argument above, which I renamed as `invhash`, for example in line 83?

PR #1483, opened 2019-07-11
Dandelion fixes

PeterSurda:
- expiration wasn't handled correctly
- objects with no child stems never expired. While this is better for
  anonymity, it can cause objects getting stuck
- upon expiration the nodes weren't marked as not having the object, causing it
  to not be advertised
- I haven't tested it but at least I don't think can make things worse
- also minor PEP8 fixes

### DIFF
--- a/src/network/dandelion.py
+++ b/src/network/dandelion.py
@@ -122,12 +122,14 @@ class Dandelion():
     def expire(self):
         with self.lock:
             deadline = time()
-            toDelete = [[v.stream, k, v.child]
-                        for k, v in self.hashMap.iteritems()
-                        if v.timeout < deadline]
+            toDelete = [
+                [v.stream, k, v.child] for k, v in self.hashMap.iteritems()
+                if v.timeout < deadline
+            ]
+
             for row in toDelete:
                 self.removeHash(row[1], 'expiration')
-                invQueue.put((row[0], row[1], row[2]))
+                invQueue.put(row)
         return toDelete
 
     def reRandomiseStems(self):
--- a/src/network/dandelion.py
+++ b/src/network/dandelion.py
@@ -122,11 +122,13 @@ class Dandelion():
     def expire(self):
         with self.lock:
             deadline = time()
-            # only expire those that have a child node, i.e. those without a child not will stick around
-            toDelete = [[v.stream, k, v.child] for k, v in self.hashMap.iteritems() if v.timeout < deadline and v.child]
+            toDelete = [[v.stream, k, v.child]
+                        for k, v in self.hashMap.iteritems()
+                        if v.timeout < deadline]
             for row in toDelete:
                 self.removeHash(row[1], 'expiration')
                 invQueue.put((row[0], row[1], row[2]))
+        return toDelete
 
     def reRandomiseStems(self):
         with self.lock:
--- a/src/network/invthread.py
+++ b/src/network/invthread.py
@@ -12,6 +12,27 @@ from queues import invQueue
 import protocol
 import state
 
+
+def handleExpiredDandelion(expired):
+    """For expired dandelion objects, mark all remotes as not having
+       the object"""
+    if not expired:
+        return
+    for i in \
+        BMConnectionPool().inboundConnections.values() + \
+            BMConnectionPool().outboundConnections.values():
+        if not i.fullyEstablished:
+            continue
+        for x in expired:
+            streamNumber, hashid, _ = x
+            try:
+                del i.objectsNewToMe[hashid]
+            except KeyError:
+                if streamNumber in i.streams:
+                    with i.objectsNewToThemLock:
+                        i.objectsNewToThem[hashid] = time()
+
+
 class InvThread(threading.Thread, StoppableThread):
     def __init__(self):
         threading.Thread.__init__(self, name="InvBroadcaster")
@@ -20,8 +41,9 @@ class InvThread(threading.Thread, StoppableThread):
 
     def handleLocallyGenerated(self, stream, hashId):
         Dandelion().addHash(hashId, stream=stream)
-        for connection in BMConnectionPool().inboundConnections.values() + \
-            BMConnectionPool().outboundConnections.values():
+        for connection in \
+                BMConnectionPool().inboundConnections.values() + \
+                BMConnectionPool().outboundConnections.values():
                 if state.dandelion and connection != Dandelion().objectChildStem(hashId):
                     continue
                 connection.objectsNewToThem[hashId] = time()
@@ -31,7 +53,7 @@ class InvThread(threading.Thread, StoppableThread):
             chunk = []
             while True:
                 # Dandelion fluff trigger by expiration
-                Dandelion().expire()
+                handleExpiredDandelion(Dandelion().expire())
                 try:
                     data = invQueue.get(False)
                     chunk.append((data[0], data[1]))


src/network/dandelion.py
```diff
@@ -122,11 +122,13 @@ def getNodeStem(self, node=None):
...
             deadline = time()
-            # only expire those that have a child node, i.e. those without a child not will stick around
-            toDelete = [[v.stream, k, v.child] for k, v in self.hashMap.iteritems() if v.timeout < deadline and v.child]
+            toDelete = [[v.stream, k, v.child]
+                        for k, v in self.hashMap.iteritems()
+                        if v.timeout < deadline]
             for row in toDelete:
                 self.removeHash(row[1], 'expiration')
                 invQueue.put((row[0], row[1], row[2]))
+        return toDelete
```
g1itch:
```python
toDelete = [
    v.stream, k, v.child for k, v in self.hashMap.iteritems()
    if v.timeout < deadline
]

for row in toDelete:
    self.removeHash(row[1], 'expiration')
    invQueue.put(row)
```

src/network/invthread.py
```diff
@@ -12,6 +12,27 @@
...
+        if not i.fullyEstablished:
+            continue
+        for x in expired:
+            streamNumber, hashid, _ = x
+            try:
+                del i.objectsNewToMe[hashid]
+            except KeyError:
+                if streamNumber in i.streams:
+                    with i.objectsNewToThemLock:
+                        i.objectsNewToThem[hashid] = time()
```
g1itch: Maybe this function should be a method of `InvThread`, like `handleLocallyGenerated` below?

Unfortunately I don't understand the meaning of `i.objectsNewToThem[hashid] = time()` ):

src/network/invthread.py
```diff
@@ -20,8 +41,9 @@ def __init__(self):
...
     def handleLocallyGenerated(self, stream, hashId):
         Dandelion().addHash(hashId, stream=stream)
-        for connection in BMConnectionPool().inboundConnections.values() + \
-            BMConnectionPool().outboundConnections.values():
+        for connection in \
+                BMConnectionPool().inboundConnections.values() + \
+                BMConnectionPool().outboundConnections.values():
                 if state.dandelion and connection != Dandelion().objectChildStem(hashId):
                     continue
                 connection.objectsNewToThem[hashId] = time()
```
g1itch: You used `with i.objectsNewToThemLock:` for similar expression above

src/network/dandelion.py
PeterSurda: Good point.

src/network/invthread.py
PeterSurda: If I make it into a method, pylint complains that "method could be a function".

The time in objectsNewToThem is for expiration, so that if the peer doesn't request the object the tracking doesn't occupy memory forever.

src/network/invthread.py
PeterSurda: It's not necessary to lock here, because this is a new object not seen anywhere on the network. I think that's how it works.

src/network/invthread.py
g1itch: Also this seems to duplicate `handleLocallyGenerated` if child is None:

```python
                try:
                    data = invQueue.get(False)
                    chunk.append((data[0], data[1]))
                    # locally generated
                    if len(data) == 2 or data[2] is None:
                        self.handleLocallyGenerated(data[0], data[1])
                except Queue.Empty:
                    break
```

src/network/dandelion.py
g1itch: Actually there is a mistake, should be `(v.stream, k, v.child)` ):

src/network/invthread.py
PeterSurda: No, handleLocallyGenerated advertises to a single child stem, expiration to everyone.

g1itch: > If I make it into a method, pylint complains that "method could be a function".
> 

So maybe there is no need for separate function in `invthread` module? Let's make it part of `Dandelion().expire()`

```diff
diff --git a/src/network/dandelion.py b/src/network/dandelion.py
index 60d84cdf..bb5cdee7 100644
--- a/src/network/dandelion.py
+++ b/src/network/dandelion.py
@@ -130,7 +130,22 @@ class Dandelion():
             for row in toDelete:
                 self.removeHash(row[1], 'expiration')
                 invQueue.put(row)
-        return toDelete
+
+        # For expired dandelion objects, mark all remotes as not having
+        # the object
+        for connection in (
+            network.connectionpool.BMConnectionPool().inboundConnections.values() +
+            network.connectionpool.BMConnectionPool().outboundConnections.values()
+        ):
+            if not connection.fullyEstablished:
+                continue
+            for streamNumber, hashid, _ in toDelete:
+                try:
+                    del connection.objectsNewToMe[hashid]
+                except KeyError:
+                    if streamNumber in connection.streams:
+                        with connection.objectsNewToThemLock:
+                            connection.objectsNewToThem[hashid] = time()
 
     def reRandomiseStems(self):
         with self.lock:
diff --git a/src/network/invthread.py b/src/network/invthread.py
index 6f6f1364..37b22cbf 100644
--- a/src/network/invthread.py
+++ b/src/network/invthread.py
@@ -13,26 +13,6 @@ import protocol
 import state
 
 
-def handleExpiredDandelion(expired):
-    """For expired dandelion objects, mark all remotes as not having
-       the object"""
-    if not expired:
-        return
-    for i in \
-        BMConnectionPool().inboundConnections.values() + \
-            BMConnectionPool().outboundConnections.values():
-        if not i.fullyEstablished:
-            continue
-        for x in expired:
-            streamNumber, hashid, _ = x
-            try:
-                del i.objectsNewToMe[hashid]
-            except KeyError:
-                if streamNumber in i.streams:
-                    with i.objectsNewToThemLock:
-                        i.objectsNewToThem[hashid] = time()
-
-
 class InvThread(threading.Thread, StoppableThread):
     def __init__(self):
         threading.Thread.__init__(self, name="InvBroadcaster")
@@ -53,7 +33,7 @@ class InvThread(threading.Thread, StoppableThread):
             chunk = []
             while True:
                 # Dandelion fluff trigger by expiration
-                handleExpiredDandelion(Dandelion().expire())
+                Dandelion().expire()
                 try:
                     data = invQueue.get(False)
                     chunk.append((data[0], data[1]))
```

g1itch: I'd also suggest to replace this annoying expression
```python
network.connectionpool.BMConnectionPool().inboundConnections.values() +
network.connectionpool.BMConnectionPool().outboundConnections.values()
```

with something like `BMConnectionPool().connections`. Unfortunately [the generator](../../../g1itch/PyBitmessage/commit/2dc3586) I wrote not works because connections got changed while iterating ):

PeterSurda: > So maybe there is no need for separate function in `invthread` module? Let's make it part of `Dandelion().expire()`

That's what I wanted to do originally but then for some reason decided not to. Now in retrospect, that would probably be better.
> 
> ```diff
> diff --git a/src/network/dandelion.py b/src/network/dandelion.py
> index 60d84cdf..bb5cdee7 100644
> --- a/src/network/dandelion.py
> +++ b/src/network/dandelion.py
> @@ -130,7 +130,22 @@ class Dandelion():
>              for row in toDelete:
>                  self.removeHash(row[1], 'expiration')
>                  invQueue.put(row)
> -        return toDelete
> +
> +        # For expired dandelion objects, mark all remotes as not having
> +        # the object
> +        for connection in (
> +            network.connectionpool.BMConnectionPool().inboundConnections.values() +
> +            network.connectionpool.BMConnectionPool().outboundConnections.values()
> +        ):
> +            if not connection.fullyEstablished:
> +                continue
> +            for streamNumber, hashid, _ in toDelete:
> +                try:
> +                    del connection.objectsNewToMe[hashid]
> +                except KeyError:
> +                    if streamNumber in connection.streams:
> +                        with connection.objectsNewToThemLock:
> +                            connection.objectsNewToThem[hashid] = time()
>  
>      def reRandomiseStems(self):
>          with self.lock:
> diff --git a/src/network/invthread.py b/src/network/invthread.py
> index 6f6f1364..37b22cbf 100644
> --- a/src/network/invthread.py
> +++ b/src/network/invthread.py
> @@ -13,26 +13,6 @@ import protocol
>  import state
>  
>  
> -def handleExpiredDandelion(expired):
> -    """For expired dandelion objects, mark all remotes as not having
> -       the object"""
> -    if not expired:
> -        return
> -    for i in \
> -        BMConnectionPool().inboundConnections.values() + \
> -            BMConnectionPool().outboundConnections.values():
> -        if not i.fullyEstablished:
> -            continue
> -        for x in expired:
> -            streamNumber, hashid, _ = x
> -            try:
> -                del i.objectsNewToMe[hashid]
> -            except KeyError:
> -                if streamNumber in i.streams:
> -                    with i.objectsNewToThemLock:
> -                        i.objectsNewToThem[hashid] = time()
> -
> -
>  class InvThread(threading.Thread, StoppableThread):
>      def __init__(self):
>          threading.Thread.__init__(self, name="InvBroadcaster")
> @@ -53,7 +33,7 @@ class InvThread(threading.Thread, StoppableThread):
>              chunk = []
>              while True:
>                  # Dandelion fluff trigger by expiration
> -                handleExpiredDandelion(Dandelion().expire())
> +                Dandelion().expire()
>                  try:
>                      data = invQueue.get(False)
>                      chunk.append((data[0], data[1]))
> ```

PeterSurda: > I'd also suggest to replace this annoying expression

Yes, that would be great, except as you wrote yourself, due to multithreading it creates a problem with generators. One option would be to use locking, but this would negatively improve performance (or could even cause a deadlock in some cases). If on the other hand you do a list of references, it seems to work fine. Perhaps then a method that returns a combined list of references so that you don't have to `+` everytime would work.

PR #1487, opened 2019-07-15
Trying to use Socks5Resolver

g1itch:
Hello!

This is another unfinished work on which I need your comments. It turns out that PyBitmessage can connect to bootstrap nodes by hostname (works well in 994a496). But when I try to use `Socks5Resolver` instead it gets stuck on `state_auth_1`.

It seems that currently `Socks5Resolver` isn't needed at all, but I am thinking of implementing a pluggable  resolvers for using without socks5 proxy.

bm_bootstrap_connect


PeterSurda:
> It seems that currently `Socks5Resolver` isn't needed at all, but I am thinking of implementing a pluggable resolvers for using without socks5 proxy.

Well, yes and no. You can connect to a hostname through socks, or even directly, the issue is that the DNS bootstrap has multiple records, and I don't know for sure if tor or whatever DNS resolver you use rotates them optimally. It probably needs a bit of research if this approach is acceptable.

g1itch:
> Well, yes and no. You can connect to a hostname through socks, or even directly, the issue is that the DNS bootstrap has multiple records, and I don't know for sure if tor or whatever DNS resolver you use rotates them optimally. It probably needs a bit of research if this approach is acceptable.

Once you connected to bootstrap node it sends you `addr` message, so you don't need other bootstrap IPs and may disconnect immediately (or it may disconnect you).

g1itch:
Also `helper_bootstrap.dns()` currently [does the same](../blob/v0.6/src/helper_bootstrap.py#L73).

PeterSurda:
If the resolver only returns one entry and doesn't rotate and that one server is overloaded, you'll get stuck connecting on bootstrap. A normal DNS resolver can return all the entries, which is why `helper_bootstrap.dns` has a cycle.

Now, based on what I remember from my tests, SOCKS can't resolve multiple records. At least that's how tor's SOCKS behaved. Assuming it rotates the IPs the same way whether you connect or resolve, then the resolver probably isn't necessary.

g1itch:
However in practice it gets stuck when connecting by IP after resolution (last commit in this branch). The testing procedure:
 1. put `knownnodes.dat`:
```json
[
    {
     	"peer": {
            "host": "158.222.217.190",
            "port": 8080
        },
	"info": {
            "rating": 0.0,
            "self": false,
            "lastseen": 1562926234
        },
	"stream": 1
    }
]
```
 2. run pybitmessage
 2. ... it cannot connect ):
 2. when you'll stop it there will be an IPs
 2. uncomment [line 166 in connectionpool](1487/files#diff-ea10460170b73ce088d092cdfbbb78d7R166) and comment 167
 2. put the same `knownnodes.dat`
 2. run pybitmessage
 2. it will connect soon

Without tor it seems to work well.

src/api.py
```diff
@@ -204,6 +204,10 @@ def do_POST(self):
             self.wfile.flush()
             self.connection.shutdown(1)
 
+            # actually handle shutdown command after sending response
+            if state.shutdown is False:
```
PeterSurda: I think this is a misunderstanding. It just signals to the peer that you're done writing to the socket. `man 2 shutdown`

src/api.py
```diff
@@ -1386,10 +1390,11 @@ def HandleDeleteAndVacuum(self, params):
         return None
 
     def HandleShutdown(self, params):
-        """Handle a request to huutdown the client"""
+        """Handle a request to shutdown the node"""
 
         if not params:
-            shutdown.doCleanShutdown()
+            # backward compatible trick because False == 0 is True
+            state.shutdown = False
```
PeterSurda: I think this shouldn't change.

src/network/connectionchooser.py
```diff
@@ -44,7 +44,7 @@ def chooseConnection(stream):
             # onion addresses have a higher priority when SOCKS
             if peer.host.endswith('.onion') and rating > 0:
                 rating = 1
-            else:
+            elif not peer.host.startswith('bootstrap'):  # TODO: need better check
```
PeterSurda: Yes, maybe another key (in addition to `lastseen`, `rating` and `self`).

src/knownnodes.py
```diff
@@ -66,7 +65,7 @@ def json_deserialize_knownnodes(source):
         knownNodes[node['stream']][peer] = info
 
         if (
-            not (knownNodesActual or info.get('is_self')) and
+            not (knownNodesActual or info.get('self')) and
```
PeterSurda: good catch

src/network/connectionpool.py
```diff
@@ -141,7 +141,8 @@ def loop(self):
 
         if spawnConnections:
             if not knownnodes.knownNodesActual:
-                helper_bootstrap.dns()
+                knownnodes.dns()
+                knownnodes.knownNodesActual = True
```
PeterSurda: good point

src/network/connectionpool.py
```diff
@@ -67,6 +66,7 @@ def isAlreadyConnected(self, nodeid):
     def addConnection(self, connection):
         if isinstance(connection, UDPSocket):
             return
+        logger.warning('addConnection %s -> %s', connection, connection.destination)
```
PeterSurda: I would use `logger.debug`.

src/network/socks5.py
```diff
@@ -43,15 +45,18 @@ def __init__(self, address=None):
 
     def state_init(self):
         """Protocol initialisation (before connection is established)"""
+        logger.warning('> state_init')
```
PeterSurda: I would use `logger.debug`

src/tests/test_bootstrap.py
```diff
@@ -0,0 +1,47 @@
...
+                    },
+                    "stream": 1
+                }
+            ], dst)
+        cls._started = time.time()
+        super(TestBootstrap, cls).setUpClass()
+
+    def test_connected(self):
+        """Ensure that pybitmessage connects to network in 5 min"""
+        for _ in range(60):
```
PeterSurda: Maybe a different way to check time in case the API call takes too long. And what if the API call freezes, shouldn't there be checks for that as well?

src/network/connectionpool.py
g1itch: All warnings are in 42e32a3 which will be dropped before merging the PR.

src/tests/test_bootstrap.py
g1itch: It's just a first try. This test not actually tests what I need ):

src/api.py
g1itch: I reused this commit from `api` branch because I needed `TestAPIProto`.  Maybe I'll make a separate PR for shutdown to show that it never returns.

g1itch:
I changed my mind (: now, after the investigation, I'm considering bootstrapping and resolving hostnames as different, not dependent tasks.

Your existing `TCPConnection` code fits well for solving the first. The bootstrapping process should be roughly as follows:
 1. notice the lack of knownnodes (actuality, i.e. there are only default and self nodes)
 1. add bootstrapper for one of DNS addresses
 1. if the bootstrapper fails, switch back to not actual state (start from step 1)
 1. when the bootstrapper succeeds (i.e. fully established a connection and received `addr` message), close the connection in order to not abuse bootstrap servers

Such procedure does not imply adding addresses of bootstrap servers to knownnodes and permanent connections to them. Further changes may include filtering the addresses of bootstrap servers out from knownnodes to stop their propagation.

PR #1491, opened 2019-07-25
api_client example standards validation

omkar1117:
PEP8 validations done.

Modified api_client file formatted with PEP8 standards and compatible with python3.

### DIFF
--- a/examples/api_client.py
+++ b/examples/api_client.py
@@ -1,72 +1,77 @@
 # This is an example of how to connect to and use the Bitmessage API.
 # See https://bitmessage.org/wiki/API_Reference
 
-import xmlrpclib
 import json
-import time
+import xmlrpclib
 
 if __name__ == '__main__':
 
     api = xmlrpclib.ServerProxy("http://bradley:password@localhost:8442/")
 
-    print 'Let\'s test the API first.'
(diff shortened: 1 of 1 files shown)


examples/api_client.py
```diff
@@ -1,72 +1,77 @@
...
-
-    print 'Let\'s now print all of our inbox messages:'
-    print api.getAllInboxMessages()
+    print(jsonAddresses)
+    print('Now that we have our address data in a nice Python data structure,'
+          ' let\'s look at the first address (index 0) and print its label:')
+    print(jsonAddresses['addresses'][0]['label'])
+
+    print('Uncomment the next two lines to create a new random address'
+          ' with a slightly higher difficulty setting than normal.')
```
g1itch: As for me, this `print` makes no sense: it tells about the comment line which user doesn't see. Maybe for that reason it was commented (^ plenty of them ahead (8

g1itch:
Are you a robot? Why don't you write a short human readable explanation for you changes? And maybe clarify the goal of them a bit?

examples/api_client.py
dimyme: just to unclutter the screen at runtime and keep syntax hilite

PR #1496, opened 2019-07-31
shared.extPort is obsolete

g1itch:
Another unused variable


g1itch:
Are you agree with the fix in protocol?

PR #1498, opened 2019-08-01
Add self peer into knownnodes if detected external IP with UPnP

g1itch:
If UPnP is used it means that user wants to accept incoming connections. So if there is known external IP, it should be added into knownnodes to propagate in `addr` messages.

### DIFF
--- a/src/knownnodes.py
+++ b/src/knownnodes.py
@@ -98,7 +98,7 @@ def saveKnownNodes(dirName=None):
 def addKnownNode(stream, peer, lastseen=None, is_self=False):
     knownNodes[stream][peer] = {
         "lastseen": lastseen or time.time(),
-        "rating": 0,
+        "rating": 1 if is_self else 0,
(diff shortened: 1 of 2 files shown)


src/knownnodes.py
```diff
@@ -98,7 +98,7 @@ def saveKnownNodes(dirName=None):
 def addKnownNode(stream, peer, lastseen=None, is_self=False):
     knownNodes[stream][peer] = {
         "lastseen": lastseen or time.time(),
-        "rating": 0,
+        "rating": 1 if is_self else 0,
```
PeterSurda: I see, so that it doesn't unnecessarily expire / isn't unnecessarily excluded, makes sense.

g1itch: Moreover, I doubt that expired and low rating nodes should be deleted. Because then them may appear again from `addr` messages with rating 0.

PR #1499, opened 2019-08-01
Do random.seed() in the __init__ of each thread which uses random

g1itch:
The threads using random I found are:
 - BMNetworkThread (StoppableThread)
 - DownloadThread (StoppableThread)
 - UploadThread (StoppableThread)
 - InvThread (StoppableThread)
 - singleWorker (StoppableThread)
 - objectProcessor

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -21,7 +21,6 @@ import json
 import random  # nosec
 import socket
 import subprocess
-import threading
 import time
 from binascii import hexlify, unhexlify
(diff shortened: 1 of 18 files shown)


PeterSurda:
Isn't it easier to put the `random.seed()` into the `__init__` of `StoppableThread`?

g1itch:
I thought of it. But not all of these threads inherit `StoppableThread`.

PeterSurda:
Ok, so how about putting it in the `StoppableThread`, and individually into those who don't inherit it? I also considered a mixin, but mixins don't have an `__init__` so there is no benefit in that.

PR #1507, opened 2019-08-18
New logging approach

g1itch:
Hello!

All that changes are to reduce imports of outer modules from `network` package. Any `StoppableThread` here has a `logger` attribute which can be used instead of one from `debug` module. All other modules inside the `network` use the following code:

```python
import logging

logger = logging.getLogger('default')
```

### DIFF
--- a/src/network/udp.py
+++ b/src/network/udp.py
@@ -2,24 +2,27 @@
 src/network/udp.py
 ==================
 """
+import logging
 import time
 import socket
 
 import state
 import protocol
 from bmproto import BMProto
-from debug import logger
 from objectracker import ObjectTracker
 from queues import receiveDataQueue
 
+logger = logging.getLogger('default')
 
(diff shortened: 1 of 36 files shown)


src/network/udp.py
```diff
@@ -79,7 +82,8 @@ def bm_command_addr(self):
             decodedIP = protocol.checkIPAddress(str(ip))
             if stream not in state.streamsInWhichIAmParticipating:
                 continue
-            if (seenTime < time.time() - self.maxTimeOffset or seenTime > time.time() + self.maxTimeOffset):
+            if (seenTime < time.time() - self.maxTimeOffset
+                    or seenTime > time.time() + self.maxTimeOffset):
```
PeterSurda: Can you fix this one please?

g1itch: This is proper formatting. The fix is in #1539

Issue #1409, opened 2018-12-15
Blind signatures as an alternate proof a message is not spam 


PeterSurda:
# Rationale
Underpowered devices (e.g. on ARM architecture) take very long to calculate the PoW. As time progresses, the difference in hashing capacity may diverge even further, some people unable to communicate with each other. This proposal provides an alternative method for proving a message is not spam, using blind signatures added to the protocol. The exact details will be updated as the proposal evolves.

**The purpose of the blind signatures isn't to replace the PoW mechanism, but to augment it, and to allow alternative method for accessing the network, for example by paying for the blind signatures. The payments can be used to finance the infrastructure and development.**

# TLDR;
The protocol will contain a parallel inventory of objects which are signed using blind signatures, instead of having a nonce for a PoW. A miner would be able to replace the blind signature with a nonce, allowing the object to be transferred to the "normal" inventory, providing backwards compatibility. If the blind signature was obtained against a pament, the miner can participate in the revenue.

If you're familiar with bitcoin, an analogy would be a permissioned unidirectioinal sidechain, which can be used by people who for some reason are unable to use bitcoin directly. Even though this bitmessage "sidechain" doesn't have as strong censorship resistance, through the use of blind signatures, a high level of anonymity is achieved, allowing for censorship resistance to be maintained indirectly. Also, this is only unidirectional, used for sending new objects. For recipients, the miners would provide backwards compatibility and no change is necessary.

# Blind signatures
Blind signatures are a mechanism for certifier to authenticate data without knowing its content. A third party can then use the certifiers public key to verify that a particular piece of data was authenticated by them. An implementation based on pre-existing research, was created for pyelliptic for a bounty: https://bountify.co/blind-signature-library-in-python-or-c

# New object structure
Instead of the "nonce" at the beginning, the object would contain an array of signatures+pubkeys (with a leading varint indicating the depth). Here an example with 2 levels. The lowest level is always a signed inventory vector, and the highest is always the top-level certifier (CA). Mid-level certifiers are possible if they are needed for some reason. This part will probably be updated regarding pubkey structure (e.g. what curve is used etc).

1. depth - varint
2. lvl1sig - signature of the inventory vector
3. lvl1F - F (blinding factor)
4. lvl1pubkey - public key for the signature
5. lvl2sig - signature of the lvl1pubkey
6. lvl2F - F (blinding factor)
7. lvl2pubkey - pubkey of the certifier

# Routing
[...]

PR #1509, opened 2019-08-26
Blind signature support in pyelliptic

PeterSurda:
- add blind signature functionality to pyelliptic as described in #1409
- add tests for blind signatures
- PEP8 fixes for pyelliptic
- some minor refactoring is necessary for further integration, this is just a
  minimal implementation to pass a test

### DIFF
--- a/src/pyelliptic/eccblind.py
+++ b/src/pyelliptic/eccblind.py
@@ -1,0 +1,179 @@
+#!/usr/bin/env python
+"""
+ECC blind signature functionality based on "An Efficient Blind Signature Scheme
+Based on the Elliptic CurveDiscrete Logarithm Problem" by Morteza Nikooghadama
+<mnikooghadam@sbu.ac.ir> and Ali Zakerolhosseini <a-zaker@sbu.ac.ir>,
+http://www.isecure-journal.com/article_39171_47f9ec605dd3918c2793565ec21fcd7a.pdf
+"""
+
+# variable names are based on the math in the paper, so they don't conform
+# to PEP8
+# pylint: disable=invalid-name
+
+from .openssl import OpenSSL
+
+
+class ECCBlind(object):  # pylint: disable=too-many-instance-attributes
+    """
+    Class for ECC blind signature functionality
+    """
+
+    # init
+    k = None
+    R = None
+    keypair = None
+    F = None
+    Q = None
+    a = None
+    b = None
+    c = None
+    binv = None
+    r = None
+    m = None
+    m_ = None
+    s_ = None
+    signature = None
+
+    @staticmethod
+    def ec_get_random(group, ctx):
+        """
+        Random point from finite field
+        """
+        order = OpenSSL.BN_new()
+        OpenSSL.EC_GROUP_get_order(group, order, ctx)
+        OpenSSL.BN_rand(order, OpenSSL.BN_num_bits(order), 0, 0)
+        return order
+
+    @staticmethod
+    def ec_invert(group, a, ctx):
+        """
+        ECC inversion
+        """
+        order = OpenSSL.BN_new()
+        OpenSSL.EC_GROUP_get_order(group, order, ctx)
+        inverse = OpenSSL.BN_mod_inverse(0, a, order, ctx)
+        return inverse
+
+    @staticmethod
+    def ec_gen_keypair(group, ctx):
+        """
+        Generate an ECC keypair
+        """
+        d = ECCBlind.ec_get_random(group, ctx)
+        Q = OpenSSL.EC_POINT_new(group)
+        OpenSSL.EC_POINT_mul(group, Q, d, 0, 0, 0)
+        return (d, Q)
+
+    def __init__(self, curve="secp256k1"):
+        self.group = OpenSSL.EC_GROUP_new_by_curve_name(OpenSSL.get_curve(curve))
+        self.ctx = OpenSSL.BN_CTX_new()
+
+        # Order n
+        self.n = OpenSSL.BN_new()
+        OpenSSL.EC_GROUP_get_order(self.group, self.n, self.ctx)
+
+        # Identity O (infinity)
+        self.iO = OpenSSL.EC_POINT_new(self.group)
+        OpenSSL.EC_POINT_set_to_infinity(self.group, self.iO)
+
+        # Generator G
+        self.G = OpenSSL.EC_GROUP_get0_generator(self.group)
+
+        # Certifier's pubkey
+        self.pubkey = (self.group, self.G, self.n)
+
+    def signer_init(self):
+        """
+        Init signer
+        """
+        # Signer: Random integer k
+        self.k = ECCBlind.ec_get_random(self.group, self.ctx)
+
+        # R = kG
+        self.R = OpenSSL.EC_POINT_new(self.group)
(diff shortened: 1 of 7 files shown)


PeterSurda:
I'll fix the pylint complaints.

src/tests/test_blindsig.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python
```
g1itch: Shebang is not needed here.

src/tests/test_blindsig.py
```diff
@@ -0,0 +1,27 @@
...
+"""
+Test for ECC blind signatures
+"""
+import os
+import unittest
+
+from src.pyelliptic.eccblind import ECCBlind
```
g1itch: `from pybitmessage.pyelliptic.eccblind import ECCBlind`

src/tests/test_blindsig.py
```diff
@@ -0,0 +1,27 @@
...
+        Perform a test of full sequence using a random certifier's key and a
+        random message
+        """
+        blind_sig = ECCBlind()
+        blind_sig.signer_init()
+        msg = os.urandom(64)
+        blind_sig.create_signing_request(msg)
+        blind_sig.blind_sign()
+        blind_sig.unblind()
+        assert blind_sig.verify()
```
g1itch: `self.assertTrue(blind_sig.verify())`

src/tests/test_blindsig.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class TestBlindSig(unittest.TestCase):
+    """
+    Test case for ECC blind signature
+    """
+    @staticmethod
+    def test_blind_sig():
+        """
+        Perform a test of full sequence using a random certifier's key and a
```
g1itch: If you write docstring in one line it will be shown in the test results. e.g.
```python
    def test_blind_sig(self):
        """Test full sequence using a random certifier key and a random msg"""
```

PeterSurda:
@g1itch I'll make the changes you requested.

The low level math I also don't understand, but it looks like it does what the paper says. On a high level, the paper mentions 5 steps (in Section 4), these are roughly mapped to 6 high level methods in the `ECCBlind` class, with some reshuffling because some are executed by the signer and some by the requester. It needs to be refactored, because the signer and requester need to communicate and pass data, so there will be one object at the signer's side and one at the requester's side, but here I just reuse a single object. Also in `__init__`, it now generates a new pubkey, wereas for actual deployment the pubkey will be loaded from the disk (like a CA for traditional PKI). There still are open questions about how to serialise the data. This all will be solved in the subsequent PRs.

src/tests/test_blindsig.py
PeterSurda: This breaks the test.

g1itch: It depends on how you run it.

g1itch: Travis CI should be OK, because I used such imports before: https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/tests/test_config.py#L8

g1itch:
I also expected to find the blinded message and its signature and check the signature's validity.

PeterSurda:
@g1itch The message is only blinded internally and such blind message is only available to the requester, it's not made public. Similarly with the blinded signature, that's calculated by the signer and then transmitted to the requester and unblinded by him.

What you see publicly is the message (in the case of PyBitmessage's wire protocol, this will be the object data), the signature (which has two components, `s` and `F`), and pubkey of the signer (`Q`). There is also `r`, but that's derived from `F`. `F` is a point on a finite field and `r` is its x coordinate modulo n, although it looks to me like the current code skips the modulo n, the way I understand the code is that n is selected from the elliptic curve used, so x can't higher than that. But I don't really understand ECC so I may be wrong.

Then during the verification, you take the signature, the signer's pubkey and the message, and if verification succeeds, it means the singer signed the message, just like with traditional PKI, except here the signer doesn't know which of the requests he signed correspond to which message..

src/pyelliptic/eccblind.py
```diff
@@ -0,0 +1,179 @@
...
+        OpenSSL.EC_GROUP_get_order(group, order, ctx)
+        inverse = OpenSSL.BN_mod_inverse(0, a, order, ctx)
+        return inverse
+
+    @staticmethod
+    def ec_gen_keypair(group, ctx):
+        """
+        Generate an ECC keypair
+        """
+        d = ECCBlind.ec_get_random(group, ctx)
```
g1itch: d = self.ec_get_random(group, ctx) ?

PeterSurda: pylint/flake8 complained.

g1itch: because of `@staticmethod`. well, maybe this really should be a static method

g1itch:
I am trying to split signer and requester with no luck ):
```python
    def test_blind_sig(self):
        """Test full sequence using a random certifier key and a random msg"""
        signer_obj = ECCBlind()
        signer_obj.signer_init()
        signer_obj.create_signing_request('dummy')

        requester_obj = ECCBlind()
        requester_obj.R = signer_obj.R
        requester_obj.keypair = (0, signer_obj.keypair[1])

        msg = os.urandom(64)
        requester_obj.create_signing_request(msg)

        signer_obj.m_ = requester_obj.m_
        signer_obj.blind_sign()

        requester_obj.s_ = signer_obj.s_
        requester_obj.unblind()

        signer_obj.m = requester_obj.m
        signer_obj.signature = requester_obj.signature
        signer_obj.F = signer_obj.signature[1]
        self.assertTrue(signer_obj.verify())
```

PeterSurda:
let me try...

PeterSurda:
@g1itch here you go.

src/tests/test_blindsig.py
```diff
@@ -0,0 +1,35 @@
...
+        # See page 127 of the paper
+        # (1) Initialization
+        signer_obj = ECCBlind()
+        point_r = signer_obj.signer_init()
+
+        # (2) Request
+        requester_obj = ECCBlind(pubkey=signer_obj.pubkey)
+        # only 64 byte messages are planned to be used in Bitmessage
+        msg = os.urandom(64)
+        msg_blinded = requester_obj.create_signing_request(point_r, msg)
```
g1itch: Maybe `self.assertNotEqual(msg, msg_blinded)`?

PeterSurda: Well that will always succeed as msg is a string and msg_blinded is an openssl bignum, but I can convert it before comparison, and I can also compare blinded an unblinded signature (those should both be bignums already).

As I said before, there is still no serialisation for the wire protocol, as that has to be designed first, then there will be minor refactoring here.

g1itch: They will obviously differ. That's just a tests logic (as I understand it) to compare just in case and document it. Maybe that's unnecessary.

Serialization will be probably done in the `network` package, so it's not related to the blind signature implementation itself. The signatures seems to work well. So are you going to merge this now?

PeterSurda: I'll add the tests and then merge.

PeterSurda:
@g1itch I added tests to compare signature with blinded signature, and message with blinded message, all in form which serialises data to bytes with OpenSSL.BN_bn2bin and ctypes.cast. I think this format is usable on wire but more tests are needed.

PR #1513, opened 2019-08-29
Code Quality fixes 

lakshyacis:
# Changes based on Flake8 and Pylint fixes.
announcethread.py Flake8 fixes.
announcethread.py Pylint fixes.
bmproto.py Pylint fixes.
connectionpool.py Pylint fixes.
dandelion.py Pylint fixes.
downloadthread.py Flake8 fixes.
downloadthread.py Pylint fixes.

### DIFF
--- a/src/network/announcethread.py
+++ b/src/network/announcethread.py
@@ -1,3 +1,7 @@
+"""
+src/network/announcethread.py
+=================================
+"""
 import time
 
 from bmconfigparser import BMConfigParser
@@ -10,6 +14,7 @@ import state
 
 
 class AnnounceThread(StoppableThread):
+    """A thread to manage regular announcing of this node"""
     def __init__(self):
         super(AnnounceThread, self).__init__(name="Announcer")
         logger.info("init announce thread")
@@ -24,7 +29,9 @@ class AnnounceThread(StoppableThread):
             if processed == 0:
                 self.stop.wait(10)
 
-    def announceSelf(self):
+    @staticmethod
+    def announceSelf():
+        """Announce our presence"""
         for connection in BMConnectionPool().udpSockets.values():
             if not connection.announcing:
                 continue
--- a/src/network/announcethread.py
+++ b/src/network/announcethread.py
@@ -29,5 +29,8 @@ class AnnounceThread(StoppableThread):
             if not connection.announcing:
                 continue
             for stream in state.streamsInWhichIAmParticipating:
-                addr = (stream, state.Peer('127.0.0.1', BMConfigParser().safeGetInt("bitmessagesettings", "port")), time.time())
+                addr = (
+                    stream,
+                    state.Peer('127.0.0.1', BMConfigParser().safeGetInt("bitmessagesettings", "port")),
+                    time.time())
                 connection.append_write_buf(BMProto.assembleAddr([addr]))
--- a/src/network/bmproto.py
+++ b/src/network/bmproto.py
@@ -1,3 +1,8 @@
+"""
+src/network/bmproto.py
+==================================
+"""
+# pylint: disable=attribute-defined-outside-init
 import base64
 import hashlib
 import socket
@@ -43,6 +48,7 @@ class BMProtoExcessiveDataError(BMProtoError):
 
 class BMProto(AdvancedDispatcher, ObjectTracker):
     """A parser for the Bitmessage Protocol"""
+    # pylint: disable=too-many-instance-attributes, too-many-public-methods
     # ~1.6 MB which is the maximum possible size of an inv message.
     maxMessageSize = 1600100
     # 2**18 = 256kB is the maximum size of an object payload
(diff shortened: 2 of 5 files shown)


src/network/announcethread.py
```diff
@@ -10,6 +14,7 @@
 
 
 class AnnounceThread(StoppableThread):
+    """Method while Announce the thread."""
```
g1itch: Better use docstrings from #1396

lakshyacis: @g1itch sure!

src/network/announcethread.py
```diff
@@ -25,9 +30,14 @@ def run(self):
                 self.stop.wait(10)
 
     def announceSelf(self):
+        """Announce our presence"""
+        # pylint: disable=no-self-use
```
g1itch: Maybe `@staticmethod` was the right decision, why did you change?

lakshyacis: Might be, it was mistakenly done.

src/network/connectionpool.py
```diff
@@ -113,7 +118,7 @@ def removeConnection(self, connection):
                     pass
         connection.handle_close()
 
-    def getListeningIP(self):
+    def getListeningIP(self):       # pylint: disable=no-self-use
```
PeterSurda: this can probably be made into a `@staticmethod`

src/network/connectionpool.py
```diff
@@ -253,8 +259,8 @@ def loop(self):
                     self.startListening()
                 else:
                     for bind in re.sub(
-                        "[^\w.]+", " ",
-                        BMConfigParser().safeGet('network', 'bind')
+                            "[^\w.]+", " ",     # pylint: disable=anomalous-backslash-in-string
```
PeterSurda: maybe this should be a `'` instead of `"`?

src/network/connectionpool.py
```diff
@@ -263,8 +269,8 @@ def loop(self):
                     self.startUDPSocket()
                 else:
                     for bind in re.sub(
-                        "[^\w.]+", " ",
-                        BMConfigParser().safeGet('network', 'bind')
+                            "[^\w.]+", " ",     # pylint: disable=anomalous-backslash-in-string
```
PeterSurda: Same here.

src/network/connectionpool.py
lakshyacis: Sometimes, @staticmethod breaks the code .So, that's why I disable the warning. 
But, will update and correct the code. :)

src/network/connectionpool.py
lakshyacis: Okay, will update the code.

src/network/connectionpool.py
lakshyacis: Okay

src/network/connectionpool.py
PeterSurda: Yes, sometimes it breaks, but I looked at it an in this case it's ok.

src/network/bmproto.py
```diff
@@ -422,16 +430,16 @@ def _decode_addr(self):
...
             decodedIP = protocol.checkIPAddress(str(ip))
             if stream not in state.streamsInWhichIAmParticipating:
                 continue
             if (
-                decodedIP and time.time() - seenTime > 0 and
-                seenTime > time.time() - BMProto.addressAlive and
-                port > 0
+                    decodedIP and time.time() - seenTime > 0 and
+                    seenTime > time.time() - BMProto.addressAlive and
+                    port > 0
```
g1itch: I was trying to save horizontal space and this should not violate PEP8 or any other style guide...

g1itch: So why 996e71a was merged without any warning?

lakshyacis: I think github check didn't run correctly.

PeterSurda: I think it may be a bug with the checks. I'm getting the same error locally.

g1itch: This is really a bug: https://github.com/PyCQA/pylint/issues/289

PeterSurda: For the checks GitHub App, pylint 1.6.5 is used, whereas locally I have 1.9.2. I don't want to change anything as the server seems to misbehave and there is no failover. We'll have to wait until I can put the new system using jenkins into production. I hope it will be by the end of this month.

PeterSurda: I would say we leave these pylint problems for later after the checks are upgraded.

g1itch: I meant we could do something like
```diff
diff --git a/setup.cfg b/setup.cfg
index 3236eed1..68a8181c 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,6 +4,10 @@
 [pycodestyle]
 max-line-length = 119
 
+[pylint.messages control]
+disable = C0330
+# C0330 - wrong-hanging-indentation
+
 [flake8]
 max-line-length = 119
 ignore = E722,F841
```

and invoke pylint with `--rcfile setup.cfg` because if you are going to format the code according to pylint bugs it will become unreadable soon.

PR #1525, opened 2019-09-23
Code Quality Fixes

lakshyacis:
__init__.py pylint fixes
cipher pylint fixes
hash pylint fixes
openssl pylint fixes

### DIFF
--- a/src/pyelliptic/cipher.py
+++ b/src/pyelliptic/cipher.py
@@ -1,5 +1,9 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+"""
+src/pyelliptic/cipher.py
+========================
+"""
 
 #  Copyright (C) 2011 Yann GUIBET <yannguibet@gmail.com>
 #  See LICENSE for details.
@@ -7,7 +11,8 @@
 from openssl import OpenSSL
 
 
-class Cipher:
+# pylint: disable=redefined-builtin
+class Cipher(object):
     """
     Symmetric encryption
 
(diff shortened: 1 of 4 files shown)


src/pyelliptic/cipher.py
```diff
@@ -44,30 +49,34 @@ def get_all_cipher():
 
     @staticmethod
     def get_blocksize(ciphername):
+        """This Method returns cipher blocksize"""
         cipher = OpenSSL.get_cipher(ciphername)
         return cipher.get_blocksize()
 
     @staticmethod
     def gen_IV(ciphername):
+        """"""
```
PeterSurda: `""Generate random initialization vector"""`

src/pyelliptic/cipher.py
```diff
@@ -1,13 +1,18 @@
...
 
 #  Copyright (C) 2011 Yann GUIBET <yannguibet@gmail.com>
 #  See LICENSE for details.
 
 from openssl import OpenSSL
 
 
-class Cipher:
+# pylint: disable=redefined-builtin
+class Cipher:    # pylint: disable=old-style-class
```
PeterSurda: Can you make this a new style class and test if anything breaks? I think that's cleaner.

lakshyacis: Okay, will try and let you know.

lakshyacis: After making old-style class to new style class. All thinks works fine.Also test sending / receiving a message from a different computer.
So should i change old-style-class to new-style-class ?

PeterSurda: Yes please

PR #1526, opened 2019-09-23
Fix deteting from addressbook in bitmessageqt

g1itch:
This fixes the bug I introduced in 34ceb98: cannot delete addressbook entry without label.


g1itch:
Excuse me, please. There are too many of my bugs in the last time ):

Issue #1373, opened 2018-10-12
No more prints


PeterSurda:
There are still occasions where `print` has remained and should be replaced by the `logger.*`. Some can remain, e.g. the notification about daemonizing or missing dependencies. This issue will be updated as I find more locations.

- [ ] `src/api.py`

coffeedogs:
@PeterSurda can we be clear about scope and expected results please? Some thoughts that come to mind:

 * Increase scope to use the opportunity of going through the code thoroughly with a mind on logging
 * Test various logging configurations (which?), remove config cruft, document what to expect out of logging, maybe testing further logging config examples (rsyslog, elasticsearch?)
 * Ensure log levels are appropriate and sent to stdout/stderr as appropriate
 * Consider consolidating the number of log levels in use
 * Consider any refactoring or standardisation of logging code (e.g. avoid code repetition, circular imports) 
 * Consider whether a log entry is actually required; maybe the point of a log entry is to watch out for a condition that better testing could have guarded against 
 * Ensure consistent log message style (tenses, first/third sing/pl person etc.)  
 * Change print statements to use the logger
 * Implement tests around logging (critical examples?)

Quick check of the scale of this task:
12 `logger.exception()` calls
12 `logger.fatal()` calls
13 `logger.critical()` calls
20 `logger.warning()` calls
167 `logger.info()` calls
185 `logger.debug()` calls
467 (approx) `print` statements

navjotcis:
I have removed the print statement of the api.py module.
https://github.com/Bitmessage/PyBitmessage/pull/1673

PR #1529, opened 2019-09-26
Code Quality Fixes

lakshyacis:
core pylint fixes
test_api pylint fixes
test_config pylint fixes

### DIFF
--- a/src/tests/core.py
+++ b/src/tests/core.py
@@ -25,6 +25,7 @@ program = None
 
 
 def pickle_knownnodes():
+    """Generate old style pickled knownnodes.dat"""
     now = time.time()
     with open(knownnodes_file, 'wb') as dst:
         pickle.dump({
@@ -40,6 +41,7 @@ def pickle_knownnodes():
 
 
 def cleanup():
+    """Cleanup application files"""
     os.remove(knownnodes_file)
 
 
--- a/src/tests/test_api.py
+++ b/src/tests/test_api.py
@@ -31,7 +31,7 @@ class TestAPIShutdown(TestAPIProto, TestProcessShutdown):
     """Separate test case for API command 'shutdown'"""
     def test_shutdown(self):
         """Shutdown the pybitmessage"""
-        self.assertEquals(self.api.shutdown(), 'done')
+        self.assertEqual(self.api.shutdown(), 'done')
         for _ in range(5):
             if not self.process.is_running():
                 break
--- a/src/tests/test_config.py
+++ b/src/tests/test_config.py
@@ -26,6 +26,7 @@ class TestConfig(unittest.TestCase):
             False
         )
         # no arg for default
+        # pylint: disable=too-many-function-args
         with self.assertRaises(TypeError):
             BMConfigParser().safeGetBoolean(
                 'nonexistent', 'nonexistent', True)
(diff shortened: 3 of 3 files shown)


src/tests/test_api.py
```diff
@@ -31,7 +31,7 @@ class TestAPIShutdown(TestAPIProto, TestProcessShutdown):
     """Separate test case for API command 'shutdown'"""
     def test_shutdown(self):
         """Shutdown the pybitmessage"""
-        self.assertEquals(self.api.shutdown(), 'done')
+        self.assertEqual(self.api.shutdown(), 'done')
```
g1itch: oops, it's really deprecated since 2.7

src/tests/core.py
```diff
@@ -91,7 +93,7 @@ def test_tcpconnection(self):
             for peer in (state.Peer("127.0.0.1", 8448),):
                 direct = TCPConnection(peer)
                 while asyncore.socket_map:
-                    print("loop, state = %s" % direct.state)
+                    print "loop, state = %s" % direct.state
```
g1itch: print function is used intentionally and it is valid for python 2.7.6 (minimum version supported) Any old style prints in code should be replaced if `print` cannot be avoided there (#1373) in order to prepare for python 3 support.

https://docs.python.org/3/whatsnew/3.0.html#print-is-a-function

src/tests/core.py
```diff
@@ -40,6 +41,7 @@ def pickle_knownnodes():
 
 
 def cleanup():
+    """Cleanup knownnodes: removing nodes"""
```
g1itch: Cleanup application files

src/tests/core.py
```diff
@@ -25,6 +25,7 @@
 
 
 def pickle_knownnodes():
+    """Organize knownnodes dict"""
```
g1itch: Generate old style pickled knownnodes.dat

src/tests/test_api.py
lakshyacis: The source code still in python 2.7, and the method alos deprecates in 2.. So, should i replace with assertEquals or leave it as it is.

src/tests/core.py
lakshyacis: So should i change it with logger?

src/tests/core.py
lakshyacis: Okay, thanks!

src/tests/core.py
lakshyacis: Okay, thanks!

src/tests/test_api.py
g1itch: Yes, replace. I agree (`assertEquals`, which I used is already deprecated), just noticed when you wrote. That's what my comment about. Sorry if it's not clear enough.

src/tests/core.py
g1itch: Better do not touch this broken test. I'm going to find an appropriate environment to run it. The print was in the original script.

src/tests/test_api.py
lakshyacis: No problem! :)

lakshyacis: No problem! :)

src/tests/core.py
lakshyacis: Okay, wil change "print "loop, state = %s" % direct.state" -> print("loop, state = %s" % direct.state).

PeterSurda: The print was there for debugging I believe, for tests we can keep it.

PR #1539, opened 2019-10-08
Follow PEP8 and trust Guido

g1itch:
Hello!

This is an appeal to contributors: write more pythonic code! Please read PEP8 carefully and don't trust linters bugs.

About 79 characters line length limit, I repeat it here: readability has nothing to do with your screen size.

Not sure is it correct to add symlink into repository. Some setuptools commands wont work without it.


PeterSurda:
I'm indifferent to the line length, however this is not a priority. There are plenty of other things that need to be fixed first, and this will screw up the code quality checks.

g1itch:
I'm afraid you cannot improve code quality violating the PEP8. PEP8 is a prerequisite for good quality code in python. Your other contributors keep formatting the code in pasta style and then I ask about PEP8 line length they reference the `setup.cfg`. This leads to recursion, when any new code I write is considered as of poor quality and becomes corrupted with random garbage.

@coffeedogs added that `max-line-length = 119` and leaved us. So now anyone who edits the code is forced to tolerate the eyes bleeding and waste time to parse that enormous lines trying to understand the logic.

PeterSurda:
There are more important code quality issues than line length. If we change the settings for line length now, it will delay other work. The code quality improvements have now been worked on for over year and a half and it's still not done.

PeterSurda:
What could work as a compromise for thebtime being is to make the checks success ignore line length warnings. Then I can accept this PR without delaying other work.

g1itch:
I guess that pylint checks shouldn't apply to any files outside the PR.

PeterSurda:
> I guess that pylint checks shouldn't apply to any files outside the PR.

That's an excellent suggestion. I'll see how complicated that is.

PeterSurda:
> Also it's better to delete more code than trying to "improve" code quality especially by adding the `pylint: disable=...` comments and weird formatting.

Yes, but there isn't anyone who can do that.

g1itch:
I'm doing my best to delete as much as possible

PeterSurda:
This PR breaks the checks somehow, I need to find out what's happening.

g1itch:
It's strange. Since I reverted the `max-line-length` it shows me

> All checks have passed
> 4 successful checks

PeterSurda:
If you look at the details, it shows that everything has been fixed, which is obviously not true.

g1itch:
Another 85 character "I just rewrite all your formatting" change ]:
https://github.com/Bitmessage/PyBitmessage/pull/1541/files#diff-da26b2a52200f6e3923af764a2d84e6bR154

g1itch:
I have no Idea what else this changes may break. Here is only a symlink and one ignored frake8 warning.

PeterSurda:
Try removing the symlink for now. I don't know how to fix the checks quickly and don't have time for more.

g1itch:
> Try removing the symlink for now. 

Is it OK if the symlink will appear in #1542?

PeterSurda:
I can't accept the symlink until the checks are fixed or migrated.

g1itch:
> I can't accept the symlink until the checks are fixed or migrated.

Can you please show me some of that failing checks?

PeterSurda:
It looks like the symlink triggers a false positive in the check system. It reports that the code doesn't contain any more problems, which is demonstrably false.

g1itch:
Do you mean the whole code? Because this PR doesn't contain any code.

PeterSurda:
Look at the check result details. It says this PR fixes everything.

g1itch:
Oh, I see, the **remaining** number

g1itch:
This is weird, it shows thousands of problems remaining in #1542, but it also contains the symlink.

PeterSurda:
Maybe I'm misinterpreting something or the problem is more complicated.

PeterSurda:
Maybe it's triggered by actually there being no changed code files. Try to commit one python file.

g1itch:
Actually it seems something wrong with processing of `setup.cfg`. So I return symlink.

Issue #1538, opened 2019-10-06
Add onlynet=onion config option


gousasin:
Please add an onlynet=onion configuration option like in Bitcoin Core to exclusively allow connections to Bitmessage nodes behind a Tor Onion Service for enhanced privacy at the network layer.

sgj3:
Implemented by PR #1545.

g1itch:
> Implemented by PR #1545.

Close but not exact implementation. In the future PR's consider to put [closes #](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords) in the commit message.

sgj3:
So, do you want this refactored to use literally "onlynet=onion"?

gousasin:
I don't really care about the syntax if the functionality is implemented. "onlynet=onion" is used by Bitcoin so I just copied it fom there. I leave it up to you.

sgj3:
OK, understood.

@g1itch What do you think?  You left it open and assigned the `refactoring` label.  Is this is what you had in mind?

sgj3:
@g1itch Bump

g1itch:
Besides the refactoring there is a new complaint from chan:

> > If onionservicesonly is enabled and there are no available onion bootstrap servers PyBitmessage will never connect to the network without connecting to bootstrapXXXX.
> 
> When onionservicesonly is enabled, the user expects PyBitmessage not to connect non-onion nodes at all.
> So if onion bootstrap nodes are not to be connected, PyBitmessage should stop bootstrapping and show error message.
> The user then decides not to use PyBitmessage for a while or to change his policy to connet to non-onion nodes.
> Security policy is more important than connectivity.
> ...
> ------------------------------------------------------
> If onionservicesonly is enabled and there are no available onion bootstrap servers PyBitmessage will never connect to the network without connecting to bootstrapXXXX.
> ...
> ------------------------------------------------------
> PyBitmessage should not connect to bootstrapXXXX when:
> * onionservicesonly is enabled, or
> * trusted mode is used.
> Currently there are no guards for these cases.

Thought it cannot be implemented right now because quzwelsuziwqgpt2.onion (sole onion bootstrap server) is down.

Issue #105, opened 2013-04-04
Feature Request: Message Encoding 'FILE'


ISibboI:
It would be really cool if there were a message encoding type to transfer files, in simple binary format.

PeterSurda:
A more flexible encoding is planned for 0.8, it may not be necessarily exactly like the one described here but will allow the same goals. I will keep this open to compare when I'm implementing it.

PR #1540, opened 2019-10-08
Seamless file attachments and option to connect only to Tor onion services

sgj3:
This PR makes the following changes:
1. The message composition 'Send' tab gets a new 'Attach File' button that launches a file picker and appends the base64-encoded contents of the chosen file onto the message being composed.
2. In HTML message view, clicking such a data blob link launches a file picker, and the base64-decoded contents of the data blob are written to the specified file.
3. The network settings tab  gets a new checkbox which allows the user to restrict connection attempts only to hosts that end with '.onion' (i.e., onion services).  This enables Bitmessage users to keep their Bitmessage network traffic within the Tor overlay network and completely avoid exit nodes in order to frustrate traffic analysis attacks.

See issues #1538 and #105

### DIFF
--- a/src/paths.py
+++ b/src/paths.py
@@ -1,76 +1,85 @@
-from os import environ, path
-import sys
+import logging
+import os
 import re
+import sys
 from datetime import datetime
+from shutil import move
+
+
+logger = logging.getLogger('default')
 
 # When using py2exe or py2app, the variable frozen is added to the sys
-# namespace.  This can be used to setup a different code path for 
+# namespace.  This can be used to setup a different code path for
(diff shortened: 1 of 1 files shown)


sgj3:
@PeterSurda Any comments?

g1itch:
Maybe you'll split this into two PR's? That features are not related.

PeterSurda:
For attachments, there is a reserved key in the extended encoding, but it isn't implemented. I would prefer that is implemented and finished, but it probably requires some alterations in the database: https://bitmessage.org/wiki/Extended_encoding

PR #1545, opened 2019-10-21
Add "onion services only" or Tor-only networking mode

sgj3:
See PR #1540.


g1itch:
Apart from the fact that config syntax differs from bitcoin's one (`onlynet=<net>`) this looks good.

g1itch:
Notice also that internal tests (`tests.core`) are currently broken. I'm still debugging the use of `stem` inside of the docker. Hope to finish that job in #1547.

g1itch:
Please format commit messages in the next PR's. You can also use first 7 characters of commit hash for reference.

PR #1547, opened 2019-10-22
Support for socksproxytype plugins in Settings dialog

g1itch:
Hello.

With this changes now it's possible to set "Proxy server / Tor Type:" to any one provided by plugin. `start_proxyconfig` will be called when user press "OK" in the Settings dialog.

This is a draft because I'm trying to fix test and maybe port in `proxyconfig_stem`.

### DIFF
--- a/src/helper_startup.py
+++ b/src/helper_startup.py
@@ -2,11 +2,12 @@
 Startup operations.
 """
 # pylint: disable=too-many-branches,too-many-statements
-from __future__ import print_function
 
+import logging
 import os
 import platform
 import sys
+import time
 from distutils.version import StrictVersion
 
 import defaults
@@ -15,6 +16,13 @@ import paths
 import state
 from bmconfigparser import BMConfigParser
 
+try:
+    from plugins.plugin import get_plugin
+except ImportError:
+    get_plugin = None
+
+
+logger = logging.getLogger('default')
 
 # The user may de-select Portable Mode in the settings if they want
 # the config files to stay in the application data folder.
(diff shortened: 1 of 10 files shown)


dimyme:
> it's possible to set "Proxy server / Tor Type:" to any one provided by plugin:

meaning i.e.:  proxy_type in SOCKS4a, SOCKS5

any one of these two, choose freely!

g1itch:
> > it's possible to set "Proxy server / Tor Type:" to any one provided by plugin:
> 
> meaning i.e.: proxy_type in SOCKS4a, SOCKS5
> 
> any one of these two, choose freely!

Seriously? tl;dr!

Actually, I meant 'stem' or any other plugin for the `bitmessage.proxyconfig` entry point. Another silly comment and I'll report abuse.

src/helper_startup.py
```diff
@@ -30,14 +33,14 @@ def loadConfig():
         needToCreateKeysFile = config.safeGet(
             'bitmessagesettings', 'settingsversion') is None
         if not needToCreateKeysFile:
-            print(
+            logger.info(
```
PeterSurda: Good, these are the few places where I left print due to initialisation order, which you now fixed.

g1itch:
I made this "Ready for review" because it contains the solution for periodic Travis CI fails, e.g. https://travis-ci.org/Bitmessage/PyBitmessage/builds/629056869: I separated clearnet and tor tests and marked tor tests for skipping (because stem package isn't installed).

Manual testing scenario I used for this changeset (and it fails):
 1. on first start select "Let me configure network..." in Connection dialog;
 1. set Proxy server / Tor Type to 'stem'
 1. pybitmessage should connect to the network through tor and start accepting incoming connections
 1. go back to the network settings and set Proxy server / Tor Type to 'none'
 1. pybitmessage should connect to the network directly using the same port 8444

PR #1550, opened 2019-11-04
Heavy imports refactoring

g1itch:
Hello!

With this changes
  * all members of `network` used outside it are available directly from package: `from network import ..`
  * all top level threads in `threads` module
  * `state.Peer` is moved to `network.node`
  * shortcuts for connections and establishedConnections in network.BMConnectionPool
  * the project imports in `bitmessagemain` are formatted alphabetically

Please, help me write more readable docstrings.

Relevant docs are:
https://pybitmessage-test.readthedocs.io/en/latest/autodoc/pybitmessage.threads.html
https://pybitmessage-test.readthedocs.io/en/latest/autodoc/pybitmessage.network.html


src/class_objectProcessor.py
```diff
@@ -471,6 +473,8 @@ def processmsg(self, data):
             try:
                 # continue decryption attempts to avoid timing attacks
                 if initialDecryptionSuccessful:
+                    if BMConnectionPool().trustedPeer:
```
PeterSurda: Interesting.

src/network/connectionpool.py
```diff
@@ -26,9 +27,24 @@
...
+    If the trustedpeer option is specified in keys.dat then this will
+    contain a Peer which will be connected to instead of using the
+    addresses advertised by other peers. The client will only connect to
+    this peer and the timing attack mitigation will be disabled in order
+    to download data faster. The expected use case is where the user has
+    a fast connection to a trusted server where they run a Bitmessage
+    daemon permanently. If they then run a second instance of the client
+    on a local machine periodically when they want to check for messages
+    it will sync with the network a lot faster without compromising
+    security.
```
PeterSurda: I find it unlikely that the performance difference will be important. First, download happens asynchronously and only stalls if the decode buffer (32MB) gets full. Second, if an object isn't for you, this performance optimisation makes no difference. So if almost none of the objects are for you (as it is normally), there is no difference. Third, if you have a lot of addresses, assuming a random order, it still needs to process at least 50% on average.

So, in total, you will save 50% of time for processing objects that are for you, and only download slower to the extent that this affects a large processing buffer.

I think the risk of introducing potential attack vectors isn't worth it.

src/network/connectionpool.py
```diff
@@ -41,9 +57,32 @@ def __init__(self):
...
-        self.bootstrapped = False
+        self._lastSpawned = 0
+        self._spawnWait = 2
+        self._bootstrapped = False
+
+        trustedPeer = BMConfigParser().safeGet('bitmessagesettings', 'trustedpeer')
+        try:
+            if trustedPeer:
+                host, port = trustedPeer.split(':')
+                self.trustedPeer = Peer(host, int(port))
```
PeterSurda: This could cause problems in the future if `trustedpeer` can be modified during runtime (now it can't so it should work correctly).

PeterSurda:
I don't see where `BMConnectionpool.connections` is defined.

src/network/connectionpool.py
```diff
@@ -247,7 +286,7 @@ def loop(self):  # pylint: disable=too-many-branches,too-many-statements
                 for i in range(
                         state.maximumNumberOfHalfOpenConnections - pending):
                     try:
-                        chosen = chooseConnection(
+                        chosen = self.trustedPeer or chooseConnection(
```
PeterSurda: nice

src/network/connectionpool.py
g1itch: Well, as you probably see, it is not originally my docstring. I moved it from `state`, but its first appearance is in `shared` in 0326315.

src/class_objectProcessor.py
g1itch: I can drop e002984. It's made in the separate commit intentionally. But then I should probably change the docstring from previous commit. Because if it will appear in the doc, it may be interpreted as promise.

src/network/connectionpool.py
g1itch: Same here: I just moved this from `helper_startup`. Introduced in 0326315.

src/network/connectionpool.py
PeterSurda: That refers to an older implementation and a different aspect. It used delays while downloading, while now it downloads as quickly as the CPU allows. This was prior to moving to asyncore. Even further in the past the downloading and decryption were in the same thread so the delays made sense, it probably was an artifact of this, but that was even before I started contributing.

The change you made changes the "constant time decryption" approach I introduced in 9239813ebb. It probably still isn't constant as it still skips AES decryption on a failure, but I think it's still an improvement as AES is very fast compared to ECC.

g1itch:
What about module docstrings? Are you OK with what I wrote in `threads`? Maybe also write something in `network`?

PeterSurda:
I personally don't care much about docstrings. Only the one in trustedpeer is obsolete and misleading. Trustedpeer doesn't really improve performance anymore, in fact it may well make things slower, but it is helpful for situations with limited network access, or when trying to isolate your instance.

g1itch:
> I personally don't care much about docstrings. Only the one in trustedpeer is obsolete and misleading. Trustedpeer doesn't really improve performance anymore, in fact it may well make things slower, but it is helpful for situations with limited network access, or when trying to isolate your instance.

I already dropped the corresponding commit and edited the docstring.

PeterSurda:
I'm happy about the refactoring though.

PeterSurda:
> I already dropped the corresponding commit and edited the docstring.

I still see it there. The explanation that it's for a situation for a fast connection to a permanent node isn't valid anymore. It used to be like that a long time ago but as I said, after moving to asyncore it isn't true.

g1itch:
> > I already dropped the corresponding commit and edited the docstring.
> 
> I still see it there. The explanation that it's for a situation for a fast connection to a permanent node isn't valid anymore. It used to be like that a long time ago but as I said, after moving to asyncore it isn't true.

Well, the relevant line in the commit: https://github.com/Bitmessage/PyBitmessage/pull/1550/commits/ebd12ab5c920b6e6342d54dc0f5ef54bad730cfd#diff-ea10460170b73ce088d092cdfbbb78d7R34

Maybe I replace

>     The expected use case is where the user has a fast connection to
>     a trusted server where they run a Bitmessage daemon permanently.

by

>     The expected use case is where the user has
>     a trusted server where they run a Bitmessage daemon permanently.

PeterSurda:
> Maybe I replace ...

Yes that sounds better. The emphasis of trust is correct, only regarding speed it is wrong.

g1itch:
OK, I'm merging it...

PR #1552, opened 2019-11-10
fix typos

dimyme:
and flesh out placeholder


dimyme:
trivial typos. no big deal.

PR #1559, opened 2019-11-14
A check for QSystemTrayIcon availability

g1itch:
Hello!

It turns out that `sni-qt` package is required to have functional system tray icon in Ubuntu. I added a check.


dimyme:
with sni-qt present, the trayicon still does not show the number of new posts on Fedora 31  :cactus:

g1itch:
Excuse me, but that wont work. The `checkdeps` script is write only. I just cannot handle so much pasta-code.

PR #1561, opened 2019-11-15
uncrumple settings window

dimyme:
full size right away 900 wide now,  with no need to manually enlarge it.


PeterSurda:
If I understand right, this will prevent it from working correctly on 1024x768. I think that should be the cutoff line.

PeterSurda:
Also squash commits, the commit messages are crap.

dimyme:
what actually is crap is the SPACE py2 grammar sticklers.

dimyme:
> If I understand right, this will prevent it from working correctly on 1024x768. I think that should be the cutoff line.

modded to 900 wide

PR #1562, opened 2019-11-15
More docstrings and formatting

g1itch:
Hello.

I reformatted for PEP8 modules `addresses`, `highlevelcrypto`, `protocol`, `shared`, `shutdown`, `state` and edited the docstrings.

There is also `.readthedocs.yml` which should finally help rftd to make proper documentation because it seems that you are not going to edit the Advanced Settings. I removed my Advanced Settings except for ''Default Branch" and there is the result: https://pybitmessage-test.readthedocs.io/en/latest/ (same blank pages with module names). In the opposite there is the documentation for the "doc" branch: https://pybitmessage-test.readthedocs.io/en/doc/index.html (with content collected from docstrings).

### DIFF
--- a/src/protocol.py
+++ b/src/protocol.py
@@ -1,7 +1,8 @@
-# pylint: disable=too-many-boolean-expressions,too-many-return-statements,too-many-locals,too-many-statements
 """
 Low-level protocol-related functions.
 """
+# pylint: disable=too-many-boolean-expressions,too-many-return-statements
+# pylint: disable=too-many-locals,too-many-statements
 
 import base64
 import hashlib
@@ -9,7 +10,6 @@ import random
 import socket
 import sys
 import time
-import traceback
 from binascii import hexlify
 from struct import pack, unpack, Struct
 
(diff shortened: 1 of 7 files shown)


src/protocol.py
```diff
@@ -24,10 +24,18 @@
...
+#: This is a normal network node
 NODE_NETWORK = 1
+#: This node supports SSL/TLS in the current connect (python < 2.7.9
+#: only supports an SSL client, so in that case it would only have this
+#: on when the connection is a client).
 NODE_SSL = 2
+# (Proposal) This node may do PoW on behalf of some its peers
+# (PoW offloading/delegating), but it doesn't have to. Clients may have
+# to meet additional requirements (e.g. TLS authentication)
+# NODE_POW = 3
```
PeterSurda: It would be 4 and not 3

PeterSurda: or `1 << 2`, then change `NODE_SSL = 1 << 1` and `NODE_DANDELION = 1 << 3`

src/shutdown.py
```diff
@@ -16,7 +16,10 @@
 
 
 def doCleanShutdown():
-    """Used to tell proof of work worker threads and the objectProcessorThread to exit."""
+    """
+    Used to tell proof of work worker threads and the
+    `.threads.objectProcessor` thread to exit.
```
PeterSurda: Well now it notifies all kinds of threads, so the reference to "worker" should be removed.

src/state.py
```diff
@@ -16,16 +16,16 @@
 
 shutdown = 0
 """
-    Set to 1 by the doCleanShutdown function.
-    Used to tell the proof of work worker threads to exit.
+Set to 1 by the `.shutdown.doCleanShutdown` function.
+Used to tell the proof of work worker threads to exit.
```
PeterSurda: Remove reference to "worker" thread and use just threads in general.

PR #1564, opened 2019-11-17
Add flushing to log test

PeterSurda:
I hope this fixes a bug

### DIFF
--- a/src/tests/test_logger.py
+++ b/src/tests/test_logger.py
@@ -7,6 +7,8 @@ import os
 import tempfile
 import unittest
 
+import pybitmessage.state
+
 
 class TestLogger(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


PeterSurda:
@g1itch Not sure why it's broken, can you look at it since you're more familiar with this part of the code.

g1itch:
> @g1itch Not sure why it's broken, can you look at it since you're more familiar with this part of the code.

It breaks because of import from `protocol` [here](https://github.com/Bitmessage/PyBitmessage/blob/5495c8cfcf4bc88d26053b901e00f2a3897af305/src/tests/test_networkgroup.py#L6). It starts full cascade of imports which sets `state.appdata` before the `test_logger`.

PeterSurda:
I see, so this PR is useless.

PR #1565, opened 2019-11-18
catch errors, exit cleanly

dimyme:
avoid console mess-up, BaseException


PeterSurda:
Can you squash the commits and use `BaseException`?

dimyme:
I already did a squash. it does not work. this stuff takes hours.

PR #1573, opened 2019-11-26
Formatting, docstring and minor changes in storage

lakshyacis:
Formatting, docstring and minor changes in storage
=================================================================
filesystem quality fixes
sqlite quality fixes
storage quality fixes

### DIFF
--- a/src/storage/filesystem.py
+++ b/src/storage/filesystem.py
@@ -1,6 +1,5 @@
 """
-src/storage/filesystem.py
-=========================
+Module for using filesystem (directory with files) for inventory storage
 """
 from binascii import hexlify, unhexlify
 from os import listdir, makedirs, path, remove, rmdir
@@ -12,8 +11,9 @@ from paths import lookupAppdataFolder
 from storage import InventoryStorage, InventoryItem
 
 
-class FilesystemInventory(InventoryStorage):    # pylint: disable=too-many-ancestors, abstract-method
-    """Module for using filesystem (directory with files) for inventory storage"""
+class FilesystemInventory(InventoryStorage):
+    """Filesystem for inventory storage"""
+    # pylint: disable=too-many-ancestors, abstract-method
     topDir = "inventory"
     objectDir = "objects"
     metadataFilename = "metadata"
@@ -21,21 +21,23 @@ class FilesystemInventory(InventoryStorage):    # pylint: disable=too-many-ances
 
     def __init__(self):
         super(FilesystemInventory, self).__init__()
-        self.baseDir = path.join(lookupAppdataFolder(), FilesystemInventory.topDir)
+        self.baseDir = path.join(
+            lookupAppdataFolder(), FilesystemInventory.topDir)
         for createDir in [self.baseDir, path.join(self.baseDir, "objects")]:
             if path.exists(createDir):
                 if not path.isdir(createDir):
-                    raise IOError("%s exists but it's not a directory" % (createDir))
+                    raise IOError(
+                        "%s exists but it's not a directory" % createDir)
             else:
                 makedirs(createDir)
-        # Guarantees that two receiveDataThreads don't receive and process the same message
+        # Guarantees that two receiveDataThreads
+        # don't receive and process the same message
         # concurrently (probably sent by a malicious individual)
         self.lock = RLock()
         self._inventory = {}
         self._load()
 
     def __contains__(self, hashval):
-        retval = False
         for streamDict in self._inventory.values():
             if hashval in streamDict:
                 return True
@@ -48,7 +50,12 @@ class FilesystemInventory(InventoryStorage):    # pylint: disable=too-many-ances
             except KeyError:
                 continue
             if retval.payload is None:
-                retval = InventoryItem(retval.type, retval.stream, self.getData(hashval), retval.expires, retval.tag)
+                retval = InventoryItem(
+                    retval.type,
+                    retval.stream,
+                    self.getData(hashval),
+                    retval.expires,
+                    retval.tag)
             return retval
         raise KeyError(hashval)
 
@@ -56,7 +63,10 @@ class FilesystemInventory(InventoryStorage):    # pylint: disable=too-many-ances
         with self.lock:
             value = InventoryItem(*value)
             try:
-                makedirs(path.join(self.baseDir, FilesystemInventory.objectDir, hexlify(hashval)))
+                makedirs(path.join(
+                    self.baseDir,
+                    FilesystemInventory.objectDir,
+                    hexlify(hashval)))
             except OSError:
                 pass
             try:
@@ -69,7 +79,11 @@ class FilesystemInventory(InventoryStorage):    # pylint: disable=too-many-ances
                     ),
                     "w",
                 ) as f:
-                    f.write("%s,%s,%s,%s," % (value.type, value.stream, value.expires, hexlify(value.tag)))
+                    f.write("%s,%s,%s,%s," % (
+                        value.type,
+                        value.stream,
+                        value.expires,
+                        hexlify(value.tag)))
                 with open(
                     path.join(
                         self.baseDir,
@@ -115,7 +129,10 @@ class FilesystemInventory(InventoryStorage):    # pylint: disable=too-many-ances
             except IOError:
                 pass
             try:
-                rmdir(path.join(self.baseDir, FilesystemInventory.objectDir, hexlify(hashval)))
+                rmdir(path.join(
+                    self.baseDir,
+                    FilesystemInventory.objectDir,
+                    hexlify(hashval)))
             except IOError:
                 pass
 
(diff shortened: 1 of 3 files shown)


src/storage/storage.py
```diff
@@ -1,21 +1,21 @@
...
 import collections
 
-InventoryItem = collections.namedtuple('InventoryItem', 'type stream payload expires tag')
+InventoryItem = collections.namedtuple(
+    'InventoryItem', 'type stream payload expires tag')
 
 
-class Storage(object):
-    """Base class for storing inventory (extendable for other items to store)"""
+class Storage(object):  # pylint: disable=too-few-public-methods
```
lakshyacis: @g1itch What if we remove this class?

g1itch: better ask @PeterSurda, I have no complete understanding of this module. If you are worried about pylint `too-few-public-methods` concider adding `__metaclass__ = ABCMeta`.

g1itch: It appeared in 36b5e2c

PeterSurda: I think the term for this is abstract class. Maybe there is a way to write it in a more pythonic way, such as defining the methods but have them all contain just `pass`

src/storage/filesystem.py
```diff
@@ -1,6 +1,5 @@
 """
-src/storage/filesystem.py
-=========================
+Filesystem for inventory storage
```
g1itch: Filesystem inventory storage

src/storage/filesystem.py
```diff
@@ -12,30 +11,34 @@
 from storage import InventoryStorage, InventoryItem
 
 
-class FilesystemInventory(InventoryStorage):    # pylint: disable=too-many-ancestors, abstract-method
-    """Module for using filesystem (directory with files) for inventory storage"""
+class FilesystemInventory(InventoryStorage):
+    """Module for using filesystem (directory with files)
```
g1itch: Maybe make this module level docstring and vice versa?

src/storage/filesystem.py
```diff
@@ -185,23 +205,33 @@ def getMetadata(self, hashId):
...
+                    int(expiresTime),
+                    unhexlify(tag)]
         except IOError:
             raise KeyError
 
     def by_type_and_tag(self, objectType, tag):
         """Get a list of objects filtered by object type and tag"""
         retval = []
-        for stream, streamDict in self._inventory:
+        for streamDict in self._inventory[1]:
```
g1itch: This wont work

src/storage/storage.py
```diff
@@ -1,21 +1,21 @@
...
+    """Base class for storing inventory(extendable for other items to store)"""
```
g1itch: Why are you removing the space?

src/storage/storage.py
```diff
@@ -53,8 +53,8 @@ def clean(self):
         raise NotImplementedError
 
 
-class MailboxStorage(Storage, collections.MutableMapping):    # pylint: disable=abstract-method
+class MailboxStorage(Storage, collections.MutableMapping):
     """Method for storing mails"""
-    def __init__(self):
-        # pylint: disable=super-init-not-called
+    # pylint: disable=abstract-method
+    def __init__(self):   # pylint: disable=super-init-not-called
```
g1itch: This `__init__` makes no sense.

src/storage/filesystem.py
lakshyacis: yeah, make sense. Thank you

src/storage/filesystem.py
lakshyacis: I am not tested this, but I thought this would work. btw thank you, i will change the soon

src/storage/storage.py
lakshyacis: By mistake. Trying to make the line within 79 length. Forget to create new line. But Fixed now. Thanks!

src/storage/storage.py
lakshyacis: Yeah, Discussed with peter too. MailboxStorage is an abstract class. Need more things to be added in that

src/storage/storage.py
g1itch: I see. Such economy makes sense only in tests where docstrings are showed below the test method names.

src/storage/storage.py
g1itch: No need to define methods because parent class already defined them

src/storage/storage.py
g1itch: So make it abstract. What a problem with using [ABCMeta](https://docs.python.org/2.7/library/abc.html?highlight=abcmeta#abc.ABCMeta)?

src/storage/filesystem.py
g1itch: So test. You iter through element 1 of inventory instead of inventory itself. It seems a mistake.

src/storage/storage.py
lakshyacis: Like this?
> class MailboxStorage(Storage, collections.MutableMapping):
>     """Method for storing mails"""
>     __metaclass__ = ABCMeta
> 
>     @abstractmethod
>     def __init__(self):
>         pass

lakshyacis: Like this?
class MailboxStorage(Storage, collections.MutableMapping):
     """Method for storing mails"""
    __metaclass__ = ABCMeta

    @abstractmethod
    def __init__(self):
        pass

g1itch: without `__init__`, just pass

g1itch: `__init__` cannot be an abstract method by it's nature

lakshyacis: I am doing this now, but still got warnings. 
```
class MailboxStorage(Storage, collections.MutableMapping):
    """Method for storing mails"""
    __metaclass__ = ABCMeta

    pass
```
https://imgur.com/a/HTlWMOu

g1itch: Because you are doing it in the wrong place

```python
class InventoryStorage(collections.MutableMapping):
    """Base class for inventory storage"""
    numberOfInventoryLookupsPerformed = 0

    @abstractmethod
    def __contains__(self, _):
        pass

    @abstractmethod
    def __iter__(self):
        pass

    @abstractmethod
    def __len__(self):
        pass

    @abstractmethod
    def by_type_and_tag(self, objectType, tag):
        """Return objects filtered by object type and tag"""
        pass

    @abstractmethod
    def unexpired_hashes_by_stream(self, stream):
        """Return unexpired inventory vectors filtered by stream"""
        pass

    @abstractmethod
    def flush(self):
        """Flush cache"""
        pass

    @abstractmethod
    def clean(self):
        """Free memory / perform garbage collection"""
        pass
```

Notice `collections.MutableMapping` is abstract class itself

g1itch: MailboxStorage class isn't used currently

lakshyacis: The functions that are defined on MutuableMapping are already abstract, like __len__ is decorated as abstract. So, why we decorate with abstractmethod here too? We can just override methods that are static? 
Can you elaborate me something more about this, please?

g1itch: In the current code nothing is decorated. Why I decorated `__iter__`  and `__len__` in this example? I don't remember - it was taken from unfinished refactoring branch.

PeterSurda:
@g1itch any more objections?

src/storage/filesystem.py
```diff
@@ -12,30 +12,33 @@
...
         super(FilesystemInventory, self).__init__()
-        self.baseDir = path.join(lookupAppdataFolder(), FilesystemInventory.topDir)
+        self.baseDir = path.join(
+            lookupAppdataFolder(), FilesystemInventory.topDir)
         for createDir in [self.baseDir, path.join(self.baseDir, "objects")]:
             if path.exists(createDir):
                 if not path.isdir(createDir):
-                    raise IOError("%s exists but it's not a directory" % (createDir))
+                    raise IOError(
+                        "%s exists but it's not a directory" % (createDir))
```
g1itch: parentheses are redundant here

src/storage/filesystem.py
```diff
@@ -185,23 +205,33 @@ def getMetadata(self, hashId):
...
+                    int(expiresTime),
+                    unhexlify(tag)]
         except IOError:
             raise KeyError
 
     def by_type_and_tag(self, objectType, tag):
         """Get a list of objects filtered by object type and tag"""
         retval = []
-        for stream, streamDict in self._inventory:
+        for _, streamDict in self._inventory:
```
g1itch: `for streamDict in self._inventory.values():`

src/storage/filesystem.py
```diff
@@ -227,7 +258,7 @@ def clean(self):
         """Clean out old items from the inventory"""
         minTime = int(time.time()) - (60 * 60 * 30)
         deletes = []
-        for stream, streamDict in self._inventory.items():
+        for _, streamDict in self._inventory.items():
```
g1itch: `values()`

g1itch:
> @g1itch any more objections?

No, I'd prefer to rewrite the `storage` subpackage completely.

src/storage/filesystem.py
lakshyacis: Thanks

Issue #1575, opened 2019-11-27
addrthread unfinished


PeterSurda:
looks like I forgot to finish it, it doesn't send anything into the network. It needs to be investigated.

g1itch:
So is it resolved after #1576?

PeterSurda:
As far as I know, yes.

g1itch:
11bec55 seems making this actual again

PR #1576, opened 2019-11-27
Addrthread finish

PeterSurda:
- addrthread is supposed to spread addresses as they appear. This was never
  finished during migration to asyncore
- conservative to prevent flood and loops
- randomises order
- see #1575

### DIFF
--- a/src/network/addrthread.py
+++ b/src/network/addrthread.py
@@ -1,6 +1,11 @@
+"""
+Announce addresses as they are received from other hosts
+"""
 import Queue
 
 import state
+from helper_random import randomshuffle
+from network.assemble import assemble_addr
 from network.connectionpool import BMConnectionPool
 from queues import addrQueue
 from threads import StoppableThread
@@ -15,15 +20,26 @@ class AddrThread(StoppableThread):
             while True:
                 try:
                     data = addrQueue.get(False)
-                    chunk.append((data[0], data[1]))
-                    if len(data) > 2:
-                        source = BMConnectionPool().getConnectionByAddr(data[2])
+                    chunk.append(data)
                 except Queue.Empty:
                     break
-                except KeyError:
-                    continue
 
-            # finish
+            if chunk:
+                # Choose peers randomly
+                connections = BMConnectionPool().establishedConnections()
+                randomshuffle(connections)
+                for i in connections:
+                    randomshuffle(chunk)
+                    filtered = []
+                    for stream, peer, seen, destination in chunk:
+                        # peer's own address or address received from peer
+                        if i.destination in (peer, destination):
+                            continue
+                        if stream not in i.streams:
+                            continue
+                        filtered.append((stream, peer, seen))
+                    if filtered:
+                        i.append_write_buf(assemble_addr(filtered))
 
             addrQueue.iterate()
             for i in range(len(chunk)):
--- a/src/network/assemble.py
+++ b/src/network/assemble.py
@@ -1,0 +1,32 @@
+"""
+Create bitmessage protocol command packets
+"""
+
+import struct
+
+import addresses
+from network.constants import MAX_ADDR_COUNT
+from network.node import Peer
+from protocol import CreatePacket, encodeHost
+
+
+def assemble_addr(peerList):
+    """Create address command"""
+    if isinstance(peerList, Peer):
+        peerList = (peerList)
+    if not peerList:
+        return b''
+    retval = b''
+    for i in range(0, len(peerList), MAX_ADDR_COUNT):
+        payload = addresses.encodeVarint(
+            len(peerList[i:i + MAX_ADDR_COUNT]))
+        for stream, peer, timestamp in peerList[i:i + MAX_ADDR_COUNT]:
+            payload += struct.pack(
+                '>Q', timestamp)  # 64-bit time
+            payload += struct.pack('>I', stream)
+            payload += struct.pack(
+                '>q', 1)  # service bit flags offered by this node
+            payload += encodeHost(peer.host)
+            payload += struct.pack('>H', peer.port)  # remote port
+        retval += CreatePacket('addr', payload)
+    return retval
--- a/src/network/bmproto.py
+++ b/src/network/bmproto.py
@@ -19,6 +19,12 @@ import state
 from bmconfigparser import BMConfigParser
 from inventory import Inventory
 from network.advanceddispatcher import AdvancedDispatcher
+from network.constants import (
+    ADDRESS_ALIVE,
+    MAX_MESSAGE_SIZE,
+    MAX_OBJECT_COUNT,
+    MAX_OBJECT_PAYLOAD_SIZE,
+    MAX_TIME_OFFSET)
 from network.dandelion import Dandelion
 from network.bmobject import (
     BMObject, BMObjectInsufficientPOWError, BMObjectInvalidDataError,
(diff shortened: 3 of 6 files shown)


src/network/constants.py
```diff
@@ -0,0 +1,17 @@
+"""
+Network protocol constants
+"""
+
+
+# address is online if online less than this many seconds ago
+ADDRESS_ALIVE = 10800
```
g1itch: Maybe comment with `#:`?

PeterSurda: ok makes sense

g1itch:
So the `AddrThread` should periodically send addr messages to all connections?

src/network/constants.py
PeterSurda: familiarising myself with sphinx ...

g1itch:
Also I don't understand `Multiqueue` ): for example [_init()](../blob/v0.6/src/multiqueue.py#L25) seems unused.

src/network/addrthread.py
```diff
@@ -15,15 +22,26 @@ def run(self):
...
-                    if len(data) > 2:
-                        source = BMConnectionPool().getConnectionByAddr(data[2])
+                    chunk.append(data)
                 except Queue.Empty:
                     break
                 except KeyError:
                     continue
 
-            # finish
+            # Choose peers randomly
```
g1itch: Shouldn't this block be inside the loop?

PeterSurda: I don't think so, it works the same way as `invthread.py`

g1itch: maybe add
```python
if not chunk:
    continue
```

g1itch: it eats 99% CPU

PeterSurda: yea I just noticed that too

PeterSurda:
> So the `AddrThread` should periodically send addr messages to all connections?

kind of, it relays those received from others

PeterSurda:
> Also I don't understand `Multiqueue` ): for example [_init()](../blob/v0.6/src/multiqueue.py#L25) seems unused.

It's inherited from Queue.Queue and called from its own `__init__`: [_init()](https://github.com/python/cpython/blob/2.7/Lib/Queue.py#L28)

PeterSurda:
In other words, with respect to Multiqueue, many calls are called from super, I just reimplement the internal methods. since Queue.Queue only has one `deque()`, whereas Multiqueue has several of them in a list and iterates over them.

src/network/addrthread.py
```diff
@@ -15,15 +22,25 @@ def run(self):
...
+                randomshuffle(connections)
+                for i in connections:
+                    randomshuffle(chunk)
+                    filtered = []
+                    for j in chunk:
+                        if j[2] == i.destination:
+                            continue
+                        if j[0] not in i.streams:
+                            continue
+                        filtered.append(Peer(j[0], j[1], time()))
```
g1itch: But Peer() has two args...

src/network/assemble.py
```diff
@@ -0,0 +1,33 @@
...
+    """Create address command"""
+    if isinstance(peerList, Peer):
+        peerList = (peerList)
+    if not peerList:
+        return b''
+    retval = b''
+    for i in range(0, len(peerList), MAX_ADDR_COUNT):
+        payload = addresses.encodeVarint(
+            len(peerList[i:i + MAX_ADDR_COUNT]))
+        for address in peerList[i:i + MAX_ADDR_COUNT]:
```
g1itch: `for stream, peer, timestamp in peerList[...`?

src/network/addrthread.py
g1itch: maybe `filtered.append(j[0], Peer(*j[2]), time())`?

g1itch:
```patch
diff --git a/src/network/addrthread.py b/src/network/addrthread.py
index 99d8e97c..cba3f454 100644
--- a/src/network/addrthread.py
+++ b/src/network/addrthread.py
@@ -1,14 +1,13 @@
 """
 Announce addresses as they are received from other hosts
 """
+import Queue
 from time import time
 
-import Queue
 import state
 from helper_random import randomshuffle
 from network.assemble import assemble_addr
 from network.connectionpool import BMConnectionPool
-from node import Peer
 from queues import addrQueue
 from threads import StoppableThread
 
@@ -33,12 +32,12 @@ class AddrThread(StoppableThread):
                 for i in connections:
                     randomshuffle(chunk)
                     filtered = []
-                    for j in chunk:
-                        if j[2] == i.destination:
+                    for stream, _, peer in chunk:
+                        if peer == i.destination:
                             continue
-                        if j[0] not in i.streams:
+                        if stream not in i.streams:
                             continue
-                        filtered.append(Peer(j[0], j[1], time()))
+                        filtered.append(stream, peer, time())
                     if filtered:
                         i.append_write_buf(assemble_addr(filtered))
 

```

src/network/bmproto.py
```diff
@@ -461,7 +455,9 @@ def bm_command_addr(self):
                                 "rating": 0,
                                 "self": False,
                             }
-                addrQueue.put((stream, peer, self.destination))
+                    # since we don't track peers outside of knownnodes,
+                    # only spread if in knownnodes to prevent flood
+                    addrQueue.put((stream, peer, self.destination))
```
g1itch: `seenTime` ): it will update the node, maybe:
```patch
diff --git a/src/network/addrthread.py b/src/network/addrthread.py
index fa7a72c6..3f88f9ee 100644
--- a/src/network/addrthread.py
+++ b/src/network/addrthread.py
@@ -32,12 +32,12 @@ class AddrThread(StoppableThread):
                 for i in connections:
                     randomshuffle(chunk)
                     filtered = []
-                    for stream, peer, source in chunk:
+                    for stream, peer, source, seen in chunk:
                         if i.destination in (peer, source):
                             continue
                         if stream not in i.streams:
                             continue
-                        filtered.append((stream, peer, time()))
+                        filtered.append((stream, peer, seen))
                     if filtered:
                         self.logger.warning(
                             'sending addr of len %s', len(filtered))
diff --git a/src/network/bmproto.py b/src/network/bmproto.py
index 0915a5ec..b496e2b0 100644
--- a/src/network/bmproto.py
+++ b/src/network/bmproto.py
@@ -457,7 +457,7 @@ class BMProto(AdvancedDispatcher, ObjectTracker):
                             }
                     # since we don't track peers outside of knownnodes,
                     # only spread if in knownnodes to prevent flood
-                    addrQueue.put((stream, peer, self.destination))
+                    addrQueue.put((stream, peer, self.destination, seenTime))
         return True
 
     def bm_command_portcheck(self):
```

PeterSurda:
Thank you for the review, I made the changes except I put `destination` the last, after `seenTime`, so that the list for `addrQueue` and `assemble_addr` have the first three variables in the same order (`addrQueue` having a fourth at the end).

src/network/addrthread.py
```diff
@@ -15,15 +21,25 @@ def run(self):
...
-            # finish
+            if chunk:
+                # Choose peers randomly
+                connections = BMConnectionPool().establishedConnections()
+                randomshuffle(connections)
+                for i in connections:
+                    randomshuffle(chunk)
+                    filtered = []
+                    for stream, peer, seen, destination in chunk:
+                        if peer in (destination, i.destination):
```
g1itch: I'm not sure. What is it mean if `peer == destination`? The node sent it's own address? But what if `i.destination == destination`?

PeterSurda: `peer == destination` means peer shouldn't receive its own address.
`peer == i.destination` means a peer shouldn't receive an address back that it just sent to me, apart from it being wasteful in general, if both sides used the same algorithm, it would cause infinite loops

PeterSurda: or maybe I didn't do it correctly

PeterSurda: yea the logic is wrong

g1itch: I thought `destination` here is the **source** of `peer`.

src/network/assemble.py
PeterSurda: you're right.

src/network/addrthread.py
PeterSurda: The labels are confusing in this case. Here, `peer` is a variable from the list, while `i.destination` is the remote IP of the connection you're assembling the message for. But in general you are right, the code from the 29f99e6 you commented on was incorrect, I fixed it in a69732f.

src/network/addrthread.py
PeterSurda: yes

src/network/bmproto.py
PeterSurda: yes

src/network/constants.py
PeterSurda: Has this been resolved? I'm not sure I did it correctly.

g1itch:
I merged the branch and trying to connect to at least two nodes. It takes so long last time ):

PeterSurda:
I'm testing it on a public IP just to make sure there are no weird bandwidth spikes. I'll leave it for a couple of days.

PeterSurda:
Testing doesn't reveal any unusual amount of traffic.

Issue #1583, opened 2019-12-07
socket error on sendato


sashamd:
When internet connection drops pybitmessage CPU utilization spikes to 100% and debug.log is quickly filled with 'ERROR - socket error on sendato: [Errno 101] Network is unreachable'.

OS: Ubuntu 16.04.6 LTS
Commit: a69732f060608428ddfebf608bdb3c8751e296ce

navjotcis:
I have tried to check the issue so here when we disconnect the internet then socket connection gives exception so that already handled what's an issue with that? I don't understand the problem here

sashamd:
Python2 doesn't work anymore on my system and PyBitmessage is shelved until better times. Sorry, I can't help with this right now.

navjotcis:
I have tested it again and found some strange behavior because when the user disconnects to the internet the socket connection start giving error so in that case, it is closing socket connection I thought which is wrong because when I again connected with the internet then it is not connecting with the socket so I have commented close connection in udp.py module

- -            if e.errno == 101:
- -                self.announcing = False
- -                self.socket.close()
-              retval = 0
- +            retval = len(self.write_buf)
- +            # if e.errno == 101:
- +            #     self.announcing = False
- +            #     self.socket.close()

and then I have test the whole process again and now it is connecting through the socket after internet connection from a disconnected state

PeterSurda:
Fixed in #1676

PR #1584, opened 2019-12-14
Stop UDPSocket on socket.error 101 (Network is unreachable)

g1itch:
Hello!

This may be a part of the solution for #1583.


g1itch:
There are more related code I don't understand. For example [here](../blob/v0.6/src/network/announcethread.py#L29) `processed` seems to be 0 always.

g1itch:
To test this properly I need a way to disconnect the network from python code.

PR #1590, opened 2019-12-27
curses fixes

lakshyacis:
curses fixes( pylint and flake8 )

### DIFF
--- a/src/bitmessagecurses/__init__.py
+++ b/src/bitmessagecurses/__init__.py
@@ -1,8 +1,6 @@
 """
-src/bitmessagecurses/__init__.py
-================================
+Bitmessage commandline interface
 """
-
 # Copyright (c) 2014 Luke Montalvo <lukemontalvo@gmail.com>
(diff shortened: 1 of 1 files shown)


src/bitmessagecurses/__init__.py
```diff
@@ -1183,10 +1189,10 @@ def loadBlackWhiteList():
 
 
 def runwrapper():
+    """Load messages from database"""
```
PeterSurda: This docstring is wrong, it should be something like "main method"

src/bitmessagecurses/__init__.py
```diff
@@ -1207,26 +1213,27 @@ def runwrapper():
 
 
 def run(stdscr):
+    """Inventory lookup and select color pairs"""
```
PeterSurda: this docstring is also wrong, it also should be something like "main loop".

lakshyacis:
> Minor docstring changes but this isn't really maintained so I'm not sure it's a problem.

Okay

PR #1593, opened 2020-01-15
Formatting and fix License

lakshyacis:
License removed from the docstring and just a minor formatting.


src/build_osx.py
```diff
@@ -13,8 +13,14 @@
     ('bitmsghash', ['bitmsghash/bitmsghash.cl', 'bitmsghash/bitmsghash.so']),
     ('translations', glob('translations/*.qm')),
     ('ui', glob('bitmessageqt/*.ui')),
-    ('translations', glob(str(QtCore.QLibraryInfo.location(QtCore.QLibraryInfo.TranslationsPath)) + '/qt_??.qm')),
-    ('translations', glob(str(QtCore.QLibraryInfo.location(QtCore.QLibraryInfo.TranslationsPath)) + '/qt_??_??.qm')),
+    (
```
PeterSurda: these should be fixed to use `os.path.join`

PR #1596, opened 2020-01-24
Imported packages sequencing and formatting 2

lakshyacis:
fix packages sequencing and formatting .

### DIFF
--- a/src/threads.py
+++ b/src/threads.py
@@ -15,6 +15,12 @@ There are also other threads in the `.network` package.
 
 import threading
 
+from class_addressGenerator import addressGenerator
+from class_objectProcessor import objectProcessor
+from class_singleCleaner import singleCleaner
+from class_singleWorker import singleWorker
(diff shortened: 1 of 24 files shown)


src/threads.py
```diff
@@ -6,9 +6,7 @@
 internal storages (like inventory and knownnodes) and do forced garbage
 collection, `singleWorker` for doing PoW, `sqlThread` for querying sqlite
 database.
-
```
PeterSurda: I'm not sure these changes are needed

src/threads.py
```diff
@@ -6,9 +6,7 @@
...
 There are also other threads in the `.network` package.
-
```
PeterSurda: This one too

src/threads.py
lakshyacis: Okay. I also think this isn't needed. I Will amend the changes that I had done. Thank you

PR #1600, opened 2020-02-05
Don't upload addresses as they appear

### LINKED ISSUE #1598: Excessive memory usage
I got a report about excessive memory usage. At first I thought it's wrongly deployed because I haven't seen it in a long time, but after upgrading to latest code I could trigger it as well. I'm now doing a poor man's bisect to see where it broke.

PeterSurda:
- attempt to fix #1598
- seems to work
- addresses won't be uploaded/announced anymore other than after connecting, Later I need to find out how to announce them without causing problems, but for the time disabling this seems an acceptable drawback

### DIFF
--- a/src/network/bmproto.py
+++ b/src/network/bmproto.py
@@ -31,7 +31,7 @@ from network.dandelion import Dandelion
 from network.proxy import ProxyError
 from node import Node, Peer
 from objectracker import ObjectTracker, missingObjects
-from queues import addrQueue, invQueue, objectProcessorQueue, portCheckerQueue
+from queues import invQueue, objectProcessorQueue, portCheckerQueue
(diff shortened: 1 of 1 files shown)


g1itch:
Why don't you just comment [the line](https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/network/bmproto.py#L469) where the peers are put into `addrQueue`?

PeterSurda:
I suppose that's also an approach.

PeterSurda:
there you go

PeterSurda:
I tested it on one of the bootstrap servers, you probably need many incoming connections and it still takes a couple of hours to trigger, even a day. Then the memory consumption starts growing until it fills up andl swap starts thrashing. Now it's been running for about 2 days without any issues.

PR #1605, opened 2020-02-17
Wine build cleanup and XP fix

PeterSurda:
- spec file was cleaned up
- 32bit build runs on XP (downgrade of PyInstaller needed)

### DIFF
--- a/buildscripts/winbuild.sh
+++ b/buildscripts/winbuild.sh
@@ -19,8 +19,6 @@ function download_sources_32 {
 		https://github.com/Bitmessage/ThirdPartyLibraries/blob/master/PyQt${PYQT_VERSION}-x32.exe?raw=true \
 		https://github.com/Bitmessage/ThirdPartyLibraries/blob/master/Win32OpenSSL-${OPENSSL_VERSION}.exe?raw=true \
 		https://github.com/Bitmessage/ThirdPartyLibraries/blob/master/pyopencl-2015.1-cp27-none-win32.whl?raw=true
-	#wget -P ${SRCPATH} -nc http://www.dll-found.com/zip/m/msvcr120.dll.zip
-	#wget -P ${SRCPATH} -nc http://www.dll-found.com/zip/m/msvcr100.dll.zip
 }
 
 function download_sources_64 {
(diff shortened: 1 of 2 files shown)


g1itch:
THere are some warnings in the debug.log:
```
2020-02-17 11:01:02,002 - ERROR - C PoW test fail.
  File "proofofwork.py", line 306, in init
  File "proofofwork.py", line 117, in _doCPoW
ValueError: Procedure probably called with too many arguments (12 bytes in excess)
```

g1itch:
Also maybe install pywin32: 0fd24c3?

PeterSurda:
> There are some warnings in the debug.log:
This is because on Windows, there are two different types of DLLs with corresponding ways to load them, and it's not easy to find in advance which one the DLL uses and how it was built. So PyBitmessage tests one first and then the second one if the first one fails. If I do it the other way around for the other type of DLL, the load succeeds first but fails later and then it's more difficult to do error handling.

There probably is a better way to do this, but I don't want to devote more resources to it now.

PeterSurda:
> Also maybe install pywin32: [0fd24c3](https://github.com/Bitmessage/PyBitmessage/commit/0fd24c3d72c820bb9570c0ad2f58239b518a0c32)?

My build logs show that pywin32 is a pyinstaller dependency and auto-installs. when you install pyinstaller.

g1itch:
> > Also maybe install pywin32: [0fd24c3](https://github.com/Bitmessage/PyBitmessage/commit/0fd24c3d72c820bb9570c0ad2f58239b518a0c32)?
> 
> My build logs show that pywin32 is a pyinstaller dependency and auto-installs. when you install pyinstaller.

I see this messages in log:
```
INFO: Processing pre-safe import module hook   win32com

  File "<string>", line 2, in <module>

ImportError: No module named win32com

14841 INFO: Processing pre-safe import module hook   win32com

  File "<string>", line 2, in <module>

ImportError: No module named win32com

16550 INFO: Analyzing hidden import 'bitmessageqt.languagebox'

16565 INFO: Analyzing hidden import 'win32com'

16572 INFO: Processing pre-safe import module hook   win32com

  File "<string>", line 2, in <module>

ImportError: No module named win32com

16690 ERROR: Hidden import 'win32com' not found
```
https://travis-ci.org/g1itch/PyBitmessage/jobs/651404444#L3162

g1itch:
BTW: It builds on travis-ci in my branch `ci` now: https://travis-ci.org/g1itch/PyBitmessage/jobs/651429366

PeterSurda:
That looks like a 64bit build, I haven't tried that yet with this PR.

PeterSurda:
I tried the 64bit build now too.I also get the `ImportError` on win32com, but the build completes and the resulting binary works on Win7 64.

g1itch:
> I tried the 64bit build now too.I also get the `ImportError` on win32com, but the build completes and the resulting binary works on Win7 64.

If I remember it right, win32com is used for notification sound. It still may not work because I reimplemented it as plugin...

PeterSurda:
> If I remember it right, win32com is used for notification sound. It still may not work because I reimplemented it as plugin...

The plugins probably don't work at all with this build script, it needs to be fixed later.

PR #1602, opened 2020-02-11
Wine build script update

PeterSurda:
- update and clean up and make sure it works
- it builds the binary but I haven't tried to run the binary
- it probably still missing some fine tuning, OpenCL probably doesn't work

PR #1606, opened 2020-02-19
Build fixes

PeterSurda:
- made the changes you suggested last week (separate `build` and `buildscript` functionality): https://github.com/Bitmessage/PyBitmessage/pull/1602#issuecomment-585128669
- removed obsolete scripts from that directory
- optionally set `pip` CA in wine build. I want to use a HTTPS cache to speed up downloading during builds in my environment, but can't find out how to cleanly set this from the outside as it needs to happen in the middle of running `winebuild.sh`. It only affects build environments that have a specific certificate in `/usr/local/share/ca-certificates/bitmessage-proxy.crt`
- these are separate commits to allow bisect in the future if there is a problem


buildscripts/winbuild.sh
```diff
@@ -67,6 +67,10 @@ function install_python(){
 		echo "Installing vc_redist (2008) for 32 bit "
 		wine vcredist_x86.exe /Q
 	fi
+	# add cert
+	if [ -f /usr/local/share/ca-certificates/bitmessage-proxy.crt ]; then
+		wine python -m pip config set global.cert 'z:\usr\local\share\ca-certificates\bitmessage-proxy.crt'
```
g1itch: I'm afraid that `z:\usr\local\share\ca-certificates\bitmessage-proxy.crt` doesn't translates to `/usr/local/share/ca-certificates/bitmessage-proxy.crt`, but rather in something like `$HOME/.wine64/dosdevices/z\:usr\local\share\ca-certificates\bitmessage-proxy.crt`

g1itch: Though, it may be the same locations.

PeterSurda: it works correctly for me

PR #1608, opened 2020-03-29
Blind signature updates

PeterSurda:
- implemented serialisation and deserialisation, as close to existing standards (openssl, bitcoin) as I could. Specification is here: https://github.com/Bitmessage/PyBitmessage/wiki/Blind-signature-specification
- more tests
- added a chain of blind signatures which is supposed to begin with a CA pubkey


PeterSurda:
The travis issue is due to OpenSSL versions, I tested it on 1.1.0 but it needs changes to be compatible with 1.0.2. I'll work on that.

PeterSurda:
It looks like a padding issue, serializing/deserializing doesn't behave consistently. I'm working on it.

PeterSurda:
The padding issue seems to be resoled, but for some reason it's still failing on travis. On my test VM with xenial the test pass. Continuing investigations.

PR #1609, opened 2020-03-31
Blind Signature Chain implementation

PeterSurda:
- implemented serialisation and deserialisation, as close to existing standards (openssl, bitcoin) as I could. Specification is here: https://github.com/Bitmessage/PyBitmessage/wiki/Blind-signature-specification
- more tests
- added a chain of blind signatures which is supposed to begin with a CA pubkey


PeterSurda:
The travis tests were failing first due to inconsistent serialisation, and then due to class variables (not initialised properly). All is fixed now and tests behave consistently.

PR #1613, opened 2020-04-22
Fix exception in class_sqlThread on Windows

g1itch:
Hello!

This is the fix for the bug on Windows which was recently reported back in chan. When the `messages.dat` file path contains unicode characters the exception `sqlite3.OperationalError: unable to open database file` is raised and PyBitmessage stops loading.

I'm not sure about necessity of the change in `class_sqlThread` (`os.path.join()` instead of strings concatenation). The main problem is in encoding of the `state.appdata` string. I used the solution from https://stackoverflow.com/questions/35833037.


PeterSurda:
Thank you, I think it is a step in the right direction. My testing shows that it still doesn't work. I found two issues.

- there are at least some cases where it doesn't convert correctly:
pybitmessage-encoding2

- an additional fix is needed if the directory with the executable / source contains unicode (as opposed to merely %APPDATA% being localised):
pybitmessage-encoding1

The tests are done on a 64bit Win10 with current updates.

g1itch:
OK, thanks for hint. I've being able to reproduce it only in portable mode with many debug messages added. Can you please tell me how to get such detailed log in console?

PeterSurda:
In the spec file set debug=True and console=True

PeterSurda:
It looks like in my case there is an additional problem, I use a US locale in the OS but created a username with non-US characters. Python doesn't like it and I'm not sure how to fix that other than changing the user locale to something compatible, and relogin. After changing the locale it works even without this PR. Unless someone can report that this actually helped him I'm inclined to reject this PR.

g1itch:
Hmm, I really cannot reproduce the issue with non-ASCII characters in username. What I see is the exception raised when you start PyBitmessage in portable mode from the directory containing non-ASCII characters:
```shell
[7372] LOADER: Post-init sys.path is C:\Users\DEBD1~1.BOG\AppData\Local\Temp\_MEI113682
[7372] LOADER: Setting sys.argv
[7372] LOADER: setting sys._MEIPASS
[7372] LOADER: importing modules from CArchive
[7372] LOADER: extracted struct
[7372] LOADER: callfunction returned...
[7372] LOADER: extracted pyimod01_os_path
[7372] LOADER: callfunction returned...
[7372] LOADER: extracted pyimod02_archive
[7372] LOADER: callfunction returned...
[7372] LOADER: extracted pyimod03_importers
[7372] LOADER: callfunction returned...
[7372] LOADER: Installing PYZ archive with Python modules.
[7372] LOADER: PYZ archive: PYZ-00.pyz
[7372] LOADER: Running pyiboot01_bootstrap.py
[7372] LOADER: Running pyi_rth__tkinter.py
[7372] LOADER: Running pyi_rth_pkgres.py
[7372] LOADER: Running pyi_rth_qt4plugins.py
[7372] LOADER: Running pyi_rth_multiprocessing.py
[7372] LOADER: Running bitmessagemain.py
INFO: PyQt Version: 4.11.4
INFO: Qt Version: 4.8.7
Number of threads: 2
Exception in thread SQL:
  File "threading.py", line 801, in __bootstrap_inner
  File "class_sqlThread.py", line 31, in run
OperationalError: unable to open database file
```

PeterSurda:
Can you tell me how to trigger it?

g1itch:
> Can you tell me how to trigger it?

Put executable into the directory with non-ASCII character in it's name. Run it. Go into Settings and set "User Interface -> Run in Portable mode". Restart PyBitmessage.

PeterSurda (commented):
Ok.

Issue #1607, opened 2020-03-10
[Packaging] Make this application available as flatpak package on FlatHub


Nokia808:
Hi. It will be very useful for Linux users if you make this application available as a flatpak package on FlatHub. Doing so you will make this application available for almost all Linux distors & users in very convenient & secure way ...... Open source applications when added as flatpaks on FlatHub will be in deterministic (reproducible package) & automatically updated & works on almost all Linux distros ... Moreover, flatpak is easier than & more convenient & more secure than AppImage ...... 

Best.

zciendor:
I created a `flatpak-builder` configuration you can use to build a flatpak file yourself. See: https://github.com/Bitmessage/PyBitmessage/pull/1618

Hosting it on Flathub would be the next step, but this requires some actions by the project owner (like signing the package with their GPG key, cross-compiling for multiple architectures, ...).

PR #1618, opened 2020-04-29
flatpak-builder support

zciendor:
Recent distributions like `Debian bullseye` don't support QT4 anymore. This pull request provides a `flatpak-builder` configuration and instructions how to package PyBitmessage as a flatpak application which bundles QT4.

This pull request is also a necessary requirement for https://github.com/Bitmessage/PyBitmessage/issues/1607

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -113,7 +113,7 @@ Once you have `flatpak` and `flatpak-builder` installed:
 git clone git://github.com/Bitmessage/PyBitmessage.git
 cd PyBitmessage/
 git submodule add https://github.com/flathub/shared-modules.git
-flatpak-builder --install build/_flatpak -install-deps-from=flathub --force-clean --user org.bitmessage.PyBitmessage.json
+flatpak-builder --install -install-deps-from=flathub --force-clean --user build/_flatpak org.bitmessage.PyBitmessage.json
 ```
 This will install PyBitmessage to your local flatpak user repository, but it 
 takes a while to compile because QT4 and PyQt4 have to be build, among others.
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -103,26 +103,7 @@ slack.sh - create a package for Slackware
 ```
 
 #### Linux flatpak
-First make sure you have `flatpak` and `flatpak-builder` installed. Follow the
-instructions for your distribution on [flathub](https://flatpak.org/setup/). The
-instructions there only cover the installation of `flatpak`, but 
-`flatpak-builder` should be the same.
-
-Once you have `flatpak` and `flatpak-builder` installed:
-```
-git clone git://github.com/Bitmessage/PyBitmessage.git
-cd PyBitmessage/
-git submodule add https://github.com/flathub/shared-modules.git
-flatpak-builder --install --user -install-deps-from=flathub --force-clean --state-dir=build/.flatpak-builder build/_flatpak org.bitmessage.PyBitmessage.json
-```
-This will install PyBitmessage to your local flatpak user repository, but it 
-takes a while to compile because QT4 and PyQt4 have to be build, among others.
-
-When done you can launch PyBitmessage via:
-`flatpak run -v org.bitmessage.PyBitmessage`
-
-If you want to free up disk space you can remove the `Sdk` runtime again:
-`flatpak uninstall org.freedesktop.Sdk//18.08`
+See [packages/flatpak/README.md](packages/flatpak/README.md)
 
 #### OS X
 Please refer to
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -113,7 +113,7 @@ Once you have `flatpak` and `flatpak-builder` installed:
 git clone git://github.com/Bitmessage/PyBitmessage.git
 cd PyBitmessage/
 git submodule add https://github.com/flathub/shared-modules.git
-flatpak-builder --install -install-deps-from=flathub --force-clean --user build/_flatpak org.bitmessage.PyBitmessage.json
+flatpak-builder --install --user -install-deps-from=flathub --force-clean --state-dir=build/.flatpak-builder build/_flatpak org.bitmessage.PyBitmessage.json
 ```
 This will install PyBitmessage to your local flatpak user repository, but it 
 takes a while to compile because QT4 and PyQt4 have to be build, among others.
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -112,6 +112,7 @@ Once you have `flatpak` and `flatpak-builder` installed:
 ```
 git clone git://github.com/Bitmessage/PyBitmessage.git
 cd PyBitmessage/
+git submodule add https://github.com/flathub/shared-modules.git
 flatpak-builder --install _flatpak -install-deps-from=flathub --force-clean --user org.bitmessage.PyBitmessage.json
 ```
 This will install PyBitmessage to your local flatpak user repository, but it 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -40,6 +40,11 @@ python bitmessagemain.py
 Voilà! Bitmessage is updated!
 
 #### Linux
+_Some recent Linux distributions don't support QT4 anymore, hence PyBitmessage
+won't run with a GUI. However, if you build PyBitmessage as a flatpak, it will
+run in a sandbox which provides QT4. See the **Linux flatpak** instructions 
+in the **Creating a package for installation** section of this document._
+
 To run PyBitmessage from the command-line, you must download the source, then
 run `src/bitmessagemain.py`.
 ```
@@ -96,6 +101,27 @@ puppy.sh - create a package for Puppy Linux
 rpm.sh - create a RPM package
 slack.sh - create a package for Slackware
 ```
+
+#### Linux flatpak
+First make sure you have `flatpak` and `flatpak-builder` installed. Follow the
+instructions for your distribution on [flathub](https://flatpak.org/setup/). The
+instructions there only cover the installation of `flatpak`, but 
+`flatpak-builder` should be the same.
+
+Once you have `flatpak` and `flatpak-builder` installed:
+```
+git clone git://github.com/Bitmessage/PyBitmessage.git
+cd PyBitmessage/
+flatpak-builder --install _flatpak -install-deps-from=flathub --force-clean --user org.bitmessage.PyBitmessage.json
+```
+This will install PyBitmessage to your local flatpak user repository, but it 
+takes a while to compile because QT4 and PyQt4 have to be build, among others.
+
+When done you can launch PyBitmessage via:
+`flatpak run -v org.bitmessage.PyBitmessage`
+
+If you want to free up disk space you can remove the `Sdk` runtime again:
+`flatpak uninstall org.freedesktop.Sdk//18.08`
 
 #### OS X
 Please refer to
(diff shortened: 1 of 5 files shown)


PeterSurda:
Thank you, this is great.

g1itch:
Maybe move the manifest to packages/flatpak?

INSTALL.md
```diff
@@ -97,6 +102,28 @@ rpm.sh - create a RPM package
...
+instructions for your distribution on [flathub](https://flatpak.org/setup/). The
+instructions there only cover the installation of `flatpak`, but 
+`flatpak-builder` should be the same.
+
+Once you have `flatpak` and `flatpak-builder` installed:
+```
+git clone git://github.com/Bitmessage/PyBitmessage.git
+cd PyBitmessage/
+git submodule add https://github.com/flathub/shared-modules.git
+flatpak-builder --install _flatpak -install-deps-from=flathub --force-clean --user org.bitmessage.PyBitmessage.json
```
g1itch, line 116: It's probably a typo here, should be `--install-deps-from=`

zciendor:
> Maybe move the manifest to packages/flatpak?

I'll try... need to change some relative paths within the manifest then.

g1itch:
If we already have all dependencies installed on Ubuntu bionic, is it possible to skip building Qt4?
See https://travis-ci.org/github/g1itch/PyBitmessage/jobs/681467891

g1itch:
> > Maybe move the manifest to packages/flatpak?
> 
> I'll try... need to change some relative paths within the manifest then.

```patch
diff --git a/org.bitmessage.PyBitmessage.json b/org.bitmessage.PyBitmessage.json
index e2bfb73d..c91f5bea 100644
--- a/org.bitmessage.PyBitmessage.json
+++ b/org.bitmessage.PyBitmessage.json
@@ -33,8 +33,8 @@
       ],
       "sources" : [
         {
-          "type" : "dir",
-          "path" : "."
+          "type" : "git",
+          "url" : "https://github.com/Bitmessage/PyBitmessage.git"
         }
       ]
     }
```

zciendor:
> If we already have all dependencies installed on Ubuntu bionic, is it possible to skip building Qt4?

No, Qt4 is not available in any of the available runtimes on flathub: https://docs.flatpak.org/en/latest/available-runtimes.html

Actually, Qt4 is build within the sandbox and not in Ubuntu.

zciendor:
> ```diff
> -          "type" : "dir",
> -          "path" : "."
> +          "type" : "git",
> +          "url" : "https://github.com/Bitmessage/PyBitmessage.git"
> ```

Well, while this works, it would check out the HEAD revision from the remote origin again. This may not be what you want in case you want to build/package the current working copy state.

INSTALL.md
zciendor: > It's probably a typo here, should be `--install-deps-from=`

Fixed. The updated build instructions are now in [packages/flatpak/README.md](https://github.com/Bitmessage/PyBitmessage/pull/1618/files#diff-8421532ac371e215f58a7bbb8a782c8a).

zciendor:
> If we already have all dependencies installed on Ubuntu bionic, is it possible to skip building Qt4?
> See https://travis-ci.org/github/g1itch/PyBitmessage/jobs/681467891

```
Timeout (20 minutes) reached.
```
flatpak-builder has a cache directory. Qt4 won't be rebuild if this exists from a previous build and the inputs for a build step haven't changed. Not sure if this could be configured in travis-ci somehow...

g1itch:
Maybe you can take my [qt5-wip branch](https://github.com/g1itch/PyBitmessage/tree/qt5-wip) and try to build with Qt5?

packages/flatpak/README.md
```diff
@@ -0,0 +1,68 @@
...
+instructions there only cover the installation of `flatpak`, but 
+`flatpak-builder` should be the same.
+
+## Build and Install
+Once you have `flatpak` and `flatpak-builder` installed:
+```
+git clone git://github.com/Bitmessage/PyBitmessage.git
+cd PyBitmessage/
+git submodule update --init --recursive
+flatpak-builder --install --user -install-deps-from=flathub --force-clean --state-dir=build/.flatpak-builder build/_flatpak org.bitmessage.PyBitmessage.json
```
PeterSurda, line 18: `flatpak-builder --install --user -install-deps-from=flathub --force-clean --state-dir=build/.flatpak-builder build/_flatpak packages/flatplak/org.bitmessage.PyBitmessage.json`

packages/flatpak/org.bitmessage.PyBitmessage.json
```diff
@@ -0,0 +1,89 @@
...
+  ],
+  "modules": [
+    "shared-modules/python2.7/python-2.7.json",
+    "shared-modules/qt4/qt4-4.8.7-minimal.json",
+    {
+      "name": "python-sip",
+      "sources": [
+        {
+          "type": "archive",
+          "url": "https://www.riverbankcomputing.com/static/Downloads/sip/4.19.22/sip-4.19.22.tar.gz",
```
PeterSurda, line 21: `          "url": "https://www.riverbankcomputing.com/static/Downloads/sip/4.19.25/sip-4.19.25.tar.gz",`

packages/flatpak/org.bitmessage.PyBitmessage.json
```diff
@@ -0,0 +1,89 @@
...
+          "sha256": "e1b768824ec1a2ee38dd536b6b6b3d06de27b00a2f5f55470d1b512306e3be45"
```
PeterSurda, line 22: `          "sha256": "b39d93e937647807bac23579edbff25fe46d16213f708370072574ab1f1b4211"`

PeterSurda (changes requested):
I tested it and it works fine, now I can use PyBM on Ubuntu 20.04.

For integration into the pipeline I'll have to cache qt and pyqt as they take long to build, but I'll deal with that. Bump the `sip` version and update docs as I requested.

PeterSurda:
Also rebase.

PeterSurda:
It would be great to be able to build the modules separately and upload them to a repo. I don't have time to find out how to do this myself. Can someone do it? Once I have the ability build and publish individual modules, I can make it into an incremental buildbot job.

zciendor:
> Once I have the ability build and publish individual modules, I can make it into an incremental buildbot job.

What do you mean exactly by "publishing" individual modules? If you mean publishing as individual flatpaks I don't think this is possible as flatpaks can only declare dependencies on "runtimes". So you would have to publish and maintain your own "runtime" on flathub, seems a bit overkill to me.

But if you just want to break up the build process that could probably be achieved by splitting the manifest into two or three, four, ... files. You would just have to add simple copy commands to backup the build artifacts to another location outside the build dir and copy them back into the build dir in the final manifest where you build PyBitmessage.

But since this depends highly on your build environment I think you have to do the separation on your own. The `--stop-at` parameter to `flatpak-builder` might help you. You can take the current manifest and stop at the last dependency and inspect the build directory:

`flatpak-builder --user --install-deps-from=flathub --state-dir=build/.flatpak-builder --build-only --stop-at=python-sip build/_flatpak packages/flatpak/org.bitmessage.PyBitmessage.json`.

The only issue is that you can't simply "continue" after using the `--stop-at` option as `flatpak-builder` always wants a clean/empty build directory and aborts if that is not the case.

The solution could be do use different build dirs for each step (per manifest) and add a copy command(s) as the first thing to do in the final `PyBitmessage module` / manifest.

PeterSurda:
Well, I'm not really familiar with flatpak, but I can compare it to what I do with brew for mac build. There I also have to build some dependencies, such as qt and sip, as the binaries aren't always publicly available. What I do is that the build first tries to install the dependency from a URL (private 'repo'), and if it fails, builds it and uploads it to said repo. Next time the build runs and the dependency hasn't changed (no version bump), no need to build it as it's installed from the URL. I was thinking I can do the same with flatpak but it isn't clear to me how to create my own repo. With brew, I can just specify a URL to install from, so any web server will do.

Each build is started from a clean environment.

zciendor:
What about simply persisting the whole `state-dir` (see `--state-dir=build/.flatpak-builder`) across build jobs? Everything is in there. Python and QT4 won't be rebuild if their metadata JSONs didn't change:


```
$ flatpak-builder --user --install-deps-from=flathub --state-dir=build/.flatpak-builder --build-only --stop-at=python-sip build/_flatpak packages/flatpak/org.bitmessage.PyBitmessage.json
Dependency Sdk: org.freedesktop.Sdk 18.08
Updating org.freedesktop.Sdk/x86_64/18.08
Nothing to do.
Dependency Runtime: org.freedesktop.Platform 18.08
Updating org.freedesktop.Platform/x86_64/18.08
Nothing to do.
Downloading sources
Stopping at module python-sip
Starting build of org.bitmessage.PyBitmessage
Cache hit for python-2.7, skipping build
Cache hit for qt4, skipping build
Stopping at module python-sip
Everything cached, checking out from cache
Pruning cache
```

PeterSurda:
> What about simply persisting the whole `state-dir` (see `--state-dir=build/.flatpak-builder`) across build jobs?

This doesn't fit into the design of my setup. All build VMs and containers are ephemeral.

zciendor:
Note that the `state-dir` is not the the `build dir`. `build dir` must be clean and empty before every build. But the purpose of the `state-dir` is to be persisted across builds to achieve all the caching and speedups you are looking for.

zciendor:
Can't you rsync or archive the whole `state-dir` at the end of a build job and upload it to your server/repository and sync it back to the VM at the beginning of a build job?

PeterSurda:
That's basically a kludge version of what I want.

zciendor:
> This doesn't fit into the design of my setup. All build VMs and containers are ephemeral.

But usually you can mount some specific persistent volumes/folders/dirs to a ephermeral VM.

PeterSurda:
Technically that may work, but isn't suitable for easy deployment and maintenance, it creates another item in infrastructure I have to worry about. Compared that to what I do with brew, it just re-uses an existing web server and cache. How about this: https://pagure.io/flatpak-module-tools , doesn't it do what I want?

zciendor:
> How about this: https://pagure.io/flatpak-module-tools , doesn't it do what I want?

As far as I understand this has something to do with "Fedora modules"... not "flatpak modules". It's just using flatpak to build Fedora modules.

The notion of a "module" within the flatpak manifest is nothing which can be exported and deployed as a standalone artifact. Flatpak only has the notion of standalone **and sandboxed** "applications" which have a single dependency on a runtime. So unless you want to become a flatpak "runtime developer" and "runtime maintainer", an "application" is the only thing you can produce. There are no "modules" from a build artifact / repository / deployment perspective. "Modules" only exist from a source code or "build steps" perspective. The source code can be structured into "modules". But "build artifacts" that can be exported are only standalone "applications".

PeterSurda:
Can I do something like a `wget && tar xJf module.tar.xz` with fallback to `flatpak build-module module1 && tar cJf module1.tar.gz && upload`?

zciendor:
Sorry, I'm only familiar with `flatpak-builder` (https://docs.flatpak.org/en/latest/building-introduction.html#flatpak-builder).

PeterSurda:
Ok, how about you just fix the other things I requested (documentation, bump sip, rebase), and the buildbot job will be run manually for now, the splitting into steps I'll deal with separately. This way at least I can have an official flatpak that can be distributed.

g1itch:
> Maybe you can take my [qt5-wip branch](https://github.com/g1itch/PyBitmessage/tree/qt5-wip) and try to build with Qt5?

Oh, python2 now also needs that git submodules. Too bad.

g1itch:
I guess now this should be considered abandoned. Duplicates: #1755.

PeterSurda:
@g1itch yes it's a duplicate, the new one has split dependencies and app, soi that you don't have to build QT4 everytime you are building a new flatpak of the app.

Issue #953, opened 2017-02-24
Move from SHA1 to SHA256


kewde:
Recently the team at Google have found the first SHA1 collision,
the ECDSA signatures use SHA1 and most of the code for a switch to SHA256 is in the comments already.

https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html

PeterSurda:
See here: https://www.reddit.com/r/bitmessage/comments/5vt3la/sha1_and_bitmessage/

- [ ] allow signing with SHA256 for people who want to experiment
- [ ] make all signatures SHA256
- [ ] disable support for SHA1 verification

martinvahi:
A few related links:

http://www.shattered.io/

https://www.fossil-scm.org/fossil/doc/trunk/www/hashpolicy.wiki

g1itch:
Wouldn't this potentially make it possible to use the bitcoin ASICs to spam bitmessage?

kewde:
@g1itch I doubt it. 
The ASICs operate under a very specific format that I doubt is applicable to BitMessage.

PeterSurda:
@g1itch Bitmessage uses double SHA512 for PoW, so no. The SHA1 -> SHA256 migration is only for sender authentication.

kewde:
Has there been any progress on this issue?

martinvahi:
> Wouldn't this potentially make it possible to use the bitcoin ASICs to spam bitmessage?

> The ASICs operate under a very specific format that I doubt is applicable to BitMessage.

If the ASIC's are implemented by using FPGAs, which might be the case to allow the same hardware, server park, to be reconfigured and reused for mining other cryptocoins after the Bitcoin "mine" has become "depleted enough", then the switch from one hash algorithm to another is not that big of an impediment for the server park owners.

PeterSurda:
@kewde You can specify that you want to send SHA256-hashed messages by specifying

    digestalg = sha256

in the bitmessagemain section of keys.dat. The other steps outlined will progress as new releases are made.

PeterSurda:
This probably should be expedited, it's been waiting for too long. I've been running with `digestalg = sha256` for a long time and haven't had issues.

mrpeertopeer:
It should be `sha512`.

PR #1623, opened 2020-05-05
Moving code from shared, easy part

g1itch:
Hello!

In this PR I moved most of the variables defined in `shared` to `state` as mentioned in the module's docstring. I also changed default digest algo from SHA1 to SHA256 to move #953 close to the second step.


PeterSurda:
Thanks I'll look at it. The SHA is probably ok, I've been using `digestalg=sha256` for a long time and haven't had problems.

PR #1647, opened 2020-06-05
fixed CQ for bitmessageqt.migrationwizard module

coolguy-cell:
flake8 fix
pycode fix
pylint fix


coolguy-cell:
This file should be removed completely.

PR #1651, opened 2020-06-09
fixed CQ for bitmessageqt.safehtmlparser module

coolguy-cell:
flake8 fixes
pycodestyle fixes
pylint fixes


src/bitmessageqt/safehtmlparser.py
```diff
@@ -1,4 +1,5 @@
 """Subclass of HTMLParser.HTMLParser for MessageView widget"""
+# pylint: disable=attribute-defined-outside-init
```
PeterSurda, line 2: fix the code rather than override pylint

Issue #1640, opened 2020-06-01
Nodes don't get added into knownnodes


PeterSurda:
Upon looking through the code, I noticed that nodes of incoming connections aren't added to `knownnodes`. Pre-asyncore, if you received an incoming connection, you added its data into knownnodes. It appears I removed it during refactoring and never re-added it. It means that knownnodes gets filled with obsolete crap. It may also be contributing to #1597.

The code should call `knownnodes.addKnownNode` for each incoming stream during `set_connection_fully_established` for inbound connections.

g1itch:
Incoming connections usually report `127.0.0.1`: https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/protocol.py#L300

PeterSurda:
Yes, but you see their IP in `self.destination` and this you can add to `knownnodes`. This is how it used to work in the past. Maybe only bootstrap servers should do this, after checking that the node is accessible from the outside.

PeterSurda:
I think #1654 resolved this.

PR #1654, opened 2020-06-09
Conservative optimizations in knownnodes processing

g1itch:
Hello!

It seems that minimal solution for #1640 is else branch in [TCPConnection.handle_read()](../blob/v0.6/src/network/tcp.py#L257) but it decreases performance. I see the only reason to do it in `.handle_read()`: to update node's `lastseen` field. So I moved that procedure into `.state_connection_fully_established()`. `lastseen` may be updated in `.handle_close()` I guess.

I also took the commits those don't introduce new entities from my `knownnodes-opt` branch, mainly related to isolation of knownnodes processing in the `knownnodes` module.


src/network/knownnodes.py
```diff
@@ -95,12 +99,44 @@ def saveKnownNodes(dirName=None):
...
 def addKnownNode(stream, peer, lastseen=None, is_self=False):
-    """Add a new node to the dict"""
+    """
+    Add a new node to the dict or update lastseen if it already exists.
+    Do it for each stream number if *stream* is `Iterable`.
+    Returns True if added a new node.
+    """
+    # pylint: disable=too-many-branches
+    if isinstance(stream, Iterable):
+        with knownNodesLock:
```
PeterSurda, line 109: Since now we're now using `knownNodesLock` in a reentrant fashion, we should redefine it as a `threading.RLock` to avoid deadlocks.

g1itch: Done. Unfortunately threads is not my thing.

g1itch:
OK, so I merge. Should I add "closes: #1640" to 8dfd9e3?

PR #1664, opened 2020-08-31
Finish timeOffsetWrongCount

### LINKED ISSUE #1433: BMProto.timeOffsetWrongCount unfinished
Forgot to finish it after moving to asyncore. There should be a check for a threshold, and a panel notification.

PeterSurda:
- now notifies in status bar
- fixes #1433

### DIFF
--- a/src/network/tcp.py
+++ b/src/network/tcp.py
@@ -2,6 +2,7 @@
 TCP protocol handler
 """
 # pylint: disable=too-many-ancestors
+import l10n
 import logging
 import math
 import random
@@ -29,11 +30,13 @@ from network.socks5 import Socks5Connection
 from network.tls import TLSDispatcher
 from node import Peer
 from queues import invQueue, receiveDataQueue, UISignalQueue
+from tr import _translate
 
 logger = logging.getLogger('default')
 
 
 maximumAgeOfNodesThatIAdvertiseToOthers = 10800  #: Equals three hours
+maximumTimeOffsetWrongCount = 3  #: Connections with wrong time offset
 
 
 class TCPConnection(BMProto, TLSDispatcher):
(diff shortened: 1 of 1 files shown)


src/network/tcp.py
```diff
@@ -124,6 +127,23 @@ def antiIntersectionDelay(self, initial=False):
...
+                maximumTimeOffsetWrongCount and \
+                not self.fullyEstablished:
+            UISignalQueue.put((
+                'updateStatusBar',
+                # pylint: disable=protected-access
+                tr._translate(
+                    "MainWindow",
+                    "The time on your computer, %1, may be wrong. "
+                    "Please verify your settings."
+                ).arg(datetime.datetime.now().strftime("%H:%M:%S"))))
```
g1itch, line 145: Maybe use `l10n.formatTimestamp()` here, like in [class_singleWorker](../../blob/v0.6/src/class_singleWorker.py#L677)?

src/network/tcp.py
```diff
@@ -124,6 +127,23 @@ def antiIntersectionDelay(self, initial=False):
...
+        """
+        Check if we have connected to too many nodes which have too high
+        time offset from us
+        """
+        if BMProto.timeOffsetWrongCount > \
+                maximumTimeOffsetWrongCount and \
+                not self.fullyEstablished:
+            UISignalQueue.put((
+                'updateStatusBar',
+                # pylint: disable=protected-access
```
g1itch, line 140: No need. Just `from tr import _translate`.

PeterSurda: ok

src/network/tcp.py
PeterSurda: ok

g1itch (approved):
Looks good.

PR #1665, opened 2020-09-05
Offline mode network settings fix

### LINKED ISSUE #1656: Offline mode first time network settings
Reported in bitmessage chan:

> I am reporting a bug.
> To reproduce:
>
> 1. Backup $HOME/.config/PyBitmessage and remove the directory
> 2. Start bitmessagemain.py
> 3. When dialog popped, select "Work offline" and push "OK"
> 4. On "Settings" menu, go "Network Settings"
>   and select "Proxy server / Tor" "Type:" "SOCKS5",
>   then push "OK"
>   (In this case Tor is required, but I guess any changes may work)
> 5. Select "Network Status" tab, and wait few minutes
> 6. The client will be automatically connected to some nodes
>
> Tested on latest git version: commit e237534335d66748b27fb6251faf6cb72912ebd1

I'm not sure if this is just unintuitive UI, or there is a check to see if you're changing network settings for the first time which mis-checks.

navjotcis:
- fixes #1656

### DIFF
--- a/src/bitmessageqt/settings.py
+++ b/src/bitmessageqt/settings.py
@@ -49,6 +49,8 @@ class SettingsDialog(QtGui.QDialog):
         self.net_restart_needed = False
         self.timer = QtCore.QTimer()
 
+        if self.config.safeGetBoolean('bitmessagesettings', 'dontconnect'):
+            self.firstrun = False
         try:
             import pkg_resources
         except ImportError:
--- a/src/bitmessageqt/settings.py
+++ b/src/bitmessageqt/settings.py
@@ -342,6 +342,8 @@ class SettingsDialog(QtGui.QDialog):
         if proxytype_index == 0:
             if self._proxy_type and state.statusIconColor != 'red':
                 self.net_restart_needed = True
+        elif state.statusIconColor == 'red' and self.config.safeGetBoolean('bitmessagesettings', 'dontconnect'):
+            self.net_restart_needed = False
         elif self.comboBoxProxyType.currentText() != self._proxy_type:
             self.net_restart_needed = True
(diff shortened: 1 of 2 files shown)


src/bitmessageqt/settings.py
```diff
@@ -342,6 +342,8 @@ def accept(self):
         if proxytype_index == 0:
             if self._proxy_type and state.statusIconColor != 'red':
                 self.net_restart_needed = True
+        elif state.statusIconColor == 'red' and self.config.safeGetBoolean('bitmessagesettings', 'dontconnect'):
```
PeterSurda, line 345: I would drop the color test, and put the `dontconnect` test as the first of the tests (if, not elif).

PeterSurda (commented):
I want feedback from @g1itch

src/bitmessageqt/settings.py
g1itch: I'm not sure the bug is here. What about test? I think I had a fix for #1656 issue when it was reported. But I preferred first to  write a test in order to prevent reintroducing it.

g1itch (changes requested):
Please, follow the rule: one issue - one PR.

navjotcis:
here I have done some tests after fixing the network connection issue in the case when the network is set offline
image

PeterSurda (approved):
I was able to reproduce the bug reported, and confirm that this fixes it.

g1itch (approved):
Excuse me, I was not going to block this PR, it was a comment.

Manual check looks OK: "Start -> Work Offline -> Settings -> select Proxy Server/Tor -> OK: not connecting".

PR #1666, opened 2020-09-08
Own address should not save in address book

### LINKED ISSUE #1485: own addresses cannot be added to addressbook
here when I tried adding my own address to address book,it adds that without any error and I think it should not be like that.

navjotcis:
1) Here I have worked on fixing own address saving issue in the address book so for that, I have added one more method in 
shared.py module for checking that the entered address is present in the config parser file or not then on the bases of that I am showing a validation message on the QT Ui.
2) Fixes #1485


g1itch:
>     I have added one more method in
>        shared.py module for checking that the entered address is present in the config parser file or not then on the bases of that I am showing a validation message on the QT Ui.

You should not add any new code into deprecated module `shared`, especially one-liners.

**note:**
 * shared.py is a file, the module is called `shared`
 * `keys.dat` is not a "config parser file", it's a configuration file
 * what you've added is not "validation message", the validation message is shown in red in the "New Chan" dialog

src/bitmessageqt/__init__.py
```diff
@@ -2470,6 +2470,13 @@ def click_pushButtonAddAddressBook(self, dialog=None):
             ))
             return
 
+        if shared.isAddressInConfigParserList(address):
```
g1itch, line 2473:
```python
if address in BMConfigParser().addresses():
```

navjotcis:
I have tried to reproduce the issue but it is fixed because wherever I am trying to save my own address on the address book it is not allowing me to save it

g1itch:
@navjotcis Did you log on travis-ci? See. https://travis-ci.community/t/some-pr-are-not-analyzed-with-message-abuse-detected/1191

PeterSurda:
I would create a new function for adding an address into addressbook, and put it into a new file `helper_addressbook.py`. Such a test shouldn't be in the frontend code.

PeterSurda:
What I wrote previously hasn't been done yet.

src/bitmessageqt/__init__.py
```diff
@@ -2435,12 +2436,14 @@ def click_pushButtonAddAddressBook(self, dialog=None):
             ))
             return
 
-        self.addEntryToAddressBook(address, label)
-
-    def addEntryToAddressBook(self, address, label):
-        if shared.isAddressInMyAddressBook(address):
+        if address in BMConfigParser().addresses():
```
PeterSurda, line 2439: This needs to be moved into the helper function. The error message can be done through UISignaller.

src/tests/test_addressbook.py
```diff
@@ -0,0 +1,20 @@
...
+
+import unittest
+from pybitmessage.bmconfigparser import BMConfigParser
+
+
+class TestAddAddressBook(unittest.TestCase):
+    """this class is used for testting add address feature"""
+
+    def test_is_own_address_add_to_addressbook(self):
+        """Check the logic of TCPConnection.local"""
```
PeterSurda, line 13: fix docstring

src/helper_addressbook.py
```diff
@@ -0,0 +1,11 @@
...
+Insert value into addressbook
+"""
+
+from helper_sql import sqlExecute
+
+
+def insert(address, label):
+    """perform insert into addressbook"""
+
+    sqlExecute('''INSERT INTO addressbook VALUES (?,?)''', label, address)
```
PeterSurda, line 11: it should pass the return value, or True/False depending on whether it's 0.

src/tests/test_addressbook.py
```diff
@@ -0,0 +1,20 @@
...
+        own_addresses = BMConfigParser().addresses()
```
PeterSurda, line 14: call the helper function twice with the same address, first it should succeed, then fail.

PeterSurda (changes requested):

src/helper_addressbook.py
```diff
@@ -0,0 +1,16 @@
...
+from bmconfigparser import BMConfigParser
+from helper_sql import sqlExecute
+
+
+def insert(address, label):
+    """perform insert into addressbook"""
+
+    is_valid = False
+    if address not in BMConfigParser().addresses():
+        sqlExecute('''INSERT INTO addressbook VALUES (?,?)''', label, address)
```
PeterSurda, line 14: you need to verify the reqturn value of `sqlExecute`, which is the number of modified rows

src/tests/test_addressbook.py
```diff
@@ -0,0 +1,25 @@
...
+from pybitmessage.bmconfigparser import BMConfigParser
+
+
+class TestAddAddressBook(unittest.TestCase):
+    """this class is used for testting add address feature"""
+
+    def test_is_own_address_add_to_addressbook(self):
+        """Checking own address adding in addressbook"""
+
+        all_addresses = BMConfigParser().addresses()
```
PeterSurda, line 17:
```
assertTrue(insert(blah, blah))
assertFalse(insert(blah, blah)) # will auto-detect duplicate and fail
```

PeterSurda (changes requested):

src/helper_addressbook.py
```diff
@@ -0,0 +1,17 @@
...
+from bmconfigparser import BMConfigParser
+from helper_sql import sqlExecute
+
+
+def insert(address, label):
+    """perform insert into addressbook"""
+
+    is_valid = False
+    if address not in BMConfigParser().addresses():
+        result = sqlExecute('''INSERT INTO addressbook VALUES (?,?)''', label, address)
```
PeterSurda, line 14: `return sqlExecute(.....) == 1`

src/helper_addressbook.py
```diff
@@ -0,0 +1,17 @@
...
+        if result == 1:
+            is_valid = True
+    return is_valid
```
PeterSurda, line 17: `return False`

PeterSurda (changes requested):
Please make the changes we discussed:
- simplify the code
- two test cases, one for inserting your own address, one for inserting the same fixed address twice (first insert should succeed, second should fail)

src/tests/test_addressbook.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class TestAddAddressBook(unittest.TestCase):
+    """this class is used for testting add address feature"""
+
+    addr = 'BM-2cVvkzJuQDsQHLqxRXc6HZGPLZnkBLzEZY'
+
+    def test_add_same_address_twich_in_addressbook(self):
+        """checking same address is added twice in addressbook"""
+        if helper_addressbook.insert(label='test1', address=self.addr):
```
PeterSurda, line 18: this isn't a correct test. It should always add it twice, like I said. No `if`s.
```
assertTrue(insert(label='test1', address=self.addr))`
assertFalse(insert(label='test1', address=self.addr)) # will auto-detect duplicate and fail

File: src/tests/test_addressbook.py [hunk 11]
```diff
@@ -0,0 +1,31 @@
...
+            self.assertTrue(helper_addressbook.insert(label='test1', address=self.addr))
+
+    def test_is_address_present_in_addressbook(self):
+        """checking is address added in addressbook or not"""
+        self.assertTrue(self.addr in BMConfigParser().addresses())
```
PeterSurda, line 23: this also doesn't test anything useful.

src/tests/test_addressbook.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def test_is_own_address_add_to_addressbook(self):
+        """Checking own address adding in addressbook"""
```
PeterSurda, line 26: also no useful test.

PeterSurda (changes requested):
Please make the test do what I asked for.

PeterSurda:
Test 1: generate a random address and try to add it into addressbook. Adding should fail.
Test 2: pick a valid fixed address, add the same address into addressbook twice. First add should succeed, second fail
Test 3: pick a valid fixed address, add it into addressbook, and make a SQL query to verify it's in the table

Since these tests operate with sql, they need to be run as core tests.

src/helper_addressbook.py
```diff
@@ -0,0 +1,16 @@
...
+Insert value into addressbook
+"""
+
+from bmconfigparser import BMConfigParser
+from helper_sql import sqlExecute, sqlQuery
+
+
+def insert(address, label):
+    """perform insert into addressbook"""
+    queryreturn = sqlQuery(
```
PeterSurda, line 11: This check isn't needed.

src/tests/core.py
```diff
@@ -274,6 +276,33 @@ def test_old_knownnodes_pickle(self):
...
+    def generate_random_address():
+        import queues
+        streamNumberForAddress = 1
+        queues.addressGeneratorQueue.put((
+            'createRandomAddress', 4, streamNumberForAddress,
+            "test1", 1, "", False))
+
+    def test_add_same_address_twich_in_addressbook(self):
+        """checking same address is added twice in addressbook"""
+        helper_addressbook.insert(label='test1', address=self.addr)
```
PeterSurda, line 289: `assertTrue`

src/tests/core.py
```diff
@@ -274,6 +276,33 @@ def test_old_knownnodes_pickle(self):
...
+        self.assertTrue(helper_addressbook.insert(label='test1', address=self.addr))
```
PeterSurda, line 290: `assertFalse`

src/tests/core.py
```diff
@@ -274,6 +276,33 @@ def test_old_knownnodes_pickle(self):
...
 
+    @staticmethod
+    def generate_random_address():
+        import queues
+        streamNumberForAddress = 1
+        queues.addressGeneratorQueue.put((
+            'createRandomAddress', 4, streamNumberForAddress,
+            "test1", 1, "", False))
+
+    def test_add_same_address_twich_in_addressbook(self):
```
PeterSurda, line 287: `twice`

src/tests/core.py
```diff
@@ -274,6 +276,33 @@ def test_old_knownnodes_pickle(self):
...
+        """checking same address is added twice in addressbook"""
+        helper_addressbook.insert(label='test1', address=self.addr)
+        self.assertTrue(helper_addressbook.insert(label='test1', address=self.addr))
+
+    def test_is_address_present_in_addressbook(self):
+        """checking is address added in addressbook or not"""
+        helper_addressbook.insert(label='test1', address=self.addr)
+        queryreturn = sqlQuery(
+            '''select count(*) from addressbook where address=?''', self.addr)
+        self.assertTrue(True if queryreturn[0][0] else False)
```
PeterSurda, line 297: `self.assertTrue(bool(queryreturn[0][0]))`

src/tests/core.py
```diff
@@ -274,6 +276,33 @@ def test_old_knownnodes_pickle(self):
...
+
+    def test_is_own_address_add_to_addressbook(self):
+        """Checking own address adding in addressbook"""
+        self.generate_random_address()
+        all_addresses = BMConfigParser().addresses()
+        if all_addresses:
```
PeterSurda, line 303:
```
try:
    self.assertFalse((helper_addressbook.insert(label='test', address=all_addresses[0]))
except KeyError:
    self.fail("Can't generate addresses")

File: src/tests/core.py [hunk 19]
```diff
@@ -274,6 +276,33 @@ def test_old_knownnodes_pickle(self):
...
+            "test1", 1, "", False))
+
+    def test_add_same_address_twich_in_addressbook(self):
+        """checking same address is added twice in addressbook"""
+        helper_addressbook.insert(label='test1', address=self.addr)
+        self.assertTrue(helper_addressbook.insert(label='test1', address=self.addr))
+
+    def test_is_address_present_in_addressbook(self):
+        """checking is address added in addressbook or not"""
+        helper_addressbook.insert(label='test1', address=self.addr)
```
PeterSurda, line 294: if this is run before `test_add_same_address_twich_in_addressbook`, that one will fail because it's duplicate already. Use a different address, or have both test cases delete the address after they are done.

PeterSurda (commented):

src/tests/core.py
navjotcis: ok

src/helper_addressbook.py
navjotcis: I have put these checks because sometimes BMConfigParser().addresses() gives an empty list even if there is an address saved in addressbook so in that case we need to check it throught sql query for duplicate addresss

PeterSurda: > ... sometimes ...

The tests must be as deterministic as possible. They must always succeed. If they sometimes fail, it needs to be fixed properly rather than skipped.

`BMConfigParser().addresses()` are your own addresses, whereas addressbook is a separate list. Maybe your confusing the two?

navjotcis: No, I am not confusing actually the problem is when you add the same address twice in Addressbook
image
then according to the above method first case will pass but when you again enter the same address in Aaddressbook then it is not failing it is also passing because here
image
in insert method BMConfigParser().addresses() always gives empty string even data is present in Addressbook so due to that I have added extra check for checking the address in addressbook

navjotcis: do you have any idea why BMConfigParser().addresses() is displaying empty list?

PeterSurda: So I checked and I mistakenly believed the `addressbook` table has a constraint. It doesn't. So it needs to be fixed.

`BMConfigParser().addresses()` is empty until a new address is generated.

navjotcis:
added Unique constraint in address column of addressbook for preventing duplicacy

src/tests/core.py
```diff
@@ -274,6 +276,42 @@ def test_old_knownnodes_pickle(self):
...
+    def generate_random_address():
+        """Generating random address"""
+        import queues
+        streamNumberForAddress = 1
+        queues.addressGeneratorQueue.put((
+            'createRandomAddress', 4, streamNumberForAddress,
+            "test1", 1, "", False))
+
+    def delete_address_from_database(self):
+        """Deleting random address"""
```
PeterSurda, line 289: `"""Clean up addressbook"""`

src/class_sqlThread.py
```diff
@@ -388,7 +388,27 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
             logger.debug(
                 'In messages.dat database, done adding address field to the pubkeys table'
                 ' and removing the hash field.')
-            self.cur.execute('''update settings set value=10 WHERE key='version';''')
+            item = '''update settings set value=? WHERE key='version';'''
```
PeterSurda, line 391: why can't you keep it like it was and just bump version?

src/class_sqlThread.py
```diff
@@ -388,7 +388,27 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+            logger.debug(
+                'In messages.dat database, updating address column to UNIQUE'
+                ' in the addressbook table.')
+            self.cur.execute(
+                '''ALTER TABLE addressbook RENAME TO old_addressbook''')
+            self.cur.execute(
+                '''CREATE TABLE addressbook'''
+                ''' (label text, address text, UNIQUE(address) ON CONFLICT REPLACE)''')
+            self.cur.execute(
+                '''INSERT INTO addressbook SELECT label, address FROM old_addressbook;''')
```
PeterSurda, line 410: add
`'''DROP TABLE old_addressbook'''`

src/class_sqlThread.py
```diff
@@ -47,7 +47,7 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
             self.cur.execute(
                 '''CREATE TABLE subscriptions (label text, address text, enabled bool)''')
             self.cur.execute(
-                '''CREATE TABLE addressbook (label text, address text)''')
+                '''CREATE TABLE addressbook (label text, address text, UNIQUE(address) ON CONFLICT REPLACE)''')
```
PeterSurda, line 50: it needs to be `ON CONFLICT IGNORE`, otherwise the checks for row count won't work correclty.

src/class_sqlThread.py
```diff
@@ -388,7 +388,27 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+        currentVersion = int(self.cur.fetchall()[0][0])
+        if currentVersion == 10:
+            logger.debug(
+                'In messages.dat database, updating address column to UNIQUE'
+                ' in the addressbook table.')
+            self.cur.execute(
+                '''ALTER TABLE addressbook RENAME TO old_addressbook''')
+            self.cur.execute(
+                '''CREATE TABLE addressbook'''
+                ''' (label text, address text, UNIQUE(address) ON CONFLICT REPLACE)''')
```
PeterSurda, line 408: needs to be `ON CONFLICT IGNORE`

src/tests/core.py
```diff
@@ -274,6 +276,52 @@ def test_old_knownnodes_pickle(self):
...
             cleanup(files=('knownnodes.dat',))
 
+    @staticmethod
+    def generate_random_address():
+        """Generating random address"""
+        import queues
+        streamNumberForAddress = 1
+        queues.addressGeneratorQueue.put((
+            'createRandomAddress', 4, streamNumberForAddress,
+            "test1", 1, "", False))
```
PeterSurda, line 286: you should also wait for the address to be generated Look at https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/bitmessageqt/support.py#L87 to see how to do it.

src/tests/core.py
```diff
@@ -274,6 +276,52 @@ def test_old_knownnodes_pickle(self):
...
+
+    @staticmethod
+    def delete_address_from_database(address):
```
PeterSurda, line 289: this method should be called `delete_address_from_addressbook` to avoid conflusion

src/tests/core.py
```diff
@@ -274,6 +276,52 @@ def test_old_knownnodes_pickle(self):
...
+        """checking is address added in addressbook or not"""
+        helper_addressbook.insert(label='test1', address=self.addr)
+        queryreturn = sqlQuery('''select count(*) from addressbook where address=?''', self.addr)
+        self.assertTrue(bool(queryreturn[0][0]))
+        self.delete_address_from_database(self.addr)
+
+    def test_is_own_address_add_to_addressbook(self):
+        """Checking own address adding in addressbook"""
+        self.generate_random_address()
+        time.sleep(.5)
```
PeterSurda, line 309: instead of this sleep look at how it's done in `support.py` and put the waiting into `generate_random_address` directly

PeterSurda (changes requested):
please make the requested changes

src/tests/core.py
```diff
@@ -274,6 +276,42 @@ def test_old_knownnodes_pickle(self):
...
+        """checking same address is added twice in addressbook"""
+        self.assertTrue(helper_addressbook.insert(label='test1', address=self.addr))
+        self.assertFalse(helper_addressbook.insert(label='test1', address=self.addr))
+        self.delete_address_from_addressbook(self.addr)
+
+    def test_is_address_present_in_addressbook(self):
+        """checking is address added in addressbook or not"""
+        helper_addressbook.insert(label='test1', address=self.addr)
+        queryreturn = sqlQuery('''select count(*) from addressbook where address=?''', self.addr)
+        self.assertTrue(bool(queryreturn[0][0]))
```
PeterSurda, line 294: I would do `assertEqual(return, 1)` in this case.

PeterSurda (changes requested):

src/tests/core.py
navjotcis: here we are checking address is present in addressbook or not if anyhow count will show more then one it means the address is stored multiple times then in this case test will fail for assertEqual(return, 1) and it will pass for 
self.assertTrue(bool(queryreturn[0][0])) so i think we should keep it

PeterSurda (approved):

PR #1667, opened 2020-09-11
fixed socket error on sendato

navjotcis:
here when the internet connection is disconnected then the socket connection starts giving an exception so in the previous code socket was closed when socket.error produces but it creates an issue because when the user again connects with the internet then it is not connecting with sockets so it causes and issues so I have commented socket close line then it starts working fine.

### DIFF
--- a/src/network/udp.py
+++ b/src/network/udp.py
@@ -145,8 +145,5 @@ class UDPSocket(BMProto):  # pylint: disable=too-many-instance-attributes
                 self.write_buf, ('<broadcast>', self.port))
         except socket.error as e:
             logger.error("socket error on sendto: %s", e)
-            if e.errno == 101:
-                self.announcing = False
-                self.socket.close()
-            retval = 0
+            retval = len(self.write_buf)
         self.slice_write_buf(retval)


src/network/udp.py
```diff
@@ -145,8 +145,9 @@ def handle_write(self):
                 self.write_buf, ('<broadcast>', self.port))
         except socket.error as e:
             logger.error("socket error on sendto: %s", e)
-            if e.errno == 101:
-                self.announcing = False
-                self.socket.close()
             retval = 0
```
PeterSurda, line 148: remove line 148

src/network/udp.py
```diff
@@ -145,8 +145,9 @@ def handle_write(self):
...
+            retval = len(self.write_buf)
+            # if e.errno == 101:
```
PeterSurda, line 150: remove lines 150 through 152

PeterSurda (changes requested):

src/network/udp.py
PeterSurda: Still not done (line 148 in new, 151 in old)

PeterSurda (changes requested):

navjotcis:
I have tried to reproduce the issue but I did not able to reproduce it.

src/network/udp.py
navjotcis: removed line 148

PeterSurda (approved):

PeterSurda (approved):

PeterSurda:
I refactored the use of `state.appdata` and removed the environment variable from `.travis.yml`. It now passes buildbot's bionic and xenial, as well as xenian on travis-ci.org. Can you try rebasing your changes on top of it again?

Issue #1662, opened 2020-08-27
Pickle old knownnodes missing/incomplete error handler


PeterSurda:
This happened during upgrade:

```
Traceback (most recent call last):
  File "/usr/src/PyBitmessage/src/bitmessagemain.py", line 485, in <module>
    main()
  File "/usr/src/PyBitmessage/src/bitmessagemain.py", line 481, in main
    mainprogram.start()
  File "/usr/src/PyBitmessage/src/bitmessagemain.py", line 238, in start
    readKnownNodes()
  File "/usr/src/PyBitmessage/src/network/knownnodes.py", line 158, in readKnownNodes
    pickle_deserialize_old_knownnodes(source)
  File "/usr/src/PyBitmessage/src/network/knownnodes.py", line 84, in pickle_deserialize_old_knownnodes
    knownNodes = pickle.load(source)
  File "/usr/lib/python2.7/pickle.py", line 1378, in load
    return Unpickler(file).load()
  File "/usr/lib/python2.7/pickle.py", line 858, in load
    dispatch[key](self)
  File "/usr/lib/python2.7/pickle.py", line 1090, in load_global
    klass = self.find_class(module, name)
  File "/usr/lib/python2.7/pickle.py", line 1126, in find_class
    klass = getattr(mod, name)
AttributeError: 'module' object has no attribute 'Peer'
```

maybe the `pickle.load` at line 84 of knownnodes.py should be wrapped in try/except.

g1itch:
Oops. d56191e was probably caused new bug

g1itch:
I also suspect that I broke the process of upgrade from versions before 0.6.2 in other places. The upgrade process should probably have it's own test suite.

PeterSurda:
fixed in 9fe4ad0489ca42801e06abf05333a42213c3c9c8

PR #1661, opened 2020-07-20
Wrong arguments order in bitmessageqt.MyForm.updateStatusBar()

g1itch:
Hello!

I found that `proofofwork.notifyBuild()` puts arguments into `UISignalQueue` in the wrong order. I added a pseudo test so you can see error message, but to write a proper test I need some special technique, maybe involving `excQueue`.

PR #1668, opened 2020-09-14
Fix backward compatibility in pickle_deserialize_old_knownnodes()

g1itch:
Hello!

#1662 is caused by changed package structure.

Here I've set up a minimal upgrade from v0.6.3 to reproduce the bug. Using v0.6.2 would be difficult, because it has no command line args.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,12 @@ addons:
       - xvfb
 install:
   - pip install -r requirements.txt
-  - ln -s src pybitmessage  # tests environment
   - python setup.py install
 script:
   - python checkdeps.py
+  - git clone -b v0.6.3 https://github.com/Bitmessage/PyBitmessage.git old;
+    cd old; src/bitmessagemain.py -d; sleep 10; cd ..;
+    kill -QUIT $(ps ax | grep bitmessagemain | cut -d ' ' -f 2);
+    sleep 10;
   - xvfb-run src/bitmessagemain.py -t
   - python setup.py test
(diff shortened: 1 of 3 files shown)


g1itch:
I'm going to squash e2673fb with d8099fd when you check the failure.

navjotcis:
I was testing PR and faces this issue on logs

g1itch:
> I was testing PR and faces this issue on logs.
> image

It is #1661, the bug, present in current v0.6.

PeterSurda:
can you rebase this on top of #1683 when ready?

.travis.yml
```diff
@@ -11,9 +11,12 @@ addons:
...
 install:
   - pip install -r requirements.txt
-  - ln -s src pybitmessage  # tests environment
   - python setup.py install
 script:
   - python checkdeps.py
+  - git clone -b v0.6.3 https://github.com/Bitmessage/PyBitmessage.git old;
+    cd old; src/bitmessagemain.py -d; sleep 10; cd ..;
+    kill -QUIT $(ps ax | grep bitmessagemain | cut -d ' ' -f 2);
+    sleep 10;
```
PeterSurda, line 20: How about instead we simply include a copy of old-format `knownnodes.dat`, and the test will put it into the config directory inside `setUp` and remove it during `tearDown`?

PeterSurda (changes requested):
I think the test should be done in a diffent manner. Maybe an additional test for old-style pickle, which will use a static file which will be included in the PR.

PR #1669, opened 2020-09-14
replaced insert into sent query with helper_sent.insert method and al…

### LINKED ISSUE #1524: uuid instead of empty msgid for new messages in sent fold
newly created messages have a empty msgid, which is later updated by the worker thread. This creates problems. I recommend using `uuid.uuid4().bytes` instead, which will again be replaced by the worker thread later. This can probably be done in `helper_sent` (except bitmessageqt which doesn't use that, so that needs to be refactord to use helper_sent). The signals the worker thread sends back should also be modified to send the old msgid if it differs from the new msgid, and not to use `updateSentItemStatusByToAddress` anymore.

The interfaces (curses, qt, api, kivy) need to be adjusted accordingly, so that they can replace the msgid if they get a new one from the signal.

- [ ] all interfaces that insert into sent table should use `helper_sent.insert`
  - [ ] Qt
  - [ ] kivy
  - [ ] curses
- [ ] `helper_sent.insert` should check the first argument is empty or None, and if yet, replace it with uuid, and return said uuid (otherwise return the first argument)
- [ ] the UI should retrieve this UUID and adjust the UI accordingly
  - [ ] Qt
  - [ ] kivy
  - [ ] curses

navjotcis:
1) I have replaced insert into sent query execution syntax with the helper_sent.insert method.
2) I have also replaced the empty msgid with UUID which will later be updated by a worker thread.
3) fixes #1524

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -90,7 +90,6 @@ from addresses import (
 )
 from bmconfigparser import BMConfigParser
 from debug import logger
-from helper_ackPayload import genAckPayload
 from helper_sql import SqlBulkExecute, sqlExecute, sqlQuery, sqlStoredProcedure
 from inventory import Inventory
 from network.threads import StoppableThread
@@ -1101,7 +1100,6 @@ class BMRPCDispatcher(object):
             TTL = 28 * 24 * 60 * 60
         toAddress = addBMIfNotPresent(toAddress)
         fromAddress = addBMIfNotPresent(fromAddress)
-        streamNumber, toRipe = self._verifyAddress(toAddress)[2:]
         self._verifyAddress(fromAddress)
         try:
             fromAddressEnabled = self.config.getboolean(
@@ -1112,11 +1110,7 @@ class BMRPCDispatcher(object):
         if not fromAddressEnabled:
             raise APIError(14, 'Your fromAddress is disabled. Cannot send.')
 
-        stealthLevel = self.config.safeGetInt(
-            'bitmessagesettings', 'ackstealthlevel')
-        ackdata = genAckPayload(streamNumber, stealthLevel)
-
-        helper_sent.insert(
+        ackdata = helper_sent.insert(
             toAddress=toAddress, fromAddress=fromAddress,
             subject=subject, message=message, encoding=encodingType)
 
@@ -1158,11 +1152,9 @@ class BMRPCDispatcher(object):
         except BaseException:
             raise APIError(
                 13, 'Could not find your fromAddress in the keys.dat file.')
-        streamNumber = decodeAddress(fromAddress)[2]
-        ackdata = genAckPayload(streamNumber, 0)
         toAddress = str_broadcast_subscribers
 
-        helper_sent.insert(
+        ackdata = helper_sent.insert(
             fromAddress=fromAddress, subject=subject,
             message=message, status='broadcastqueued',
             encoding=encodingType)
--- a/src/api.py
+++ b/src/api.py
@@ -1163,8 +1163,8 @@ class BMRPCDispatcher(object):
         toAddress = str_broadcast_subscribers
 
         helper_sent.insert(
-            toAddress=toAddress, fromAddress=fromAddress,
-            subject=subject, message=message, status='broadcastqueued',
+            fromAddress=fromAddress, subject=subject,
+            message=message, status='broadcastqueued',
             encoding=encodingType)
 
         toLabel = str_broadcast_subscribers
--- a/src/api.py
+++ b/src/api.py
@@ -1116,22 +1116,22 @@ class BMRPCDispatcher(object):
             'bitmessagesettings', 'ackstealthlevel')
         ackdata = genAckPayload(streamNumber, stealthLevel)
 
-        t = ('',
-             toAddress,
-             toRipe,
-             fromAddress,
-             subject,
-             message,
-             ackdata,
-             int(time.time()),  # sentTime (this won't change)
-             int(time.time()),  # lastActionTime
-             0,
-             'msgqueued',
-             0,
-             'sent',
-             encodingType,
-             TTL)
-        helper_sent.insert(t)
+        helper_sent.insert(
+            '',
+            toAddress,
+            toRipe,
+            fromAddress,
+            subject,
+            message,
+            ackdata,
+            int(time.time()),  # sentTime (this won't change)
+            int(time.time()),  # lastActionTime
+            0,
+            'msgqueued',
+            0,
+            'sent',
+            encodingType,
+            TTL)
 
         toLabel = ''
         queryreturn = sqlQuery(
@@ -1176,22 +1176,22 @@ class BMRPCDispatcher(object):
         toAddress = str_broadcast_subscribers
         ripe = ''
 
-        t = ('',
-             toAddress,
-             ripe,
-             fromAddress,
-             subject,
-             message,
-             ackdata,
-             int(time.time()),  # sentTime (this doesn't change)
-             int(time.time()),  # lastActionTime
-             0,
-             'broadcastqueued',
-             0,
-             'sent',
-             encodingType,
-             TTL)
-        helper_sent.insert(t)
+        helper_sent.insert(
+            '',
+            toAddress,
+            ripe,
+            fromAddress,
+            subject,
+            message,
+            ackdata,
+            int(time.time()),  # sentTime (this doesn't change)
+            int(time.time()),  # lastActionTime
+            0,
+            'broadcastqueued',
+            0,
+            'sent',
+            encodingType,
+            TTL)
 
         toLabel = str_broadcast_subscribers
         queues.UISignalQueue.put(('displayNewSentMessage', (
--- a/src/api.py
+++ b/src/api.py
@@ -1117,9 +1117,8 @@ class BMRPCDispatcher(object):
         ackdata = genAckPayload(streamNumber, stealthLevel)
 
         helper_sent.insert(
-            msgid='', toAddress=toAddress, ripe=toRipe, fromAddress=fromAddress,
-            subject=subject, message=message, ackdata=ackdata, encoding=encodingType,
-            ttl=TTL)
+            toAddress=toAddress, fromAddress=fromAddress,
+            subject=subject, message=message, encoding=encodingType)
 
         toLabel = ''
         queryreturn = sqlQuery(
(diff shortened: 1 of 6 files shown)


src/bitmessagecurses/__init__.py
```diff
@@ -31,6 +31,8 @@
 from helper_ackPayload import genAckPayload
 from helper_sql import sqlExecute, sqlQuery
 from inventory import Inventory
+
+import helper_sent
```
PeterSurda, line 35: wrong order, should be between lines 21 and 22

src/bitmessagecurses/__init__.py
```diff
@@ -982,7 +983,9 @@ def sendMessage(sender="", recv="", broadcast=None, subject="", body="", reply=F
                         0,      # retryNumber
                         "sent",
                         2,      # encodingType
-                        BMConfigParser().getint('bitmessagesettings', 'ttl'))
+                        BMConfigParser().getint('bitmessagesettings', 'ttl')
```
PeterSurda, line 986: I would move this variable into `helper_sent` as well. Since you're already working on the code, may as well do it.

src/helper_sent.py
```diff
@@ -2,9 +2,14 @@
 Insert values into sent table
 """
 
+import uuid
 from helper_sql import sqlExecute
 
 
 def insert(t):
     """Perform an insert into the `sent` table"""
+    msgid = uuid.uuid4().bytes
```
PeterSurda, line 11: only `if not t[0]:`

PeterSurda (changes requested):

src/bitmessagecurses/__init__.py
navjotcis: ok

src/helper_sent.py
navjotcis: ok

src/bitmessagecurses/__init__.py
navjotcis: ok

src/helper_sent.py
```diff
@@ -2,9 +2,16 @@
...
+from bmconfigparser import BMConfigParser
 from helper_sql import sqlExecute
 
 
 def insert(t):
     """Perform an insert into the `sent` table"""
+    temp = list(t)
+    if not t[0]:
+        temp[0] = uuid.uuid4().bytes    # if msgid is empty the put uuid
+    temp[-1] = BMConfigParser().getint('bitmessagesettings', 'ttl')
```
PeterSurda, line 15: This should be optional.

src/helper_sent.py
```diff
@@ -2,9 +2,16 @@
...
+    t = tuple(temp)
```
PeterSurda, line 16: Why is this necessary?

src/helper_sent.py
```diff
@@ -2,9 +2,16 @@
...
+import uuid
+from bmconfigparser import BMConfigParser
 from helper_sql import sqlExecute
 
 
 def insert(t):
     """Perform an insert into the `sent` table"""
+    temp = list(t)
+    if not t[0]:
+        temp[0] = uuid.uuid4().bytes    # if msgid is empty the put uuid
```
PeterSurda, line 14: ... empty **then** put ...

PeterSurda, line 14: Needs to be tested regarding sqlite data type. Write a test for this.

PeterSurda (changes requested):

src/helper_sent.py
```diff
@@ -2,9 +2,20 @@
...
+from bmconfigparser import BMConfigParser
 from helper_sql import sqlExecute
 
 
 def insert(t):
     """Perform an insert into the `sent` table"""
+    if not t[0] or not t[-1]:
+        temp = list(t)
+        if not t[0]:
+            print('zero index is not available')
```
PeterSurda, line 15: no `print`

src/helper_sent.py
```diff
@@ -2,9 +2,20 @@
...
+            temp[0] = uuid.uuid4().bytes    # if msgid is empty the put uuid
+        if not t[-1]:
+            print('Lasr index is not available')
```
PeterSurda, line 18: no `print`

PeterSurda (changes requested):
Remove `print`.

Write a unit tests that verifies that insert is working correctly, in particular look out that `uuid.uuid4().bytes` needs to be recorded in the database as the correct data type after retrieving it.

PeterSurda (changes requested):
no prints, add defaults, improve test

src/helper_sent.py
```diff
@@ -10,38 +10,41 @@
 from helper_sql import sqlExecute
 
 
-def insert(t, is_testcase=False):
+# pylint: disable=too-many-arguments
+def insert(msgid, toAddress, ripe, fromAddress, subject, message, ackdata,
+           sentTime, lastActionTime, sleeptill=0, status='msgqueued',
+           retryNumber=0, folder='sent', encoding=2, ttl=0, is_testcase=False):
```
PeterSurda, line 16: default values missing

src/bitmessageqt/__init__.py
```diff
@@ -2221,23 +2220,23 @@ def click_pushButtonSend(self):
...
-                     int(time.time()), # lastActionTime
-                     0, # sleepTill time. This will get set when the POW gets done.
-                     'broadcastqueued', 
-                     0, # retryNumber
-                     'sent', # folder
-                     encoding, # encoding type
-                     0
-                     )
-                helper_sent.insert(t)
+                helper_sent.insert(
```
PeterSurda, line 2223: should use named arguments, defaults should be omitted

PeterSurda (changes requested):
The helper function arguments need to have defaults, and calling it should only specify those arguments that aren't default.

src/helper_sent.py
navjotcis: which values are missing?

src/bitmessagecurses/__init__.py
```diff
@@ -992,24 +979,9 @@ def sendMessage(sender="", recv="", broadcast=None, subject="", body="", reply=F
             # dummy ackdata, no need for stealth
             ackdata = genAckPayload(decodeAddress(addr)[2], 0)
```
PeterSurda, line 980: can be removed now

src/helper_sent.py
```diff
@@ -2,9 +2,44 @@
...
+
+    status = status if status else 'msgqueued'
+    ttl = ttl if ttl else BMConfigParser().getint('bitmessagesettings', 'ttl')
+
+    t = (msgid, toAddress, ripe, fromAddress, subject, message, ackdata,
+         sentTime, lastActionTime, sleeptill, status, retryNumber, folder,
+         encoding, ttl)
+
     sqlExecute('''INSERT INTO sent VALUES (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)''', *t)
+    return msgid
```
PeterSurda, line 45: I changed my mind here, it should return `ackdata` and not `msgid`, based on reviewing existing code.

src/tests/core.py
```diff
@@ -238,6 +240,19 @@ def test_version(self):
...
+        fromAddress = 'BM-2cTrmD22fLRrumi3pPLg1ELJ6PdAaTRTdfg'
+        toAddress = 'BM-2cVWtdUzPwF7UNGDrZftWuHWgjdfkj89fdf'
+        message = 'test message'
+        subject = 'test subject'
+        status = 'msgqueued'
+        result = helper_sent.insert(
+            toAddress=toAddress, fromAddress=fromAddress, subject=subject,
+            message=message, status=status,
+        )
+        self.assertNotEqual(result, '')
```
PeterSurda, line 254: It should retrieve the row from the sent table based on `result` and do some check on the values.

src/class_objectProcessor.py
```diff
@@ -749,28 +749,14 @@ def processmsg(self, data):
 
                 ackdata = genAckPayload(streamNumber, 0)
                 toAddress = '[Broadcast subscribers]'
```
PeterSurda, line 751: these two lines can be removed.

src/helper_sent.py
```diff
@@ -2,9 +2,44 @@
...
+           status=None, ripe=None, ackdata=None, sentTime=None, lastActionTime=None,
+           sleeptill=0, retryNumber=0, encoding=2, ttl=None, folder='sent'):
     """Perform an insert into the `sent` table"""
+    # pylint: disable=unused-variable
+    # pylint: disable-msg=too-many-locals
+
+    msgid = msgid if msgid else uuid.uuid4().bytes
+
+    if not ripe or not ackdata:
+        new_status, addressVersionNumber, streamNumber, new_ripe = decodeAddress(toAddress)
```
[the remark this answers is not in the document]
PeterSurda: It should check if it's a broadcast either based on `status`, or on missing `toAddress`.

src/helper_sent.py
```diff
@@ -2,9 +2,44 @@
...
+import uuid
+from addresses import decodeAddress
+from bmconfigparser import BMConfigParser
+from helper_ackPayload import genAckPayload
 from helper_sql import sqlExecute
 
 
-def insert(t):
+# pylint: disable=too-many-arguments
+def insert(msgid=None, toAddress=None, fromAddress=None, subject=None, message=None,
```
PeterSurda, line 14: default `toAddress` should be `'[Broadcast subscribers]'`

PeterSurda (changes requested):
Remove statements from the calling methods that were moved to `insert` (`ripe`, `genAckPayload`, ...).

src/helper_sent.py
navjotcis: ok

src/class_objectProcessor.py
navjotcis: we can not remove line 751 because it is used in below lines

PeterSurda: ok

src/api.py
```diff
@@ -1174,24 +1161,11 @@ def HandleSendBroadcast(
         streamNumber = decodeAddress(fromAddress)[2]
```
PeterSurda, line 1161: remove

src/tests/core.py
```diff
@@ -238,6 +241,20 @@ def test_version(self):
...
+        toAddress = 'BM-2cVWtdUzPwF7UNGDrZftWuHWgjdfkj89fdf'
+        message = 'test message'
+        subject = 'test subject'
+        result = helper_sent.insert(
+            toAddress=toAddress, fromAddress=fromAddress,
+            subject=subject, message=message,
+        )
+        queryreturn = sqlQuery(
+            '''select msgid from sent where ackdata=?''', result)
+        self.assertNotEqual(queryreturn[0][0], '')
```
PeterSurda, line 256: test the data type for ackdata as well.

PeterSurda (changes requested):

PeterSurda (approved):

PR #1671, opened 2020-09-18
fixed the app closing issue on the press of quit button

navjotcis:
Here when user try to quit the app then the app does not close it takes time longer than expected so I have fixed it for two cases
1) when the user open the app and closes it before the connection popup shows then the app does not close. It runs in a loop for n no. of times which takes sometimes 2-3 minutes or longer then it so I have added the condition with it for checking if the user presses quit then it will not go into the loop and close the app.
2) This case appears sometimes so here also app not close when the user tries to quit because there is condition pendingDownload() > 0 so app-only close when pendingDownload() returns 0 or less so the value of pendingDownload()  method is not fixed sometimes it gives a large no. which run in the loop N no. of times so also it does not close the app until loop executes N no. of times so I have also added the condition here when the user press the quit then it does not go into the loop and close the app.


PeterSurda:
This isn't working correctly. I know I made recommendations, but at best they were incomplete. Find a way to fix this properly.

navjotcis:
I have tested all the cases so I have made a table for that and all the cases are passed

navjotcis:
I have changed the UI of the app closing popup and also when the user click on the Quit button I am showing a progress bar and on the press of don't quit progress bar will disappear.

PR #1672, opened 2020-09-19
implemented filter query in sent table on the bases of folder= 'sent'

navjotcis:
I have implemented a filter query for the sent table on the bases of folder ='sent' because in  kivy app there is one more feature draft message so for storing draft message they are using the same table sent only the folder value is change on the bases of folder type so I have implemented all the sent table filter query in class_singleWorker.py and class_singleCleaner.py files.

### DIFF
--- a/src/class_singleWorker.py
+++ b/src/class_singleWorker.py
@@ -109,7 +109,7 @@ class singleWorker(StoppableThread):
                 newack = '\x00\x00\x00\x02\x01\x01' + oldack
                 state.ackdataForWhichImWatching[newack] = 0
                 sqlExecute(
-                    'UPDATE sent SET ackdata=? WHERE ackdata=? AND folder = "sent" ',
+                    '''UPDATE sent SET ackdata=? WHERE ackdata=? AND folder = 'sent' ''',
                     newack, oldack
                 )
                 del state.ackdataForWhichImWatching[oldack]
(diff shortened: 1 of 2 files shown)


src/class_singleWorker.py
```diff
@@ -681,7 +683,7 @@ def sendBroadcast(self):
             # a 'broadcastsent' status
             sqlExecute(
                 'UPDATE sent SET msgid=?, status=?, lastactiontime=?'
-                ' WHERE ackdata=?',
+                ' WHERE ackdata=? AND folder="sent" ',
```
PeterSurda, line 686: Single quotes: ` AND folder='sent' `

src/class_singleWorker.py
```diff
@@ -1294,11 +1300,10 @@ def sendMsg(self):
             sleepTill = int(time.time() + TTL * 1.1)
             sqlExecute(
                 '''UPDATE sent SET msgid=?, status=?, retrynumber=?, '''
-                ''' sleeptill=?, lastactiontime=? WHERE ackdata=?''',
+                ''' sleeptill=?, lastactiontime=? WHERE ackdata=? AND folder='sent' ''',
                 inventoryHash, newStatus, retryNumber + 1,
                 sleepTill, int(time.time()), ackdata
             )
-
```
PeterSurda, line 1301: why?

PeterSurda (changes requested):

src/class_singleWorker.py
navjotcis: ok

PeterSurda (approved):

PR #1676, opened 2020-09-25
implemented test case for sendBroadcast

navjotcis:
I have written a test case for sendBroadcast here I am sending a message and checking the status of the message on the bases of ackdata.


g1itch:
I don't understand what are you trying to do. There is the PR for API refactoring already containing similar changes:
https://github.com/Bitmessage/PyBitmessage/pull/1382/files#diff-5e0b291ba4f1474709fc8d4bcf4805d7R272

Issue #1147, opened 2018-03-07
Try Icon color, inconsistent state


PeterSurda:
This came in the bitmessage chan:

hello!
since last update the tray icon is always red, independently if there are connections. The icon in the main window correctly shows yellow. 

And another issue that I see sometimes, also with former version:
- I visit the lists of new messages and read and/or delete all of them
- new messages: 0
restart Bitmessage
- the line "new" is bold and an old number is listed beside
- when I click on "new" the list is empty, and the state is corrected

g1itch:
I doubt that the two issues are related. About tray icon my first suggestion that OP runs v0.6.2. 

I also cannot reproduce the second issue but I've seen another effect. Then you add some unread messages to trash after restart BM shows no unread messages in that tree.

navjotcis:
Here the first issue seems to be fixed and I have checked with second issue but I am not able to reproduce it

navjotcis:
> I doubt that the two issues are related. About tray icon my first suggestion that OP runs v0.6.2.
> 
> I also cannot reproduce the second issue but I've seen another effect. Then you add some unread messages to the trash after restart BM shows no unread messages in that tree.

this is also fixed I think as I have checked first I have deleted the inbox unread message and move it to the trash then it is showing unread in the trash as well.
![image](https://user-images.githubusercontent.com/33522186/92246941-17f35a80-eee4-11ea-8dd3-09ad7b4c1808.png)

navjotcis:
I have deleted all the inbox messages and restart the Bitmessage then "new" is showing in the bold and the old count is showing there and when I open the "new" then the list was empty so this issue is only reproducible some times because I tried to check it again 5 times but it appears only 2 times. 
There is also one more issue here when the user moves an unread message of the inbox into the trash it shows unread in the trash list so I have tried to fixed the unread trash message issue and then both the issues get solved.

navjotcis:
fixes https://github.com/Bitmessage/PyBitmessage/pull/1680

g1itch:
Today I received the comment about red tray icon, suggesting that it's also related to setuptools.
![image](https://user-images.githubusercontent.com/4012700/107979726-bfaea180-6fc7-11eb-9075-70bbc5e23eca.png)

As for marking all trashed messages read, I reverted that change in my qt5-wip branch. I disagree with that decision.

PR #1680, opened 2020-10-08
fixed unread trash message listing issue

navjotcis:
I have worked on updating the inbox table read value to 1(which means the message is read) so now when the user will move the unread inbox message to the trash message listing then he will not see the unread sign in the trash message listing.because all the message which moves in trash message listing should change the status from unread to read.


PeterSurda (approved):
I haven't tested it but even if it doesn't fully fix everything, it fixes some things.

navjotcis:
This PR has also fixed the problem of https://github.com/Bitmessage/PyBitmessage/issues/1147 issue.

PR #1681, opened 2020-10-15
Fixes #846

### LINKED ISSUE #846: Spell checking
A user suggested over BM that spellchecking would be a good feature. We obviously can't use online checking API so somehow the languages would have to be bundled, or there could be an option to download it separately.

scottwedge:
Ran spell check against all Python files and found these minor errors.


g1itch:
#846 is about spellchecking an outgoing message in `MessageCompose` widget.

PR #1683, opened 2020-10-23
Fix broken tests

PeterSurda:
- setting `BITMESSAGE_HOME` doesn't always work, so we wrote a function to do this
- `logger` and `BMConfigParser` now clear old config when loading new
- add method for flushing log files
- rework logic in `test_logger`
- detect thread names on UNIX-like OSes
- be more accurate when starting or stopping a process, better waiting mechanisms
- add extra cleanup during `TestProcess` start

### DIFF
--- a/src/tests/test_config.py
+++ b/src/tests/test_config.py
@@ -45,6 +45,7 @@ class TestProcessConfig(TestProcessProto):
     def test_config_defaults(self):
         """Test settings in the generated config"""
         self._stop_process()
+        self._kill_process()
         config = BMConfigParser()
         config.read(os.path.join(self.home, 'keys.dat'))
 
--- a/src/tests/test_logger.py
+++ b/src/tests/test_logger.py
@@ -7,6 +7,8 @@ import os
 import tempfile
 import unittest
 
+from pybitmessage.paths import set_appdata_folder
+
 
 class TestLogger(unittest.TestCase):
     """A test case for bmconfigparser"""
@@ -35,11 +37,14 @@ level=DEBUG
 handlers=default
 '''
 
-    def test_fileConfig(self):
+    def test_fileConfig(self):  # pylint: disable=invalid-name
         """Put logging.dat with special pattern and check it was used"""
-        tmp = os.environ['BITMESSAGE_HOME'] = tempfile.gettempdir()
+        # FIXME: this doesn't always work, probably due to module
+        # load order. Workaround through .travis.yml
+        tmp = set_appdata_folder(tempfile.gettempdir())
         log_config = os.path.join(tmp, 'logging.dat')
         log_file = os.path.join(tmp, 'debug.log')
+        pass_ = 1
 
         def gen_log_config(pattern):
             """A small closure to generate logging.dat with custom pattern"""
@@ -48,11 +53,14 @@ handlers=default
 
         pattern = r' o_0 '
         gen_log_config(pattern)
+        pass_ = 1
 
         try:
-            from pybitmessage.debug import logger, resetLogging
+            # pylint: disable=import-outside-toplevel
+            from pybitmessage.debug import logger, resetLogging, flush_logs
             if not os.path.isfile(log_file):  # second pass
                 pattern = r' <===> '
+                pass_ = 2
                 gen_log_config(pattern)
                 resetLogging()
         except ImportError:
@@ -60,10 +68,17 @@ handlers=default
         finally:
             os.remove(log_config)
 
+        # pylint: disable=invalid-name
+        self.longMessage = True
+
         logger_ = logging.getLogger('default')
 
-        self.assertEqual(logger, logger_)
+        self.assertEqual(logger, logger_, ", pass {}".
+                         format(pass_))
 
         logger_.info('Testing the logger...')
+        flush_logs()
 
-        self.assertRegexpMatches(open(log_file).read(), pattern)
+        # pylint: disable=deprecated-method
+        self.assertRegexpMatches(open(log_file).read(), pattern,
+                                 ", pass {}".format(pass_))
--- a/src/tests/test_process.py
+++ b/src/tests/test_process.py
@@ -5,11 +5,14 @@ Common reusable code for tests and tests for pybitmessage process.
 import os
 import signal
 import subprocess  # nosec
+import sys
 import tempfile
 import time
 import unittest
 
 import psutil
+
+from pybitmessage.paths import set_appdata_folder
 
 
 def put_signal_file(path, filename):
@@ -23,7 +26,26 @@ class TestProcessProto(unittest.TestCase):
     it starts pybitmessage in setUpClass() and stops it in tearDownClass()
     """
     _process_cmd = ['pybitmessage', '-d']
-    _threads_count = 15
+    _threads_count_min = 15
+    _threads_count_max = 16
+    _threads_names = [
+        'PyBitmessage',
+        'addressGenerato',
+        'singleWorker',
+        'SQL',
+        'objectProcessor',
+        'singleCleaner',
+        'singleAPI',
+        'Asyncore',
+        'ReceiveQueue_0',
+        'ReceiveQueue_1',
+        'ReceiveQueue_2',
+        'Announcer',
+        'InvBroadcaster',
+        'AddrBroadcaster',
+        'Downloader',
+        'Uploader'
+    ]
     _files = (
         'keys.dat', 'debug.log', 'messages.dat', 'knownnodes.dat',
         '.api_started', 'unittest.lock'
@@ -32,10 +54,28 @@ class TestProcessProto(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         """Setup environment and start pybitmessage"""
-        cls.home = os.environ['BITMESSAGE_HOME'] = tempfile.gettempdir()
+        cls.home = set_appdata_folder(tempfile.gettempdir())
+        cls._cleanup_files()
         put_signal_file(cls.home, 'unittest.lock')
+        starttime = int(time.time())
         subprocess.call(cls._process_cmd)  # nosec
-        time.sleep(5)
+        timeout = starttime + 30
+        while time.time() <= timeout:
+            try:
+                if os.path.exists(os.path.join(cls.home,
+                                               'singleton.lock')):
+                    pstat = os.stat(os.path.join(cls.home, 'singleton.lock'))
+                    if starttime <= pstat.st_mtime and pstat.st_size > 0:
+                        break
+            except OSError:
+                break
+            time.sleep(1)
+        if time.time() >= timeout:
+            raise psutil.TimeoutExpired(
+                "Timeout starting {}".format(cls._process_cmd))
+        # wait a bit for the program to fully start
+        # 10 sec should be enough
+        time.sleep(10)
         cls.pid = int(cls._get_readline('singleton.lock'))
         cls.process = psutil.Process(cls.pid)
 
@@ -57,6 +97,18 @@ class TestProcessProto(unittest.TestCase):
         return True
 
     @classmethod
+    def _kill_process(cls, timeout=5):
+        try:
+            cls.process.send_signal(signal.SIGKILL)
+            cls.process.wait(timeout)
+        # Windows or already dead
+        except (AttributeError, psutil.NoSuchProcess):
+            return True
+        # except psutil.TimeoutExpired propagates, it means something is very
+        # wrong
+        return True
+
+    @classmethod
     def _cleanup_files(cls):
         for pfile in cls._files:
             try:
@@ -69,33 +121,57 @@ class TestProcessProto(unittest.TestCase):
         """Ensures that pybitmessage stopped and removes files"""
         try:
             if not cls._stop_process():
-                print(open(os.path.join(cls.home, 'debug.log'), 'rb').read())
                 cls.process.kill()
-        except psutil.NoSuchProcess:
+                cls.process.wait(5)
+        except (psutil.TimeoutExpired, psutil.NoSuchProcess):
             pass
         finally:
             cls._cleanup_files()
 
     def _test_threads(self):
-        # only count for now
-        # because of https://github.com/giampaolo/psutil/issues/613
-        # PyBitmessage
-        #   - addressGenerator
-        #   - singleWorker
-        #   - SQL
-        #   - objectProcessor
-        #   - singleCleaner
-        #   - singleAPI
-        #   - Asyncore
-        #   - ReceiveQueue_0
-        #   - ReceiveQueue_1
-        #   - ReceiveQueue_2
-        #   - Announcer
-        #   - InvBroadcaster
-        #   - AddrBroadcaster
-        #   - Downloader
-        self.assertEqual(
-            len(self.process.threads()), self._threads_count)
+        """Test number and names of threads"""
+
+        # pylint: disable=invalid-name
+        self.longMessage = True
+
+        try:
+            # using ps for posix platforms
+            # because of https://github.com/giampaolo/psutil/issues/613
+            thread_names = subprocess.check_output([
+                "ps", "-L", "-o", "comm=", "--pid",
+                str(self.process.pid)
+            ]).split()
+        except:  # pylint: disable=bare-except
+            thread_names = []
+
+        running_threads = len(thread_names)
+        if 0 < running_threads < 30:  # adequacy check
+            extra_threads = []
+            missing_threads = []
+            for thread_name in thread_names:
+                if thread_name not in self._threads_names:
+                    extra_threads.append(thread_name)
+            for thread_name in self._threads_names:
+                if thread_name not in thread_names:
+                    missing_threads.append(thread_name)
+
+            msg = "Missing threads: {}, Extra threads: {}".format(
+                ",".join(missing_threads), ",".join(extra_threads))
+        else:
+            running_threads = self.process.num_threads()
+            if sys.platform.startswith('win'):
+                running_threads -= 1  # one extra thread on Windows!
+            msg = "Unexpected running thread count"
+
+        self.assertGreaterEqual(
+            running_threads,
+            self._threads_count_min,
+            msg)
+
+        self.assertLessEqual(
+            running_threads,
+            self._threads_count_max,
+            msg)
 
 
 class TestProcessShutdown(TestProcessProto):
@@ -110,7 +186,14 @@ class TestProcessShutdown(TestProcessProto):
     @classmethod
     def tearDownClass(cls):
         """Special teardown because pybitmessage is already stopped"""
-        cls._cleanup_files()
+        try:
+            if cls.process.is_running():
+                cls.process.kill()
+                cls.process.wait(5)
+        except (psutil.TimeoutExpired, psutil.NoSuchProcess):
+            pass
+        finally:
+            cls._cleanup_files()
 
 
 class TestProcess(TestProcessProto):
--- a/src/bmconfigparser.py
+++ b/src/bmconfigparser.py
@@ -7,7 +7,7 @@ import os
 import shutil
 from datetime import datetime
 
-import state
+from paths import get_active_config_folder
 from singleton import Singleton
 
 BMConfigDefaults = {
@@ -124,7 +124,16 @@ class BMConfigParser(ConfigParser.SafeConfigParser):
         return [
             x for x in BMConfigParser().sections() if x.startswith('BM-')]
 
+    def _reset(self):
+        """Reset current config. There doesn't appear to be a built in
+           method for this"""
+        sections = self.sections()
+        for x in sections:
+            self.remove_section(x)
+
     def read(self, filenames):
+        """Read config and populate defaults"""
+        self._reset()
         ConfigParser.ConfigParser.read(self, filenames)
         for section in self.sections():
             for option in self.options(section):
@@ -144,7 +153,7 @@ class BMConfigParser(ConfigParser.SafeConfigParser):
 
     def save(self):
         """Save the runtime config onto the filesystem"""
-        fileName = os.path.join(state.appdata, 'keys.dat')
+        fileName = os.path.join(get_active_config_folder(), 'keys.dat')
         fileNameBak = '.'.join([
             fileName, datetime.now().strftime("%Y%j%H%M%S%f"), 'bak'])
         # create a backup copy to prevent the accidental loss due to
--- a/src/class_sqlThread.py
+++ b/src/class_sqlThread.py
@@ -28,7 +28,9 @@ class sqlThread(threading.Thread):
 
     def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements
         """Process SQL queries from `.helper_sql.sqlSubmitQueue`"""
-        self.conn = sqlite3.connect(state.appdata + 'messages.dat')
+        self.conn = sqlite3.connect(
+            os.path.join(paths.get_active_config_folder(),
+                         'messages.dat'))
         self.conn.text_factory = str
         self.cur = self.conn.cursor()
 
--- a/src/debug.py
+++ b/src/debug.py
@@ -42,7 +42,8 @@ import os
 import sys
 
 import helper_startup
-import state
+
+from paths import get_active_config_folder
 
 helper_startup.loadConfig()
 
@@ -67,7 +68,7 @@ def configureLogging():
     sys.excepthook = log_uncaught_exceptions
     fail_msg = ''
     try:
-        logging_config = os.path.join(state.appdata, 'logging.dat')
+        logging_config = os.path.join(get_active_config_folder(), 'logging.dat')
         logging.config.fileConfig(
             logging_config, disable_existing_loggers=False)
         return (
@@ -103,7 +104,7 @@ def configureLogging():
                 'class': 'logging.handlers.RotatingFileHandler',
                 'formatter': 'default',
                 'level': log_level,
-                'filename': os.path.join(state.appdata, 'debug.log'),
+                'filename': os.path.join(get_active_config_folder(), 'debug.log'),
                 'maxBytes': 2097152,  # 2 MiB
                 'backupCount': 1,
                 'encoding': 'UTF-8',
@@ -140,12 +141,21 @@ def resetLogging():
     """Reconfigure logging in runtime when state.appdata dir changed"""
     # pylint: disable=global-statement, used-before-assignment
     global logger
-    for i in logger.handlers:
-        logger.removeHandler(i)
+    for i in list(logger.handlers):
+        try:
+            logger.removeHandler(i)
+        except ValueError:
+            pass
         i.flush()
         i.close()
     configureLogging()
     logger = logging.getLogger('default')
+
+
+def flush_logs():
+    """Tell handlers to flush"""
+    for i in list(logger.handlers):
+        i.flush()
 
 
 # !
--- a/src/network/knownnodes.py
+++ b/src/network/knownnodes.py
@@ -16,6 +16,7 @@ except ImportError:
 import state
 from bmconfigparser import BMConfigParser
 from network.node import Peer
+from paths import get_active_config_folder
 
 knownNodesLock = threading.RLock()
 """Thread lock for knownnodes modification"""
@@ -91,7 +92,7 @@ def pickle_deserialize_old_knownnodes(source):
 def saveKnownNodes(dirName=None):
     """Save knownnodes to filesystem"""
     if dirName is None:
-        dirName = state.appdata
+        dirName = get_active_config_folder()
     with knownNodesLock:
         with open(os.path.join(dirName, 'knownnodes.dat'), 'wb') as output:
             json_serialize_knownnodes(output)
@@ -149,7 +150,9 @@ def createDefaultKnownNodes():
 def readKnownNodes():
     """Load knownnodes from filesystem"""
     try:
-        with open(state.appdata + 'knownnodes.dat', 'rb') as source:
+        with open(os.path.join(get_active_config_folder(),
+                               'knownnodes.dat'),
+                  'rb') as source:
             with knownNodesLock:
                 try:
                     json_deserialize_knownnodes(source)
--- a/src/paths.py
+++ b/src/paths.py
@@ -8,6 +8,8 @@ import sys
 from datetime import datetime
 from shutil import move
 
+import state
+
 logger = logging.getLogger('default')
 
 # When using py2exe or py2app, the variable frozen is added to the sys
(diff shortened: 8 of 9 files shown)


g1itch:
Ooogh, it looks sick ):

.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: python
 python:
   - "2.7"
+env:
+  - BITMESSAGE_HOME=/tmp
```
g1itch, line 5: Sad. I didn't planned to bound the tests to travis.

src/helper_addressbook.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python
```
g1itch, line 1: Why shebang here?

src/tests/test_process.py
```diff
@@ -110,7 +172,16 @@ def test_shutdown(self):
     @classmethod
     def tearDownClass(cls):
         """Special teardown because pybitmessage is already stopped"""
-        cls._cleanup_files()
+        try:
+            if cls.process.is_running():
+                cls.process.kill()
```
g1itch, line 177: This looks like a cheat. `TestProcessProto` does the same: https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/tests/test_process.py#L73

g1itch:
I don't think that increasing timeouts gonna help.

src/bitmessageqt/tests/support.py
```diff
@@ -12,6 +13,10 @@ class TestSupport(TestBase):
     SUPPORT_ADDRESS = 'BM-2cUdgkDDAahwPAU6oD2A7DnjqZz3hgY832'
     SUPPORT_SUBJECT = 'Support request'
 
+    def tearDown(self):
+        """Delete the support address from addressbook"""
+        delete_from_addressbook(self.SUPPORT_ADDRESS)
```
g1itch, line 18: core tests run from appdata

src/helper_addressbook.py
g1itch: Another helper ):

src/bitmessageqt/tests/support.py
```diff
@@ -11,6 +13,17 @@ class TestSupport(TestBase):
     """A test case for support module"""
     SUPPORT_ADDRESS = 'BM-2cUdgkDDAahwPAU6oD2A7DnjqZz3hgY832'
     SUPPORT_SUBJECT = 'Support request'
+    SUPPORT_MY_LABEL = 'My new address'
```
g1itch, line 16: _translate ?

src/tests/test_process.py
```diff
@@ -94,8 +129,35 @@ def _test_threads(self):
         #   - InvBroadcaster
         #   - AddrBroadcaster
         #   - Downloader
-        self.assertEqual(
-            len(self.process.threads()), self._threads_count)
+        #   - Uploader
+
+        self.longMessage = True
```
g1itch, line 134: I cannot find where it's used

g1itch: Oh, it's from unittest.TestCase. Got it.

src/tests/test_process.py
```diff
@@ -94,8 +129,35 @@ def _test_threads(self):
...
+
```
g1itch, line 135: If this hunk is platform dependent, maybe here should be some condition?

g1itch (approved):
At your discretion

g1itch:
See more: https://travis-ci.org/github/g1itch/PyBitmessage/builds/738338172

.travis.yml
PeterSurda: I agree but I don't think it can be fixed without a lot of refactoring.

src/helper_addressbook.py
PeterSurda: Ok.

PeterSurda: I mean, I agree about the shebang, but the addressbook code shouldn't be in the UI like it is now. What other options are there? Not another thread to deal with addressbook?

src/tests/test_process.py
PeterSurda: Yes it's so that we have more info about the failures.

src/tests/test_process.py
PeterSurda: I haven't tested it on Windows, but I assumed the result will be empty. Maybe it can be wrapped in try/except. The output is only used if there is an assertion error, it's not an assertion itself.

src/tests/test_process.py
PeterSurda: 1. it isn't the same, because `TestProcessProto` first calls `_stop_process` to send a `SIGTERM`. This expects the program to already have closed on its own, and just send `SIGKILL` in an emergency.
2. it allows you to put custom handlers into `TimeoutExpired` and `NoSuchProcess` if you want the tests to be more strict. `TimeoutExpired` means that there is something very wrong as it's blocking a `SIGKILL`.
3. However, the main point here is line 178 as previously, it wasn't waiting for the program to finish after `SIGKILL`, which I think lead unpredictably to problems.

src/bitmessageqt/tests/support.py
PeterSurda: Maybe yes, but noone is going to do that.

PeterSurda:
> I don't think that increasing timeouts gonna help.

Noone is working on a proper fix and other work can't be done until the tests succeed.

PeterSurda:
> See more: https://travis-ci.org/github/g1itch/PyBitmessage/builds/738338172
> I'm waiting.

`test_config_defaults` isn't waiting for the process to finish.

src/bitmessageqt/tests/support.py
PeterSurda: ok

src/bitmessageqt/tests/support.py
g1itch: I did it in one of my rebased branches. It did not help, because cleanup is the least part of the problem.

g1itch:
> > I don't think that increasing timeouts gonna help.
> 
> Noone is working on a proper fix and other work can't be done until the tests succeed.

I don't remember any shutdown tests failing in your test environment or current xenial travis setup. APIShutdown failed on windows and on my recent fedora31 container.

src/tests/test_process.py
```diff
@@ -34,8 +53,22 @@ def setUpClass(cls):
...
+        while time.time() <= timeout:
+            try:
+                pstat = os.stat(os.path.join(cls.home, 'singleton.lock'))
+                if starttime <= pstat.st_mtime and pstat.st_size > 0:
+                    break
+            except OSError:
+                pass
+            time.sleep(1)
+        if time.time() >= timeout:
+            raise OSError("Timeout starting {}".format(cls._process_cmd))
```
g1itch, line 68: https://travis-ci.org/github/g1itch/PyBitmessage/jobs/738338187

src/helper_addressbook.py
g1itch: I agree that it's cool to have all CRUD tasks (on messages, addressbook and black/white lists) in the main package to use them in `api`, `bitmessageqt` and any other frontend. Also I was working on the single helper with all SQL queries in #1534 because it is bad to have SQL scattered amond the code. It needs heavy testing. Currently it breaks `bitmessageqt` because I missed some type cast or check.

But hey, you need this helper now only because of `BITMESSAGE_HOME` in env and the absence of cleanup.

Another thread is the last thing I would suggest. But recently I thought about using multiprocessing for test cases which use potentially dangerous imports.

src/bitmessageqt/tests/support.py
g1itch: Did you try this?
```patch
--- a/src/tests/test_process.py
+++ b/src/tests/test_process.py
@@ -33,6 +33,7 @@ class TestProcessProto(unittest.TestCase):
     def setUpClass(cls):
         """Setup environment and start pybitmessage"""
         cls.home = os.environ['BITMESSAGE_HOME'] = tempfile.gettempdir()
+        cls._cleanup_files()
         put_signal_file(cls.home, 'unittest.lock')
         subprocess.call(cls._process_cmd)  # nosec
         time.sleep(5)

```

PeterSurda:
> I don't remember any shutdown tests failing in your test environment or current xenial travis setup. `TestAPIShutdown` failed on windows and on my recent fedora31 container.

I can put the timeouts back, maybe some other changes that I made fixed the issues.
 
> Windows failed until this run: https://travis-ci.org/github/g1itch/PyBitmessage/jobs/736655954

Looks like pyelliptic pointers.

src/tests/test_process.py
PeterSurda: That can be fixed later. We don't have a working OSX at the moment anyway.

src/bitmessageqt/tests/support.py
PeterSurda: ok let me try

PeterSurda:
@g1itch I refactored the use of state.appdata and removed the environment variable from .travis.yml. It now passes buildbot's bionic and xenial, as well as xenial on travis-ci.org. Can you try rebasing your changes on top of it again?

src/tests/test_process.py
g1itch: OK, I get it:
- it doesn't disable the test because it's failing by it's own if process not stops.
- it is not the same as super

g1itch:
Wow! It looks like you really fixed the major issue with appdata while I was on a half way retracing it from debug -> helper_startap <-> state (loadConfig).

g1itch:
> @g1itch I refactored the use of `state.appdata` and removed the environment variable from `.travis.yml`. It now passes buildbot's bionic and xenial, as well as xenial on travis-ci.org. Can you try rebasing your changes on top of it again?

I do the opposite: I merge it to cleaned out ci branch: https://travis-ci.org/github/g1itch/PyBitmessage/builds/738506280

PeterSurda:
> I do the opposite

Whichever you prefer.

src/tests/test_process.py
```diff
@@ -56,6 +92,17 @@ def _stop_process(cls, timeout=5):
...
 
+    @classmethod
+    def _kill_process(cls, timeout=5):
+        try:
+            cls.process.send_signal(signal.SIGKILL)
+            cls.process.wait(timeout)
+        except psutil.NoSuchProcess:
+            return True
+        except psutil.TimeoutExpired:
+            raise OSError("Process is stuck")
```
g1itch, line 103: Maybe `self.fail()` here?

src/tests/test_process.py
g1itch: https://travis-ci.org/github/g1itch/PyBitmessage/jobs/738510444

src/tests/test_config.py
```diff
@@ -45,6 +45,7 @@ class TestProcessConfig(TestProcessProto):
     def test_config_defaults(self):
         """Test settings in the generated config"""
         self._stop_process()
+        self._kill_process()
```
g1itch, line 48: This is like `tearDown()` without cleaning the files.

g1itch (approved):
Some formatting options and two errors on windows I found.

g1itch:
> > Windows failed until this run: https://travis-ci.org/github/g1itch/PyBitmessage/jobs/736655954
> 
> Looks like pyelliptic pointers.

That blindsig tests failed because of openssl-1.0.1 dll. If I install PyQt4 like in `winbuild.sh` it will fix the tests, because pyelliptic loads openssl-1.0.1 dll bundled with pyqt in that case.

src/tests/test_process.py
```diff
@@ -56,6 +92,17 @@ def _stop_process(cls, timeout=5):
             return False
         return True
 
+    @classmethod
+    def _kill_process(cls, timeout=5):
+        try:
+            cls.process.send_signal(signal.SIGKILL)
```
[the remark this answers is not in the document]
PeterSurda: ok

src/tests/test_process.py
```diff
@@ -69,8 +116,10 @@ def tearDownClass(cls):
         """Ensures that pybitmessage stopped and removes files"""
         try:
             if not cls._stop_process():
-                print(open(os.path.join(cls.home, 'debug.log'), 'rb').read())
                 cls.process.kill()
+                cls.process.wait(5)
+        except psutil.TimeoutExpired:
+            pass
```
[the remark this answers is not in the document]
PeterSurda: ok, it can be split later in case you want to handle them differently

src/tests/test_process.py
```diff
@@ -110,7 +191,16 @@ def test_shutdown(self):
...
         """Special teardown because pybitmessage is already stopped"""
-        cls._cleanup_files()
+        try:
+            if cls.process.is_running():
+                cls.process.kill()
+                cls.process.wait(5)
+        except psutil.TimeoutExpired:
+            pass
+        except psutil.NoSuchProcess:
+            pass
```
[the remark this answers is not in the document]
PeterSurda: same as above

src/tests/test_process.py
PeterSurda: ok

src/tests/test_process.py
```diff
@@ -94,8 +143,40 @@ def _test_threads(self):
...
+                extra_threads.append(thread_name)
+        for thread_name in self._threads_names:
+            if thread_name not in thread_names:
+                missing_threads.append(thread_name)
+
+        msg = "Missing threads: {}, Extra threads: {}".format(
+            ",".join(missing_threads), ",".join(extra_threads))
+
+        self.assertGreaterEqual(
+            len(self.process.threads()),
```
g1itch, line 172: `AccessDenied` on osx: https://travis-ci.org/github/g1itch/PyBitmessage/jobs/738524694

PeterSurda:
@g1itch Thanks for the feedback so far, I'll fix those issues.

src/tests/test_process.py
PeterSurda: `_kill_process` is a class method so you'd have to use `cls.fail()` and I don't think that would work correctly. The documentation doesn't say what happens if there are exceptions in `tearDownClass`, only in `tearDown`. Looking at [unittest source](https://github.com/python/cpython/blob/0f221d09cad46bee38d1b7a7822772df66c53028/Lib/unittest/suite.py), it looks like an exception in `tearDownClass` also adds a new error entry in the totals, just like `tearDown` is supposed to. So I think a more traditional exception like `OSError` is a better choice here. Perhaps an even better choise is to simply delete this `except` and let the `psutil.TimeoutExpired` propagate up.

src/tests/test_config.py
PeterSurda: Yes I think you're correct.

src/tests/test_process.py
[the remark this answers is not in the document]
PeterSurda: Added an exception handler.

src/tests/test_process.py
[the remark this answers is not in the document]
PeterSurda: Done

src/tests/test_process.py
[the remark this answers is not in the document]
PeterSurda: done

src/tests/test_process.py
PeterSurda: Test is now skipped on OSX.

g1itch (approved):
Non-linux issues may be fixed later

src/tests/test_process.py
```diff
@@ -32,10 +55,25 @@ class TestProcessProto(unittest.TestCase):
...
-        time.sleep(5)
+        timeout = starttime + 30.0
+        while time.time() <= timeout:
+            try:
+                pstat = os.stat(os.path.join(cls.home, 'singleton.lock'))
+                if starttime <= pstat.st_mtime and pstat.st_size > 0:
+                    break
+            except OSError:
+                pass
+            time.sleep(1)
```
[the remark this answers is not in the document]
g1itch: Not so easy: https://travis-ci.org/github/g1itch/PyBitmessage/builds/738699572

src/tests/test_process.py
```diff
@@ -94,8 +144,42 @@ def _test_threads(self):
         #   - InvBroadcaster
         #   - AddrBroadcaster
         #   - Downloader
-        self.assertEqual(
-            len(self.process.threads()), self._threads_count)
+        #   - Uploader
```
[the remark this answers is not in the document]
PeterSurda: Yea makes sense.

src/tests/test_process.py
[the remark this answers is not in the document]
g1itch: Rebased in my favor: https://travis-ci.org/github/g1itch/PyBitmessage/builds/738705081

src/tests/test_process.py
```diff
@@ -94,8 +144,42 @@ def _test_threads(self):
...
+
+        self.assertGreaterEqual(
+            running_threads,
+            self._threads_count_min,
+            msg)
+
+        self.assertLessEqual(
+            running_threads,
+            self._threads_count_max,
+            msg)
```
g1itch, line 182:
```patch
commit 77680a2ef4a9dd16736960e457529a66d7a6856d
Author: Dmitri Bogomolov <4glitch@gmail.com>
Date:   Mon Oct 26 07:57:52 2020 +0200

    Fix test_threads for non-linux environment:
    
     - check found thread_names adequacy, use num_threads()
     - take into account one extra thread on windows

diff --git a/src/tests/test_process.py b/src/tests/test_process.py
index cd114323..63aa2e9c 100644
--- a/src/tests/test_process.py
+++ b/src/tests/test_process.py
@@ -12,8 +12,6 @@ import unittest
 
 import psutil
 
-from sys import platform
-
 from pybitmessage.paths import set_appdata_folder
 
 
@@ -126,49 +124,36 @@ class TestProcessProto(unittest.TestCase):
             cls._cleanup_files()
 
     def _test_threads(self):
-        # only count for now
-        # because of https://github.com/giampaolo/psutil/issues/613
-        # PyBitmessage
-        #   - addressGenerator
-        #   - singleWorker
-        #   - SQL
-        #   - objectProcessor
-        #   - singleCleaner
-        #   - singleAPI
-        #   - Asyncore
-        #   - ReceiveQueue_0
-        #   - ReceiveQueue_1
-        #   - ReceiveQueue_2
-        #   - Announcer
-        #   - InvBroadcaster
-        #   - AddrBroadcaster
-        #   - Downloader
-        #   - Uploader
-
         self.longMessage = True
 
         try:
-            thread_names = \
-                subprocess.check_output(["ps", "-L", "-o", "comm=",
-                                         "--pid",
-                                         str(self.process.pid)]).split()
-        # pylint: disable=broad-except
-        except BaseException:
+            # using ps for posix platforms
+            # because of https://github.com/giampaolo/psutil/issues/613
+            thread_names = subprocess.check_output([
+                "ps", "-L", "-o", "comm=", "--pid",
+                str(self.process.pid)
+            ]).split()
+        except:  # pylint: disable=broad-except
             thread_names = []
 
-        extra_threads = []
-        missing_threads = []
-        for thread_name in thread_names:
-            if thread_name not in self._threads_names:
-                extra_threads.append(thread_name)
-        for thread_name in self._threads_names:
-            if thread_name not in thread_names:
-                missing_threads.append(thread_name)
-
-        msg = "Missing threads: {}, Extra threads: {}".format(
-            ",".join(missing_threads), ",".join(extra_threads))
-
-        running_threads = len(self.process.threads())
+        running_threads = len(thread_names)
+        if 0 < running_threads < 30:  # adequacy check
+            extra_threads = []
+            missing_threads = []
+            for thread_name in thread_names:
+                if thread_name not in self._threads_names:
+                    extra_threads.append(thread_name)
+            for thread_name in self._threads_names:
+                if thread_name not in thread_names:
+                    missing_threads.append(thread_name)
+
+            msg = "Missing threads: {}, Extra threads: {}".format(
+                ",".join(missing_threads), ",".join(extra_threads))
+        else:
+            running_threads = self.process.num_threads()
+            if sys.platform.startswith('win'):
+                running_threads -= 1  # one extra thread on Windows!
+            msg = "Unexpected running thread count"
 
         self.assertGreaterEqual(
             running_threads,
@@ -220,7 +205,6 @@ class TestProcess(TestProcessProto):
                 'Failed to read file %s' % pfile
             )
 
-    @unittest.skipIf(platform.startswith('darwin'), "Doesn't work on OSX")
     def test_threads(self):
         """Testing PyBitmessage threads"""
         self._test_threads()

```

g1itch: Almost clear: https://travis-ci.org/github/g1itch/PyBitmessage/builds/738955127

src/tests/test_process.py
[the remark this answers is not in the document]
PeterSurda: I added `os.path.exists`.

src/tests/test_process.py
PeterSurda: I added this now.

src/tests/test_process.py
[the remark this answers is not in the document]
g1itch: if not `os.path.exists` it will be full wait until `pybitmessage -t` stops. See:
https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/bitmessagemain.py#L180
https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/bitmessagemain.py#L371

src/tests/test_process.py
```diff
@@ -32,10 +54,27 @@ class TestProcessProto(unittest.TestCase):
...
+                                           'singleton.lock')):
+                try:
+                    pstat = os.stat(os.path.join(cls.home, 'singleton.lock'))
+                    if starttime <= pstat.st_mtime and pstat.st_size > 0:
+                        break
+                except OSError:
+                    break
+            time.sleep(1)
+        if time.time() >= timeout:
+            raise psutil.TimeoutExpired("Timeout starting {}".format(cls._process_cmd))
```
g1itch, line 74: 87 char line

g1itch (approved):
TImeout expired every time on osx: https://travis-ci.org/github/g1itch/PyBitmessage/jobs/739252746

src/tests/test_process.py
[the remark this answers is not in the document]
g1itch: Hmm https://travis-ci.org/github/g1itch/PyBitmessage/jobs/739258079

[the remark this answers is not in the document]
g1itch: no fractional part in `st_mtime`: https://travis-ci.org/github/g1itch/PyBitmessage/jobs/739262652

src/tests/test_process.py
```diff
@@ -32,10 +54,27 @@ class TestProcessProto(unittest.TestCase):
...
     def setUpClass(cls):
         """Setup environment and start pybitmessage"""
-        cls.home = os.environ['BITMESSAGE_HOME'] = tempfile.gettempdir()
+        cls.home = set_appdata_folder(tempfile.gettempdir())
+        cls._cleanup_files()
         put_signal_file(cls.home, 'unittest.lock')
+        starttime = time.time()
         subprocess.call(cls._process_cmd)  # nosec
-        time.sleep(5)
+        timeout = starttime + 30.0
```
g1itch, line 62: milliseconds?

src/tests/test_process.py
```diff
@@ -32,10 +54,27 @@ class TestProcessProto(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         """Setup environment and start pybitmessage"""
-        cls.home = os.environ['BITMESSAGE_HOME'] = tempfile.gettempdir()
+        cls.home = set_appdata_folder(tempfile.gettempdir())
+        cls._cleanup_files()
         put_signal_file(cls.home, 'unittest.lock')
+        starttime = time.time()
```
g1itch, line 60: maybe `int(time.time())`?

src/tests/test_process.py
```diff
@@ -32,10 +54,27 @@ class TestProcessProto(unittest.TestCase):
...
+        timeout = starttime + 30.0
+        while time.time() <= timeout:
+            if os.path.exists(os.path.join(cls.home,
+                                           'singleton.lock')):
+                try:
+                    pstat = os.stat(os.path.join(cls.home, 'singleton.lock'))
+                    if starttime <= pstat.st_mtime and pstat.st_size > 0:
+                        break
+                except OSError:
+                    break
```
g1itch, line 71: As I can see `OSError` raised when `os.path.exists(os.path.join(cls.home, 'singleton.lock')) is False`.

src/tests/test_process.py
```diff
@@ -32,10 +54,27 @@ class TestProcessProto(unittest.TestCase):
...
+            time.sleep(1)
+        if time.time() >= timeout:
+            raise psutil.TimeoutExpired("Timeout starting {}".format(cls._process_cmd))
+        # wait a bit for the program to fully start
+        # 10 sec should be enough
+        time.sleep(10)
```
g1itch, line 77: Why sleep 10 seconds if singleton.lock is already checked? What may happen?

g1itch:
No handlers could be found for logger "default": https://travis-ci.org/github/Bitmessage/PyBitmessage/builds/739215140#L1084

src/tests/test_logger.py
```diff
@@ -37,33 +39,44 @@ class TestLogger(unittest.TestCase):
...
 
-        self.assertEqual(logger, logger_)
+        logger1 = logging.getLogger('default')
 
-        logger_.info('Testing the logger...')
+        self.assertEqual(logger, logger1, ", pass {}".
+                         format(pass_))
 
-        self.assertRegexpMatches(open(log_file).read(), pattern)
+        logger2 = logging.getLogger('root')
```
g1itch, line 73: I don't understand the purpose, it's not used in code:
```shell
$ grep -r 'getLogger' src | grep root
src/tests/test_logger.py:        logger2 = logging.getLogger('root')
```

src/tests/test_logger.py
```diff
@@ -37,33 +39,44 @@ class TestLogger(unittest.TestCase):
...
+
+        logger2.info('Testing the logger...')
+        flush_logs()
+
+        self.assertRegexpMatches(open(log_file).read(), pattern,
+                                 ", pass {}".format(pass_))
+
+        if pass_ == 2:
+            os.remove(log_config)
```
g1itch, line 82: Why only on second pass?

src/tests/test_process.py
g1itch: OK, I see. It is to start all threads and not fail test_threads and test_shutdown.

src/tests/test_process.py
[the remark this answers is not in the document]
PeterSurda: Ok I'll look throught this.

src/tests/test_process.py
PeterSurda: I see.

src/tests/test_process.py
PeterSurda: Ok makes sense.

src/tests/test_process.py
PeterSurda: Oh, now I get it.

src/tests/test_logger.py
PeterSurda: I think it may be leftover from debugging, perhaps it isn't necessary anymore.

src/tests/test_logger.py
PeterSurda: I don't understand the logic flow of the passes, this may be a leftover from debugging and perhaps isn't necesasry anymore.

PeterSurda:
I'll work on the requested changes.

src/tests/test_logger.py
g1itch: This not fixes the test. I'll describe how it works, how it worked and how I wanted it to work a bit later.

PeterSurda: I'll look at it.

g1itch: > I don't understand the logic flow of the passes, this may be a leftover from debugging and perhaps isn't necesasry anymore.

A proper test should not rely on other tests side effects.

What I wanted to demonstrate in this particular test:
1. when you import `logger` from `debug` it sets up the logging based on `logging.dat` config
1. `logger = logging.getLogger('root')` set after the import will be the same as in `debug`
1. if you call `resetLogging()` the configuration will be reapplied

All that works only after the first import of `debug` because it calls `helper_startup.loadConfig()`.

How it worked before:
1. precondition: there is no other test importing `debug` or other module inducing config setup or appdata change
1. first pass detected by `debug.log` in appdata remaining from core tests
1. on the second pass you need `resetLogging()`

You can reveal it by replacing all `tempfile.gettempdir()` by `tempfile.mkdtemp()` as was suggested before.

g1itch: I'm not sure that I can precisely reconstruct what's happening now. It places `logging.dat` which is used by subsequent runs of pybitmessage process and can be detected on second pass.

g1itch: Maybe subclass this test case from `TestProcessProto` and check only the presence of pattern and maybe severity and initial message? Then we can replace `logging.getLogger('default')` by conventional `logging.getLogger(__name__)`, compose proper `logging.dat` and check presence of all submodule names which write logs.

src/tests/test_process.py
g1itch: But I hope that 5 sec will be enough for any environment (on travis it works without delay). This delay is used before starting tests in any test case subclassed from `TestProcessProto` on each pass.

g1itch:
I'm waiting to check the new approach: https://travis-ci.org/github/g1itch/PyBitmessage/builds/739605155

g1itch:
The cleanup of the core tests is [here](https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/tests/core.py#L51).

g1itch:
xenial, bionic and fedora31: OK, small hangup on focal, waiting for windows

https://travis-ci.org/github/g1itch/PyBitmessage/builds/739657803 - process based `TestLogger` and cleanup tweaks.

src/tests/test_logger.py
PeterSurda: > first pass detected by debug.log in appdata remaining from core tests

Only if it uses the same config directory. But the tests change the location of the config directory. I think this breaks the logic. You can see now that I reverted to your original logic in `test_logger.py`, and it fails on travis-ci.org. I don't want to merge until this is resolved.

PeterSurda:
I think all your feedback has been applied, and travis-ci passes. I think the rest of the issue can be resolved later, and this is ok to merge now. Let me know if you agree.

src/tests/test_logger.py
g1itch: I wrote before that my original logic was wrong. Standalone tests should not import any module with side effects.
I debug my process-based variant, it succedes only on linux: https://travis-ci.org/github/g1itch/PyBitmessage/builds/740134631

g1itch:
> I think all your feedback has been applied, and travis-ci passes. I think the rest of the issue can be resolved later, and this is ok to merge now. Let me know if you agree.

Now I don't see any help from `get_active_config_folder()` and related code.

PeterSurda:
> Now I don't see any help from get_active_config_folder() and related code.

It works on xenial on buildbot now.

PeterSurda:
So what now?

g1itch:
> > Now I don't see any help from get_active_config_folder() and related code.
> 
> It works on xenial on buildbot now.

But it you remove initial `src/bitmessagemain.py -t` it may stop working.

PeterSurda:
Well you can fix it later.

PeterSurda:
So apart from the logger do you have any other complaints?

g1itch:
> So apart from the logger do you have any other complaints?

While this version `TestLogger` is here any newly written test case may fail unexpectedly.

`get_active_config_folder()` code is doing nothing.

Also I have one more assertion which should not fail. I'll prepare another draft PR so you can test it on your buildbot.

PeterSurda:
> get_active_config_folder() code is doing nothing.

It fixes the `test_config.py` errors. The logger tests can be fixed in a separate PR.

g1itch:
> > get_active_config_folder() code is doing nothing.
> 
> It fixes the `test_config.py` errors. The logger tests can be fixed in a separate PR.

What fixes `test_config` is proper cleanup and guaranteed kill in `tearDownClass` - that's what I'm going to demonstrate in the new rebased branch.

PeterSurda:
Ok then I close this and you can make a PR with your changes.

PR #1675, opened 2020-09-25
debian package building added

navjotcis:
I have added debian package building with the reference of https://git.bitmessage.org/citizenaspirant/PyBitmessage/commit/5f955ba52f10c6167e82ec870cbb327a6ca1de2d

PR #1686, opened 2020-10-30
Full featured apt package using stdeb

g1itch:
As you suggested this PR is instead of #1675. Maybe squash more?


PeterSurda:
Yes please squash, and remove `Dockerfile` and `src/bitmessagemain.py` from the PR.

g1itch:
The build process was documented in bionic job of the example `.travis.yml` it's basically
```sh
sudo apt-get install -yq apt-file build-essential debhelper dh-python fakeroot libssl-dev python-all-dev \
    python-msgpack python-qt4 python-setuptools python-stdeb
python setup.py sdist
python setup.py --command-packages=stdeb.command bdist_deb
```

Using `export VERSION=$(python setup.py -V)` you may find the package in `deb_dist/pybitmessage_${VERSION}-1_amd64.deb`. Such name is because it's a program, not library. `python-<libname>_<version_arch>.deb` is for libraries.

PeterSurda:
Yes I'm fixing it.

g1itch:
> Yes please squash, and remove `Dockerfile` and `src/bitmessagemain.py` from the PR.

So you OK with the old deb building procedure in the Dockerfile? Changes to `bitmessagemain` are needed if you want to use `pybitmessage -t` when installed from deb.

PeterSurda:
Won't the `Dockerfile` work if you skip `pybitmessage -t`?

g1itch:
> Won't the `Dockerfile` work if you skip `pybitmessage -t`?

`pybitmessage -t` is used to generate default keys.dat. Then it generates password and edits the config.

PeterSurda:
I thought it was supposed to create the defaults if settingsversion isn't high enough, even if the api* options are already set.

g1itch:
Without installing extra packages and running core tests it's just builds faster.

PeterSurda:
It looks like an overkill to run `pybitmessage -t` just to create a default config file.

g1itch:
> It looks like an overkill to run `pybitmessage -t` just to create a default config file.

Otherwise user should connect to container upon first run and manually enable API and set username and password.

PeterSurda:
I think instead the bmconfigparser should set defaults when it's missing. Then you could create a minimal keys.dat and wouldn't have to run `pybitmessage -t`:

```
[bitmessagesettings]
settingsversion = auto
dontconnect = false
apiusername = aaa
apipassword = bbbb
apiinterface = 0.0.0.0
apienabled = true
```

PeterSurda:
I also don't understand why these two things (Dockerfile and dpkg) are in one PR.

g1itch:
> I also don't understand why these two things (Dockerfile and dpkg) are in one PR.

I removed dpkg use from Dockerfile because it was a documentation for building deb. I can squash those changes with the removal of previous package files.

PeterSurda (approved):
Ok approved, just fix pylint.

g1itch:
I forgot the apparmor config here.

PeterSurda (approved):
Ok

PR #1688, opened 2020-11-12
replace print statements with logger

### LINKED ISSUE #1373: No more prints

navjotcis:
I have replaced print statement with logger for below given module
`network/randomtrackingdict.py`
`singleinstance.py`

### DIFF
--- a/src/tests/test_randomtrackingdict.py
+++ b/src/tests/test_randomtrackingdict.py
@@ -1,0 +1,47 @@
+"""
+Tests for RandomTrackingDict class
+"""
+import random
+import unittest
+
+from network import randomtrackingdict
+from time import time
+
+
+class TestRandomTrackingDict(unittest.TestCase):
+    """Main protocol test case"""
+
+    @staticmethod
+    def randString():
+        """helper function for tests, generates a random string"""
+        retval = b''
+        for _ in range(32):
+            retval += chr(random.randint(0, 255))
+        return retval
+
+    def test_check_randomtrackingdict(self):
+        """Check the logic of RandomTrackingDict class"""
+        a = []
+        k = randomtrackingdict.RandomTrackingDict()
+
+        a.append(time())
+        for i in range(50000):
+            k[self.randString()] = True
+        a.append(time())
+
+        while k:
+            retval = k.randomKeys(1000)
+            if not retval:
+                self.fail("error getting random keys")
+
+            try:
+                k.randomKeys(100)
+                self.fail("bad")
+            except KeyError:
+                pass
+            for i in retval:
+                del k[i]
+        a.append(time())
+
+        for x in range(len(a) - 1):
+            self.assertLess(a[x + 1] - a[x], 10)
--- a/src/network/addrthread.py
+++ b/src/network/addrthread.py
@@ -3,11 +3,11 @@ Announce addresses as they are received from other hosts
 """
 import Queue
 
-import state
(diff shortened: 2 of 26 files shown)


PeterSurda (changes requested):
- the test in randomtrackingdict should be refactored as a unit test
- remove singleinstnace from PR
- filesystem is OK

PeterSurda (changes requested):
Move tests into unit tests.

PeterSurda:
- remove main section from `randomtrackingdict.py`
- fix errors reported by travis

src/tests/test_randomtrackingdict.py
```diff
@@ -0,0 +1,48 @@
...
+            try:
+                k.randomKeys(100)
+                print('bad')
+            except KeyError:
+                pass
+            for i in retval:
+                del k[i]
+        a.append(time())
+
+        for x in range(len(a) - 1):
```
PeterSurda, line 46: This is debugging information. We should check if this is not too slow. Let's say all values of `a[x + 1] - a[x]` need to be less than `10.0`.

PeterSurda (changes requested):
Please do the requested changes and fix code quality.

PeterSurda:
Also you can't write "Fixes #xxxxx" unless it's a complete fix, that screws up the dependency tracking.

navjotcis:
I have made the suggested changes not it is ready to merge.

src/tests/test_randomtrackingdict.py
```diff
@@ -0,0 +1,47 @@
...
+    @staticmethod
+    def randString():
+        """helper function for tests, generates a random string"""
+        retval = b''
+        for _ in range(32):
+            retval += chr(random.randint(0, 255))
+        return retval
+
+    def test_check_randomtrackingdict(self):
+        """Check the logic of RandomTrackingDict class"""
```
PeterSurda, line 23: `"""Test performance of RandomTrackingDict"""`

src/tests/test_randomtrackingdict.py
```diff
@@ -0,0 +1,47 @@
...
+                k.randomKeys(100)
+                self.fail("bad")
+            except KeyError:
+                pass
+            for i in retval:
+                del k[i]
+        a.append(time())
+
+        for x in range(len(a) - 1):
+            self.assertLess(a[x + 1] - a[x], 10)
```
PeterSurda, line 47: Maybe this needs to be `10.0`, I'm not fully sure, we'll see once the import is fixed.

PeterSurda (changes requested):
Please fix import.

src/network/__init__.py
```diff
@@ -8,6 +8,7 @@
 from downloadthread import DownloadThread
 from invthread import InvThread
 from networkthread import BMNetworkThread
+from .randomtrackingdict import RandomTrackingDict
```
PeterSurda, line 11: why dot at the beginning?

PeterSurda (changes requested):
fix

src/network/__init__.py
navjotcis: checking with every possible way

PR #1689, opened 2020-11-15
add pendingDownload and pendingUpload to return of API clientStatus

813492291816:
adds some information to clientStatus

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -1419,9 +1419,9 @@ class BMRPCDispatcher(object):
         """
         Returns the bitmessage status as dict with keys *networkConnections*,
         *numberOfMessagesProcessed*, *numberOfBroadcastsProcessed*,
-        *numberOfPubkeysProcessed*, *networkStatus*,
-        *softwareName*, *softwareVersion*. *networkStatus* will be one
-        of these strings: "notConnected",
+        *numberOfPubkeysProcessed*, *pendingDownload*, *networkStatus*,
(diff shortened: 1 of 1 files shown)


PeterSurda (changes requested):
Since `pendingUpload` isn't working correctly now I would remove it from the report. Other than that looks ok.

813492291816:
Thanks for the feedback. pendingDownload is really what I wanted, but figured I should add pendingUpload as well after seeing network.stats. I've gone ahead and removed pendingUpload.

PeterSurda (approved):

PeterSurda:
The commits aren't PGP signed. Please sign them and paste your PGP key into your GitHub account's settings.

813492291816:
Sure. I'm setting it up now and sign my last commit when finished.

813492291816:
My last commit has been signed and I will sign any future commits.

PeterSurda:
All comments must be signed. Maybe squash, then you can sign easier.

813492291816:
Thank you for pointing me in the right direction and your patience. I've squashed my two commits into one verified commit.

PeterSurda:
I introduced strict policies about merging for security reasons, so it's more difficult to contribute. However I'd like to update contribution guidelines for a step by step description about how to do that. Most of the checks are already automated so the contributors sees the result directly in the PR, the signatures aren't, but I'll add that too eventually.

PR #1691, opened 2020-11-25
Fix dependency checks for optional modules

### LINKED ISSUE #1316: checkdeps.py broken
```
  File "checkdeps.py", line 142, in <module>
    if x in EXTRAS_REQUIRE_DEPS
KeyError: None
```

navjotcis:
I have put import_module for checking already installed dependency in checkdeps module so with the use of it buildbot checks get 
improve and fix so issues of  #1316


PeterSurda (approved):
This isn't a full fix, just a partial. The description needs to be updated so that it doesn't auto-close the issue.

navjotcis:
Now you can merge it

PR #1694, opened 2020-12-08
Closes #1538. Refactors 'onionservicesonly' to 'onlynet=onion', prevents bootstrapping from non-onion nodes when true, and prevents bootstrapping from any node but the trusted peer (when set)

### LINKED ISSUE #1538: Add onlynet=onion config option

sgj3:
This PR fixes #1538.  It:
- Gracefully changes the config option 'onionservicesonly=true' to 'onlynet=onion', upgrading old config directives upon GUI settings dialog "OK"
- When onion services only is requested, prevents bootstrapping from a non-onion node
- When a trusted peer is set, prevents bootstrapping from any other node 

This PR *does not* add any warning or error dialogs, which would be useful for a user unknowingly unable to bootstrap because they set 'onlynet=onion'.

(Please ignore #1693. I accidentally hit <ENTER> when typing the title, which GitHub oh-so-helpfully interpreted as a click on "Create pull request".)

### DIFF
--- a/src/bitmessageqt/settings.py
+++ b/src/bitmessageqt/settings.py
@@ -358,6 +358,11 @@ class SettingsDialog(QtGui.QDialog):
         if proxytype_index > 2:  # last literal proxytype in ui
             start_proxyconfig()
 
+        onionOnly_deprecated = BMConfigParser().safeGetBoolean(
+            "bitmessagesettings", "onionservicesonly")
+        onionOnly = BMConfigParser().safeGet(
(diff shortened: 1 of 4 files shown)


PeterSurda:
I need more time to look at it more closely, but upon first glance it looks ok, just fix the code quality issues and keep it rebased so that I can merge it when ready.

PR #1697, opened 2020-12-19
test case for testing old knownnodes pickle

### LINKED ISSUE #1662: Pickle old knownnodes missing/incomplete error handler

navjotcis:
Added test case for old formate knownnodes pickle file and handle the exception so which will fix #1668 issue


PeterSurda:
- you can use a different extension for the sample file to avoid changing `.gitignore`
- you can copy the file instead of read/write (`shutil.copyfile`)
- the sample file is too big
- you need to add the fixes from #1668 (at least the one from `src/network/knownnodes.py`

src/tests/core.py
```diff
@@ -259,6 +260,30 @@ def test_insert_method_msgid(self):
...
+        """Testing old(v.0.6.2) version knownnodes.dat file"""
+        is_pass = False
+        try:
+            self.load_old_knownnodes_file()
+            knownnodes.readKnownNodes()
+            is_pass = True
+        except (AttributeError, Exception) as e:
+            print('Error: ', e)
+            self.cleanup()
+        self.assertTrue(is_pass)
```
PeterSurda, line 273: `self.cleanup` should be inside `finally:`

PeterSurda (changes requested):
The fix isn't implemented, this is only a test. Copy the fix from g1itch's PR.

src/tests/core.py
PeterSurda: Also no print.

src/tests/core.py
```diff
@@ -259,6 +260,30 @@ def test_insert_method_msgid(self):
...
+    def test_old_knownnodes_pickle(self):
+        """Testing old(v.0.6.2) version knownnodes.dat file"""
+        is_pass = False
+        try:
+            self.load_old_knownnodes_file()
+            knownnodes.readKnownNodes()
+            is_pass = True
+        except (AttributeError, Exception) as e:
+            print('Error: ', e)
+            self.cleanup()
```
g1itch, line 272: `cleanup('knownnodes.dat')` The cleanup function is imported from module `common`

g1itch: Also maybe it's better to use `self.fail('Failed to load knownnodes: %s' % e)` here instead of `self.assertTrue(is_pass)`? What other `Exception` do you expect, I don't understand.

src/tests/core.py
```diff
@@ -259,6 +260,30 @@ def test_insert_method_msgid(self):
...
+        self.assertTrue(is_pass)
+
+    @staticmethod
+    def load_old_knownnodes_file():
+        """loading old knownnodes file in .config directory"""
+        old_source_file = os.path.join(os.getcwd(), 'tests/test_pattern', 'knownnodes.dat')
```
g1itch, line 278: You probably can use `os.path.abspath(__file__)` here.

src/tests/core.py
```diff
@@ -259,6 +260,30 @@ def test_insert_method_msgid(self):
             '''select typeof(msgid) from sent where ackdata=?''', result)
         self.assertEqual(column_type[0][0] if column_type else '', 'text')
 
+    def test_old_knownnodes_pickle(self):
+        """Testing old(v.0.6.2) version knownnodes.dat file"""
+        is_pass = False
+        try:
+            self.load_old_knownnodes_file()
```
g1itch, line 267: just `shutl.copy(...` no need for extra method

navjotcis:
I have made the suggested changes now it is ready to merge

PeterSurda (approved):

PR #1698, opened 2020-12-22
test cases python3 porting initial

surbhicis:
setup file changes for python3 porting

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,17 @@ EXTRAS_REQUIRE = {
     'docs': ['sphinx', 'sphinxcontrib-apidoc', 'm2r']
 }
 
+if sys.version_info[0] == 2:
+    version_dependencies = {
+        'requirements_file': 'requirements.txt',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -61,8 +72,8 @@ def run(self):
     installRequires = []
     packages = [
         'pybitmessage',
-        'pybitmessage.bitmessageqt',
-        'pybitmessage.bitmessagecurses',
+        # 'pybitmessage.bitmessageqt',
```
PeterSurda, line 75: this will break things

setup.py
```diff
@@ -127,7 +137,7 @@ def run(self):
         package_dir={'pybitmessage': 'src'},
         packages=packages,
         package_data={'': [
-            'bitmessageqt/*.ui', 'bitmsghash/*.cl', 'sslkeys/*.pem',
+            'sslkeys/*.pem',
```
PeterSurda, line 140: this will break things

PeterSurda (changes requested):

PR #1699, opened 2020-12-29
Randomtrackingdict module code quality and unit test

navjotcis:
1) written test case for randomtrackingdict module
2) moved randomtrackingdict.py outside the network directory
3) replaced the print statement of filesystem.py file with logger for fixing some of the issues of #1373


g1itch:
>     2. moved randomtrackingdict.py outside the network directory

Why? Where it's used outside of the `network` module?

PeterSurda:
@g1itch It isn't used outside, but there is no network-specific code in it. More importantly, it is difficult to have unit tests executed correctly for stuff under `network`. Probably still some leftover circular imports or maintaining compatibility for running it without setuptools. I think once we have a sufficient number of distro-independent packaging procedures available (mainly appimage, perhaps snap), then we can revisit how to fix this properly and perhaps it can be moved to network again.

src/storage/filesystem.py
```diff
@@ -7,6 +7,7 @@
 from os import listdir, makedirs, path, remove, rmdir
 from threading import RLock
 
+from debug import logger
```
g1itch, line 10: bad practice, please read the `debug` doc

g1itch:
> @g1itch It isn't used outside, but there is no network-specific code in it. More importantly, it is difficult to have unit tests executed correctly for stuff under `network`.

Maybe, but I still don't understand why `from pybitmessage.network import ..` not works, but `from pybitmessage import ..` works.

PeterSurda:
> Maybe, but I still don't understand why ...

I also don't fully understand it, my guess is that the migration to setuptools isn't finished and some leftover code that expects code to be in the current directory rather than as a module. It needs to be fixed eventually but for now we'll have to work around it.

g1itch:
> > Maybe, but I still don't understand why ...
> 
> I also don't fully understand it, my guess is that the migration to setuptools isn't finished and some leftover code that expects code to be in the current directory rather than as a module. It needs to be fixed eventually but for now we'll have to work around it.

Oh, I see another import from outer package not working without pathmagick:
```
>>> from pybitmessage.network import randomtrackingdict
  File "<stdin>", line 1, in <module>
  File "pybitmessage/network/__init__.py", line 4, in <module>
    from addrthread import AddrThread
  File "pybitmessage/network/addrthread.py", line 6, in <module>
    import state
ImportError: No module named state
```

src/storage/filesystem.py
PeterSurda: I don't understand, what's wrong here? This is like it's used everywhere else in the project.

PeterSurda (commented):
Looks ok.

src/storage/filesystem.py
g1itch: Another import from outer package. When I refactored the logger I left the [note](https://pybitmessage.rtfd.io/en/v0.6/autodoc/pybitmessage.debug.html):
> The old form: from debug import logger is also may be used, but only in the top level modules.

PeterSurda: I see.

navjotcis: I have replaced debug with logging import

PeterSurda (approved):

PR #1701, opened 2021-01-01
SQL operations asserts and code quality

PeterSurda:
- complain if trying to execute SQL statements without a running `.threads.sqlThread`. This is to give better test feedback if used incorrectly
- refactor `.helper_sql.sql_ready` as a `threading.Event`
- code quality


g1itch (approved):
Good: minus one state variable.

PR #1708, opened 2021-01-08
Add missing TTL to API sendMessage

813492291816:
I noticed that the TTL was being set to 4 days when sendMessage was being called with the API, despite a TTL of a different time being used. It appears TTL was not incorporated into one of the latest changes. This PR should restore TTL-setting functionality.


PeterSurda (approved):

PeterSurda:
Thank you.

PR #1709, opened 2021-01-08
add helper_addressGenerator module

navjotcis:
created helper_addressGenerator.py file for using it into a test case

### DIFF
--- a/src/helper_addressGenerator.py
+++ b/src/helper_addressGenerator.py
@@ -1,0 +1,37 @@
+"""
+Create random address
+"""
+
+import time
+
+import defaults
+import queues
+import state
+
+from bitmessageqt import account
+from bitmessageqt.foldertree import AccountMixin
+from bmconfigparser import BMConfigParser
+
+
+def checkHasNormalAddress():
+    """method for checking address"""
+    for address in account.getSortedAccounts():
+        acct = account.accountClass(address)
+        if acct.type == AccountMixin.NORMAL and BMConfigParser().safeGetBoolean(address, 'enabled'):
+            return address
+    return False
+
+
+def createAddressIfNeeded(label_text, streamNumberForAddress=1):
+    """method for creating random address"""
+    if not checkHasNormalAddress():
+        queues.addressGeneratorQueue.put((
+            'createRandomAddress', 4, streamNumberForAddress,
+            label_text,
+            1, "", False,
+            defaults.networkDefaultProofOfWorkNonceTrialsPerByte,
+            defaults.networkDefaultPayloadLengthExtraBytes
+        ))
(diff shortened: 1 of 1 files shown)


src/helper_addressGenerator.py
```diff
@@ -0,0 +1,52 @@
...
+                BMConfigParser().get(
+                    y,
+                    'label'),
+                'utf-8').lower()))
+    return configSections
+
+
+def checkHasNormalAddress():
+    """method for checking address"""
+    for address in getSortedAccounts():
```
PeterSurda, line 34: no need for sorting, just `BMConfigParser().addresses`.

src/helper_addressGenerator.py
```diff
@@ -0,0 +1,52 @@
...
+        if BMConfigParser().safeGetBoolean(address, 'enabled'):
```
PeterSurda, line 35: you should also filter broadcasts and chans.

src/helper_addressGenerator.py
```diff
@@ -0,0 +1,52 @@
...
+    """method for creating random address"""
+    if not checkHasNormalAddress():
+        queues.addressGeneratorQueue.put((
+            'createRandomAddress', 4, streamNumberForAddress,
+            label_text,
+            1, "", False,
+            defaults.networkDefaultProofOfWorkNonceTrialsPerByte,
+            defaults.networkDefaultPayloadLengthExtraBytes
+        ))
+    while state.shutdown == 0 and not checkHasNormalAddress():
```
PeterSurda, line 50: Add another timeout in case something is wrong, so that it doesn't get stuck. In that case just return None or raise an exception.

PeterSurda (changes requested):

src/helper_addressGenerator.py
navjotcis: ok

src/helper_addressGenerator.py
```diff
@@ -0,0 +1,37 @@
...
+        queues.addressGeneratorQueue.put((
+            'createRandomAddress', 4, streamNumberForAddress,
+            label_text,
+            1, "", False,
+            defaults.networkDefaultProofOfWorkNonceTrialsPerByte,
+            defaults.networkDefaultPayloadLengthExtraBytes
+        ))
+    start_time = time.time()
+    while state.shutdown == 0 and not checkHasNormalAddress():
+        time.sleep(.2)
```
PeterSurda, line 34: I mean add another condition to the while cycle.

PeterSurda (changes requested):

PeterSurda (approved):

PR #1713, opened 2021-01-11
ignoring ValueError from proofofwork module

### LINKED ISSUE #895: I can not run Bitmessage when I double click it.
I run v0.6,  I can not run Bitmessage when I double click it.

debug.log :
2016-08-29 19:01:30,918 - ERROR - C PoW test fail.
  File "F:\Coin\FTC\pack\PyInstaller-2.1-0.6\bitmessagemain\build\bitmessagemain\out00-PYZ.pyz\proofofwork", line 194, in <module>
  File "F:\Coin\FTC\pack\PyInstaller-2.1-0.6\bitmessagemain\build\bitmessagemain\out00-PYZ.pyz\ctypes", line 365, in **init**
WindowsError: [Error 126] 
2016-08-29 19:01:30,918 - ERROR - C PoW test fail.
  File "F:\Coin\FTC\pack\PyInstaller-2.1-0.6\bitmessagemain\build\bitmessagemain\out00-PYZ.pyz\proofofwork", line 204, in <module>
  File "F:\Coin\FTC\pack\PyInstaller-2.1-0.6\bitmessagemain\build\bitmessagemain\out00-PYZ.pyz\ctypes", line 365, in **init**
WindowsError: [Error 126]

navjotcis:
ignoring ValueError from proofofwork module for fix #895

### DIFF
--- a/src/proofofwork.py
+++ b/src/proofofwork.py
@@ -316,6 +316,9 @@ def init():
             except Exception as e:
                 logger.error("Error: %s", e, exc_info=True)
                 bso = None
+        except Exception as e:
+            logger.error("Error: %s", e, exc_info=True)
+            bso = None
     else:
         try:
             bso = ctypes.CDLL(os.path.join(paths.codePath(), "bitmsghash", bitmsglib))
--- a/src/proofofwork.py
+++ b/src/proofofwork.py
@@ -291,7 +291,6 @@ def init():
     global bitmsglib, bmpow
 
     openclpow.initCL()
-
     if sys.platform == "win32":
         if ctypes.sizeof(ctypes.c_voidp) == 4:
             bitmsglib = 'bitmsghash32.dll'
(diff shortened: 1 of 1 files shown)


src/proofofwork.py
```diff
@@ -315,8 +313,8 @@ def init():
                 bmpow.restype = ctypes.c_ulonglong
                 _doCPoW(2**63, "")
                 logger.info("Successfully tested C PoW DLL (cdecl) %s", bitmsglib)
-            except:
-                logger.error("C PoW test fail.", exc_info=True)
+            except Exception as e:
+                logger.error("Error: %s", e, exc_info=True)
```
PeterSurda, line 317: add another `except Exception as e:` that handles the `try:` from line 299

src/proofofwork.py
```diff
@@ -305,8 +304,7 @@ def init():
             bmpow.restype = ctypes.c_ulonglong
             _doCPoW(2**63, "")
             logger.info("Successfully tested C PoW DLL (stdcall) %s", bitmsglib)
-        except:
-            logger.error("C PoW test fail.", exc_info=True)
+        except (ValueError, Exception) as e:
```
PeterSurda, line 307: The general exception handler needs to be separate, it needs to do something else. `ValueError` indicates that the DLL is there but doesn't accept the parameters. In that case, we should silently try different parameters. But if there is a general exception, it means the DLL may be missing, or corrupt, or some other issue, and it that case it doesn't make sense to test it with different parameters, but it makes sens to notify the user.

PeterSurda (changes requested):

PeterSurda (approved):

PR #1718, opened 2021-01-16
test case for openclpow module

navjotcis:
written test case for testing openclpow module functionality.

### DIFF
--- a/src/tests/test_openclpow.py
+++ b/src/tests/test_openclpow.py
@@ -8,17 +8,23 @@ import pybitmessage.openclpow as openclpow
 
 from struct import pack, unpack
 
+from test_process import TestProcessProto
 
-class TestOpenClPow(unittest.TestCase):
+
+class TestOpenClPow(TestProcessProto):
     """
     Main opencl test case
     """
 
-    @unittest.skipIf(not openclpow.libAvailable,
-                     "pyopencl is not installed")
+    @classmethod
+    def setUpClass(cls):
+        openclpow.initCL()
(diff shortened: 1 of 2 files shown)


PeterSurda (changes requested):
Good start, fix CQ, add `@skipTestIf` so that it skips it if opencl isn't avaialble.

src/tests/test_openclpow.py
```diff
@@ -0,0 +1,28 @@
...
+
+from struct import pack, unpack
+
+
+class TestOpenClPow(unittest.TestCase):
+    """
+    Main opencl test case
+    """
+
+    @unittest.skipIf(not openclpow.libAvailable,
```
PeterSurda, line 17: @unittest.skipIf(not openclpow.enabledGpus,
                                "No GPUs found / enabled")

PeterSurda (changes requested):
change the skip decorator, otherwise ok

PeterSurda:
Actually, now that I think about it, it's more complicated. You need to put `initCL` into `setupClass`.

navjotcis:
in which module `setupClass` is present?

PeterSurda (approved):

g1itch:
I don't understand why is this test case subclassed from TestProcessProto.

PeterSurda:
@g1itch You're right. I'll request a fix.

g1itch:
> @g1itch You're right. I'll request a fix.

ec6f078

PR #1723, opened 2021-01-27
Make it possible to disable UDP announcing in settings

g1itch:
Hello!

UDP announcing is useless without LAN. Users with unstable wi-fi or mobile broadband connection are probably experiencing 
continuous log warnings about `socket.error` even if they have no bitmessage nodes in their LAN. It may be annoying so I suggest add an option to disable UDP announcing in settings.

### DIFF
--- a/src/bitmessageqt/tests/settings.py
+++ b/src/bitmessageqt/tests/settings.py
@@ -1,0 +1,34 @@
+import threading
+import time
+
+from main import TestBase
+from bmconfigparser import BMConfigParser
+from bitmessageqt import settings
+
+
+class TestSettings(TestBase):
+    """A test case for the "Settings" dialog"""
+    def setUp(self):
+        super(TestSettings, self).setUp()
+        self.dialog = settings.SettingsDialog(self.window)
+
+    def test_udp(self):
+        """Test the effect of checkBoxUDP"""
+        udp_setting = BMConfigParser().safeGetBoolean(
+            'bitmessagesettings', 'udp')
+        self.assertEqual(udp_setting, self.dialog.checkBoxUDP.isChecked())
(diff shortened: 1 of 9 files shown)


PeterSurda (approved):
Looks OK to me

src/bitmessageqt/tests/settings.py
```diff
@@ -0,0 +1,34 @@
...
+        """Test the effect of checkBoxUDP"""
+        udp_setting = BMConfigParser().safeGetBoolean(
+            'bitmessagesettings', 'udp')
+        self.assertEqual(udp_setting, self.dialog.checkBoxUDP.isChecked())
+        self.dialog.checkBoxUDP.setChecked(not udp_setting)
+        self.dialog.accept()
+        self.assertEqual(
+            not udp_setting,
+            BMConfigParser().safeGetBoolean('bitmessagesettings', 'udp'))
+        time.sleep(5)
```
PeterSurda, line 25: It's a good practice to avoid `time.sleep` in tests, but unfortunately I don't have a good alternative for the time being.

PeterSurda (approved):
I don't like the `sleep`s, it introduces non-deterministic behaviour and increases the time needed to run the tests. However, as I said, I don't have a good alternative now, other than a lot of refactoring.

src/bitmessageqt/tests/settings.py
g1itch: So why `QtTest` has it's own [qSleep()](https://www.riverbankcomputing.com/static/Docs/PyQt5/api/qttest/qtest.html#qWait) if it's considered a bad practice? In general I agree, refactoring of the code and improving tests will take some time.

Issue #1711, opened 2021-01-09
Visual tests


PeterSurda:
The GUI tests can be extended with something like https://github.com/cgoldberg/xvfbwrapper to produce a video. This can the be made available by buildbot for watching.

PR #1738, opened 2021-02-22
Screenshots in bitmessageqt tests

g1itch:
Hello!

This may address a part of #1711: not video but a working method to take screenshots. I also added `xvfbwrapper` into tests requirements for Linux and returned platform independent form `python src/bitmessagemain.py -t` to run core tests.


PeterSurda:
This is interesting. How about this experiment, using ffmpeg: https://gist.github.com/t-io/9e743ee070f8d2b693402039b1809174

g1itch:
> This is interesting. How about this experiment, using ffmpeg: https://gist.github.com/t-io/9e743ee070f8d2b693402039b1809174

OK, I'll try to adapt it.

PR #1739, opened 2021-03-01
Graphical unit test case for python 3 kivy desktop app

navjotcis:
written graphical unit test cases for kivy desktop app using telenium library


g1itch:
What is it doing here? There is no even such package `bitmessagekivy`. If it was, shouldn't in be placed inside it's dir, not into ` src/GraphicalUnitTest`? I didn't read your `telenium.tests`, but it doesn't look like unit tests for me - it's more like a demo program.

navjotcis:
asdasdasd

PR #1742, opened 2021-03-16
fix configparser import error for py3

cis-muzahid:
fix configparser import error for py3

### DIFF
--- a/src/bmconfigparser.py
+++ b/src/bmconfigparser.py
@@ -2,13 +2,22 @@
 BMConfigParser class definition and default configuration settings
 """
 
-import ConfigParser
+import sys
+if sys.version_info[0] == 3:
+    # python 3
+    import configparser as ConfigParser
+    SafeConfigParser = ConfigParser.ConfigParser
+else:
+    # python 2
+    import ConfigParser
+    SafeConfigParser = ConfigParser.SafeConfigParser
+
+import state
+from singleton import Singleton
 import os
 import shutil
 from datetime import datetime
 
-import state
-from singleton import Singleton
 
 BMConfigDefaults = {
     "bitmessagesettings": {
@@ -43,7 +52,8 @@ BMConfigDefaults = {
 
 
 @Singleton
-class BMConfigParser(ConfigParser.SafeConfigParser):
+class BMConfigParser(SafeConfigParser):
+
     """
     Singleton class inherited from :class:`ConfigParser.SafeConfigParser`
     with additional methods specific to bitmessage config.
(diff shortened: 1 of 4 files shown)


PeterSurda:
you need to enable configparser test for python3

PeterSurda (changes requested):
Please make the requested changes, there appear to be debugging code leftover.

src/bmconfigparser.py
```diff
@@ -73,8 +83,9 @@ def get(self, section, option, raw=False, variables=None):
             return ConfigParser.ConfigParser.get(
                 self, section, option, True, variables)
         except ConfigParser.InterpolationError:
-            return ConfigParser.ConfigParser.get(
-                self, section, option, True, variables)
+            return ConfigParser.ConfigParser.get(self, section, option, True, variables)
+        except TypeError:
+            return ConfigParser.ConfigParser.get(section, option, variables)
```
PeterSurda, line 88: This could also throw an exception, if the default is missing, so it's incomplete. Why not use `super` instead, wouldn't that make it easier to write code that runs on both python2 and python3?

src/bmconfigparser.py
```diff
@@ -91,9 +102,10 @@ def setTemp(self, section, option, value=None):
     def safeGetBoolean(self, section, field):
         """Return value as boolean, False on exceptions"""
         try:
+            # fallback = _UNSET, ** kwargs
```
PeterSurda, line 105: remove

src/bmconfigparser.py
```diff
@@ -91,9 +102,10 @@ def setTemp(self, section, option, value=None):
...
             return self.getboolean(section, field)
         except (ConfigParser.NoSectionError, ConfigParser.NoOptionError,
-                ValueError, AttributeError):
+                ValueError, AttributeError, TypeError):
```
PeterSurda, line 108: When does it raise `TypeError`? Isn't that a bug?

src/bmconfigparser.py
```diff
@@ -102,7 +114,7 @@ def safeGetInt(self, section, field, default=0):
         try:
             return self.getint(section, field)
         except (ConfigParser.NoSectionError, ConfigParser.NoOptionError,
-                ValueError, AttributeError):
+                ValueError, AttributeError, TypeError):
```
PeterSurda, line 117: also when `TypeError`?

PeterSurda (changes requested):

PeterSurda (approved):

src/tests/test_config_functional.py
```diff
@@ -0,0 +1,39 @@
+"""
+Various tests for config
+"""
+
+import os
+import unittest
+import tempfile
+
+from .test_process import TestProcessProto
```
PeterSurda, line 9: remove

src/tests/test_config_functional.py
```diff
@@ -0,0 +1,39 @@
...
+from pybitmessage.bmconfigparser import BMConfigParser
+
+
+class TestProcessConfig(TestProcessProto):
+    """A test case for keys.dat"""
+    home = tempfile.mkdtemp()
+
+    def test_config_defaults(self):
+        """Test settings in the generated config"""
+        self._stop_process()
```
PeterSurda, line 19: remove

src/tests/test_config_functional.py
```diff
@@ -0,0 +1,39 @@
...
+        self._kill_process()
```
PeterSurda, line 20: remove

PeterSurda (changes requested):
Remove the lines related to TestProcessProto, these prevent the moved tests to run in python3

src/tests/test_config_functional.py
cis-muzahid: `TestProcessProto` is required for `TestProcessConfig` class

PeterSurda: Sorry I was wrong. Please put the stop process / kill process back. Also, rename this file to `test_config_process.py`. That should be enough.

PeterSurda (changes requested):

PeterSurda (approved):

Issue #1712, opened 2021-01-10
python 3 port plan


PeterSurda:
Using the principles of continuous delivery, port to python 3 should progress as following:

- the first PR will add allow travis to run python3 while successfully doing nothing. The `pybitmessage` binary will check if it's running python 3 and if yes, exit with return code 0. Unit tests will be decorated to skip all tests when using python 3. This should allow travis to succeed in python 3. Not sure what to do if requirements.txt diverge between python 2 and python 3.
- gradually tests will be updated to allow running in python 3
- maybe new tests will be written to cater for areas that currently don't have tests to facilitate gradual porting
- once everything is working correctly in python 3, as well as binary builds for the main platforms, python 2 support will be dropped

g1itch:
> the first PR will add allow travis to run python3 while successfully doing nothing. The pybitmessage binary will check if it's running python 3 and if yes, exit with return code 0. Unit tests will be decorated to skip all tests when using python 3. This should allow travis to succeed in python 3. Not sure what to do if requirements.txt diverge between python 2 and python 3.

In my `ci` branch I [have](https://github.com/g1itch/PyBitmessage/blob/ci/.travis.yml#L20) `python "3.6"` in `allow_failures` and build itself succeeds if python3 part fails. But I we all see the failures.

g1itch:
> Not sure what to do if requirements.txt diverge between python 2 and python 3.

Unlikely

PeterSurda:
> In my `ci` branch I [have](https://github.com/g1itch/PyBitmessage/blob/ci/.travis.yml#L20) `python "3.6"` in `allow_failures` and build itself succeeds if python3 part fails. But we all see the failures.

This is interesting, unfortunately it doesn't help me here. I need something that allows a gradual compatibility with python 3 with as little human intervention as possible. If all the errors are just ignored, then a PR reviewer needs more time to figure out what's working and what not by looking at the detailed output. The creator of the PR also needs to spend more time evaluating whether he still needs to continue working or wait for a review. Whereas in my proposal, a failure means the PR is rejected and the PR creator has immediate feedback that the isn't finished, whereas a success means it's ready for a review. This way, PR after PR, python 3 support will be implemented.

[...]

PR #1389, opened 2018-11-05
Porting bitmessageqt to Qt5

g1itch:
Hello!

I realized that column width problem in `QTableWidget` was actually the [Qt bug](https://bugreports.qt.io/browse/QTBUG-67131).

I mostly finished work in this branch though it still needs some testing.

PR #1745, opened 2021-03-21
Python 2 -> 3 changes.

iljah:
Didn't see much work being done lately for transition from python 2 to 3 and was wondering if something like this would be helpful. If yes I can submit request(s) that follow guidelines better...


iljah:
@PeterSurda in case github doesn't ping you automatically.

PeterSurda:
Thanks, maybe we can use parts of it. As it is unfortunately I can't accept it as none of the test jobs succeed, it breaks compatibility with python 2.7 while still not working correctly with 3.7. You can run the tests yourself with `run-tests-in-docker.sh` if you want, and you can register with buildbot so that the tests are run on my servers, as described here: https://wiki.bitmessage.org/index.php/Contribute#Register_with_buildbot

iljah:
> breaks compatibility with python 2.7
This doesn't surprise me since I haven't been able to set up everything required by bitmessage under python 2 in latest fedora and cannot test compatibility of my changes. But if you're open to these contributions and no one else is working on this at the moment I can test python 2 stuff in a virtual machine so that things don't break.

PeterSurda:
@iljah well, there are developers working on python3 port, it just isn't always visible here as the work may not be ready for a PR until it's done correctly and I'm happy with it. I worry perhaps you're doing duplicate work.

If you want to continue development, as I said you can run the tests with the `run-tests-in-docker.sh`, it executes both python 2.7 and python 3.7 tests, no need to install a VM. I tested the script on Ubuntu 20.04 and Mac OS Big Sur, I'm confident it runs anywhere docker is available. Or, you can register with buildbot, and then upon pushing to your own repo, buildbot will queue the tests to be run by my workers, and you'll have the results online in about 5 minutes (this actually runs more tests, including windows builds, code quality, ...).

src/bitmessagemain.py
```diff
@@ -42,10 +42,15 @@
...
-)
+from network.addrthread import AddrThread
+from network.connectionpool import BMConnectionPool
+from network.dandelion import Dandelion
+from network.announcethread import AnnounceThread
+from network.networkthread import BMNetworkThread
+from network.invthread import InvThread
+from network.receivequeuethread import ReceiveQueueThread
+from network.downloadthread import DownloadThread
+from network.uploadthread import UploadThread
```
g1itch, line 53: You are binding the code to pathmagic harder instead of preparing its removal ):

`network` is a package, so you can write `from .network import ..` ant it will work both in python2.7 and python3 in installed and portable mode.

PeterSurda:
Also, there is WIP to get PyBitmessage flatpak builds working, #1618 . This may also alleviate pressure from a python3 port and serve as a workaround for new releases of linux distros.

src/singleinstance.py
```diff
@@ -75,7 +75,7 @@ def lock(self):
                     fcntl.lockf(self.fp, fcntl.LOCK_EX | fcntl.LOCK_NB)
                 self.lockPid = os.getpid()
             except IOError:
-                print 'Another instance of this application is already running'
+                print('Another instance of this application is already running')
                 sys.exit(-1)
```
g1itch, line 79: `sys.exit('Another instance of this application is already running')`?

src/storage/__init__.py
```diff
@@ -0,0 +1 @@
+from .storage import InventoryItem, InventoryStorage
```
g1itch, line 1: You are adding `__init__` here and remove it from `network`, using dotted imports here, but rely on pathmagic in other places. So inconsistent.

src/tr.py
```diff
@@ -40,12 +40,12 @@ def translateText(context, text, n=None):
         try:
```
g1itch, line 40: I rewritten it 2 years ago ): https://github.com/Bitmessage/PyBitmessage/pull/1389/files?file-filters%5B%5D=.py#diff-a804aa473dd72c3caf797608b9abc7931b439092f114a9e6b60e9a2aaf5fccc4

src/bitmessagemain.py
iljah: Thanks it wasn't very clear to me what the best way to fix imports for python 3 would be. I'll give .network a try.

src/singleinstance.py
iljah: Possible but in that case return value can't be specified. If nothing relies on particular return values then exit('') is probably better.

src/storage/__init__.py
iljah: I've (luckily? :)) never had to deal with this part of python so don't know what would be best/correct/portable fix. I'll try .stuff

src/singleinstance.py
g1itch: Only two values are significant here: 0 and any other. When you pass a string it's interpreted as program error, like any non-zero exit code.

src/tr.py
iljah: Didn't know this existed. Do you suggest starting my work from that or incorporating some changes into my fork?

src/singleinstance.py
iljah: I guess nothing relies on specific values of non-zero so exit('') is fine.
```
$ python -c 'exit(5)'; echo $?
5
$ python -c 'exit(6)'; echo $?
6
$ python -c 'exit("error")'; echo $?
error
1
```

src/tr.py
g1itch: You can see all the forks on the network page: https://github.com/Bitmessage/PyBitmessage/network I suggest you to keep trying if you wanna merge, but be prepared to long wait. Existing PR's and issues are also worth reading, for example #1712

iljah:
Created https://github.com/Bitmessage/PyBitmessage/pull/1746 that includes fewer changes but still works using python 2.

PR #1746, opened 2021-03-23
Python 2 -> 3 part 1

iljah:
Few changes to support python 3. Supercedes https://github.com/Bitmessage/PyBitmessage/pull/1745. Works with python 2.7.15 in fedora 29.

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -58,7 +58,11 @@ For further examples please reference `.tests.test_api`.
 """
 
 import base64
-import ConfigParser
+try:
+	import ConfigParser
+except ModuleNotFoundError:
+	import configparser as ConfigParser
+	ConfigParser.SafeConfigParser = ConfigParser.ConfigParser
 import errno
 import hashlib
 import httplib
--- a/src/bitmessagecurses/__init__.py
+++ b/src/bitmessagecurses/__init__.py
@@ -10,7 +10,11 @@ Bitmessage commandline interface
 #     * python2-pythondialog
 #  * dialog
 
-import ConfigParser
+try:
+	import ConfigParser
+except ModuleNotFoundError:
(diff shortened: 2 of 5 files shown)


iljah:
Not sure if I should remove check for python 3 in https://github.com/Bitmessage/PyBitmessage/pull/1746/commits/7c66577671115614753c37c62ea12b8a267c88fa before everything works

g1itch:
> Not sure if I should remove check for python 3 in [7c66577](https://github.com/Bitmessage/PyBitmessage/commit/7c66577671115614753c37c62ea12b8a267c88fa) before everything works

You shouldn't. Again, please read #1712.

g1itch:
Maybe you begin with some tests? Separate `TestConfig` from `TestProcessConfig` like I did here: d947e70 because the later wont work before complete port to python3 (it uses subprocess to call `pybitmessage -t`) and write more, e.g. 9808968.

It looks like in python3 you can reuse defaults from `BMConfigDefaults` and then `getint()`, `getboolean()` will act in the same way as `safeGetInt()` and `safeGetBoolean()` in current code.

Porting `BMConfigParser()` is the next big thing in python3 porting process, so I'd suggest to separate related changes in dedicated branch. You may also mock `BMConfigParser()` and write tests and changes for some other parts of code.

g1itch:
Notice the required checks

iljah:
> Maybe you begin with some tests?

Could work. I've started that in https://github.com/iljah/PyBitmessage/tree/py3-tests1 and once I get bit further I'll submit another pull request.

iljah:
After https://github.com/Bitmessage/PyBitmessage/pull/1749 I start hitting error that are fixed by stuff here so I guess I'll wait until this is merged until continuing work on above test fixes

PR #1747, opened 2021-03-23
Handle missing settingsversion.

iljah:
Otherwise I get this with python 3.8.7 when starting BM: TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType' (assuming changes in https://github.com/Bitmessage/PyBitmessage/pull/1746)

### DIFF
--- a/src/helper_startup.py
+++ b/src/helper_startup.py
@@ -128,6 +128,8 @@ def loadConfig():
 def updateConfig():
     """Save the config"""
     config = BMConfigParser()
+    if not config.has_option('bitmessagesettings', 'settingsversion'):
+        config.set('bitmessagesettings', 'settingsversion', 1)
     settingsversion = config.getint('bitmessagesettings', 'settingsversion')
     if settingsversion == 1:
         config.set('bitmessagesettings', 'socksproxytype', 'none')


src/helper_startup.py
```diff
@@ -128,6 +128,8 @@ def loadConfig():
 def updateConfig():
     """Save the config"""
     config = BMConfigParser()
+    if not config.has_option('bitmessagesettings', 'settingsversion'):
+        config.set('bitmessagesettings', 'settingsversion', 1)
     settingsversion = config.getint('bitmessagesettings', 'settingsversion')
```
PeterSurda, line 133: Better to just add a default value into `src/bmconfigparser.py`

PeterSurda (changes requested):
Look at the `BMConfigDefaults` dict inside `src/bmconfigparser.py`.

src/helper_startup.py
iljah: Adding it like this at least doesn't help:
```
    "bitmessagesettings": {
        "settingsversion": 1,
        "maxaddrperstreamsend": 500,
```

PeterSurda: Try using `safeGetInt` instead of `getint` in addition to adding the default value into the dict.

iljah: switching to safegetint makes no difference whether settingsversion is included in BMConfigDefaults or not. No idea what the underlying cause is but my proposed fix adds only two lines...

iljah: The code you give is one of the changes I tried with stated effect. To me looks like only error from tests is in Ui_MainWindow which isn't related to this as far as I can see. If you want me to make additional changes you'll have to be more specific than "rest the tests".

PR #1749, opened 2021-03-25
Couple of changes to tests to support python3

iljah:
After these changes tests start to complain about things fixed by https://github.com/Bitmessage/PyBitmessage/pull/1746

### DIFF
--- a/src/tests/test_crypto.py
+++ b/src/tests/test_crypto.py
@@ -73,7 +73,8 @@ class TestCrypto(RIPEMD160TestCase, unittest.TestCase):
     """RIPEMD160 test case for Crypto"""
     @staticmethod
     def _hashdigest(data):
-        return RIPEMD.RIPEMD160Hash(data).digest()
+        from pybitmessage.fallback import RIPEMD160Hash
+        return RIPEMD160Hash(data).digest()
 
 
 class TestAddresses(unittest.TestCase):
--- a/src/tests/test_api.py
+++ b/src/tests/test_api.py
@@ -60,9 +60,15 @@ class TestAPIShutdown(TestAPIProto, TestProcessShutdown):
 
 class TestAPI(TestAPIProto):
     """Main API test case"""
-    _seed = base64.encodestring(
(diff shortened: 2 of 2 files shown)


g1itch:
`test_api` is irrelevant to python3 because it uses `TestProcess`. It's skipped in python3 job. There are more warnings related to bytes in travis job log for `test_blindsig`.

src/tests/test_crypto.py
```diff
@@ -73,7 +73,8 @@ class TestCrypto(RIPEMD160TestCase, unittest.TestCase):
     """RIPEMD160 test case for Crypto"""
     @staticmethod
     def _hashdigest(data):
-        return RIPEMD.RIPEMD160Hash(data).digest()
+        from pybitmessage.fallback import RIPEMD160Hash
+        return RIPEMD160Hash(data).digest()
```
g1itch, line 77: You didn't get it. This part of the test should use `pycrypto` or `pycryptodome`, not fallback. Because it's the test for fallback itself.

g1itch (changes requested):
These changes does not affect anything as I can see. It's just minus one test case.

src/tests/test_crypto.py
iljah: Comment stating this would've been helpful.

iljah:
> test_api is irrelevant to python3

This get me past complaints when using python3 so looks like it's not irrelevant.

g1itch:
> > test_api is irrelevant to python3
> 
> This get me past complaints when using python3 so looks like it's not irrelevant.

You even didn't bother to read #1712 and trying to edit the code you don't understand ): `test_api` is skipped here: https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/tests/test_process.py#L18

iljah:
> You even didn't bother to read #1712

I've been aware of it for a while but if you want more specific changes you'll have to provide more specific suggestions.

iljah:
> trying to edit the code you don't understand

I'm not using skip_python3() event though it's still included here. Same as in #1712

Issue #1571, opened 2019-11-21
suppress printf somehow


dimyme:
https://github.com/Bitmessage/PyBitmessage/blob/49d731c47865d8895df7cbf0e08312bd8aba289e/src/bitmsghash/bitmsghash.cpp#L124

can this screen output be suppressed ? it messes up the screen in  --curses mode.

PeterSurda:
Duplicate #1107

PR #1757, opened 2021-04-29
Log bmpow output

g1itch:
This is a possible solution for #1571. I need to write one more test for `LogOutput()`.


g1itch:
This is another bug ): https://buildbot.bitmessage.org/#/builders/23/builds/1361 The `.so` should be already compiled in the install stage and loaded there.

PeterSurda (approved):
Looks ok.

PR #1759, opened 2021-04-30
Quickfix for windows snapshots

g1itch:
It looks like PyInstaller doesn't see the `multiqueue` import when it's written in such weird way. With these changes the snapshots should work again, I tested in wine.

### DIFF
--- a/src/queues.py
+++ b/src/queues.py
@@ -1,19 +1,17 @@
 """Most of the queues used by bitmessage threads are defined here."""
 
-import sys
-if sys.version_info[0] == 3:
-    import queue as Queue
-else:
-    import Queue
-
 import threading
 import time
+try:
+    import queue as Queue
+except ImportError:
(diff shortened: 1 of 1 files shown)


PeterSurda:
Thank you.

PeterSurda (approved):

g1itch:
What do you think, maybe add another job for minimal testing exe with wine? At least it doesn't break your buildbot tasks. Or you can add the dry run into buildbot/PyBitmessage-wine32.

PeterSurda:
@g1itch Since a couple of days, the buildbot config is stored in git. If you create an account at git.bitmessage.org, I can give you access to buildbot config repo as well as build script repo. Then you can make PRs to add new jobs or modify existing ones. The config isn't rolled out automatically yet at the moment, but at least it's on git, so that I can roll it simply.

PR #1760, opened 2021-05-05
Refactor sqlthread

cis-muzahid:


stdeb.cfg
```diff
@@ -1,7 +1,7 @@
 [DEFAULT]
 Package: pybitmessage
 Section: net
-Build-Depends: dh-python, libssl-dev, python-all-dev, python-setuptools
+Build-Depends: dh-python, libssl-dev, python-all-dev, python-setuptools, python-six
```
PeterSurda, line 4: this should be a separate PR

src/helper_startup.py
```diff
@@ -29,6 +25,8 @@
 except ImportError:
     get_plugin = None
 
+# pylint: disable=too-many-branches,too-many-statements
+
```
PeterSurda, line 29: is this a code quality only change? Then it should go into a separate PR.

PeterSurda (changes requested):
Too big, can't review. Needs to be split. Separate code quality, support functions, and the individual levels.

stdeb.cfg
g1itch: It is 16a1177 merged from v0.6.

cis-muzahid: Is it fine now?

PeterSurda:
The upgrade design is too complicated. There is no need for a separate function per level.

src/class_sqlThread.py
```diff
@@ -29,22 +18,120 @@
...
+        self.max_level = 11
+
+        # call upgrading level in loop
+        for l in range(self.current_level, self.max_level):
+            if int(l) == 3:
+                continue
+            self.upgrade_one_level(l)
+            self.run_migrations(l)
+
+    def upgrade_schema_data_level(self, level):
```
PeterSurda, line 114: ```def increment_settings_version``

src/class_sqlThread.py
```diff
@@ -29,22 +18,120 @@
...
+
+        if level != self.get_current_level():
+            return None
+
+        # Migrate Db with level
+        method_name = 'upgrade_schema_data_' + str(level)
+        method = getattr(self, method_name, lambda: "Invalid version")
+        return method()
+
+    def run_migrations(self, file_name):
```
PeterSurda, line 80: ```def _upgrade_one_level_sql_statement```

src/class_sqlThread.py
```diff
@@ -29,22 +18,120 @@
...
+    """
 
-class sqlThread(threading.Thread):
+    conn, cur = connection_build()
+
+    def __init__(self):
+        self.current_level = None
+        self.max_level = 11
+
+    def get_current_level(self):
```
PeterSurda, line 58: ```def __get_current_settings_version```
Maybe this can't be private, that's ok.

src/class_sqlThread.py
```diff
@@ -29,22 +18,120 @@
...
+        """
+            Upgrade Db with respect to their versions
+        """
+        item = '''SELECT value FROM settings WHERE key='version';'''
+        parameters = ()
+        self.cur.execute(item, parameters)
+        return int(self.cur.fetchall()[0][0])
+
+    def upgrade_one_level(self, level):
```
PeterSurda, line 67: ```def _upgrade_one_level_method```

PeterSurda (changes requested):

src/class_sqlThread.py
```diff
@@ -102,11 +189,20 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
         settingsversion = BMConfigParser().getint(
             'bitmessagesettings', 'settingsversion')
 
+        # Process SQL queries first20bytesofencryptedmessage in table inventory
```
PeterSurda, line 192: this section can probably be deleted

src/class_sqlThread.py
```diff
@@ -102,11 +189,20 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+        if settingsversion == 1:
+            item = '''ALTER TABLE inventory ADD first20bytesofencryptedmessage blob DEFAULT '' '''
+            parameters = ()
+            self.cur.execute(item, parameters)
+            item = '''update settings set value=? WHERE key='version';'''
+            parameters = (2,)
+            self.cur.execute(item, parameters)
+
         # People running earlier versions of PyBitmessage do not have the
```
PeterSurda, line 201: lines 201-227 can be made into a separate method

PeterSurda: lines 227-292 can be split into 2 or 3 methods.

cis-muzahid: Done, please check the method name.

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Done

src/tests/test_sqlthread.py
```diff
@@ -52,11 +68,42 @@ def initialise_database(self, file):
...
+
+            # Update versions DB mocking
+            self.initialise_database("init_version_{}".format(version))
+
+            if int(version) == 9:
+                sqlThread().create_function()
+
+            # Test versions
+            upgrade_db = UpgradeDB()
+            upgrade_db._upgrade_one_level_sql_statement(int(version))
```
PeterSurda, line 97: ```  # pylint: disable=protected-access```

src/class_sqlThread.py
```diff
@@ -29,22 +18,120 @@
...
+            self._upgrade_one_level_sql_statement(l)
+
+    def increment_settings_version(self, level):
+        """
+            Update version with one level
+        """
+        item = '''update settings set value=? WHERE key='version';'''
+        parameters = (level + 1,)
+        self.cur.execute(item, parameters)
+
```
PeterSurda, line 121: this should be more transparent. For example, you can use a `@property` decorator for properties `sql_schema_version` and `configparser_schema_version` and use these instead calling the methods directly. And setter can enforce a `+1` irrespective of what you pass as an argument.

PeterSurda:
- rebase
- make the versions into properties
- rename `versioning` into something else, inside the methods you can use `version`, as a decorator perhaps `sql_schema_upgrade`
- delete obsolete stuff
- I haven't thoroughly reviewed tests but for starters they are probably ok
- rename the version 1 init/test to version 3. I checked the original source and it may break things if you're upgrading from version 1. If you just rename it to 3, then it would only break if we stopped the upgrade at version 2, which isn't a practical problem. You just need to make sure version 2 and 3 tests still succeeds.

PeterSurda (changes requested):

PeterSurda:
It is certainly an improvement but still work left to do.

cis-muzahid:
Closing because having changes in #1794

PR #1766, opened 2021-05-27
Added testcase for BMConfigParser._reset()

kdcis:
- Added testcase for _reset() function in ```src/tests/test_config.py```
- Enabled and ported code for python 3


src/tests/test_config.py
```diff
@@ -34,3 +34,8 @@ def test_safeGetInt(self):
             BMConfigParser().safeGetInt('nonexistent', 'nonexistent'), 0)
         self.assertEqual(
             BMConfigParser().safeGetInt('nonexistent', 'nonexistent', 42), 42)
+
+    def test_reset(self):
+        """Reset current config"""
+        bm = BMConfigParser()
+        self.assertEqual(bm._reset(), None)
```
g1itch, line 41: I expected something like

```python
self.assertEqual(
    BMConfigParser().safeGetInt('bitmessagesettings', 'settingsversion'), 0)

```

and `self.config = BMConfigParser()` in .`setUp()` if you separate BMConfigParser logic from singleton.

src/bmconfigparser.py
```diff
@@ -150,68 +157,40 @@ def items(self, section, raw=False, variables=None):
...
-                            SafeConfigParser.set(
-                                self, section, option, newVal)
-                    except configparser.InterpolationError:
-                        continue
+        return ConfigParser.items(self, section, True, variables)
+
+    @staticmethod
+    def addresses(hidden=False):
+        """Return a list of local bitmessage addresses (from section labels)"""
+        return [x for x in BMConfigParser().sections() if x.startswith('BM-') and (
```
PeterSurda, line 165: maybe this can work with `self.sections()` instead of `BMConfigParser().sections()`? Then you can keep it as a method.

src/tests/test_config.py
```diff
@@ -34,3 +34,9 @@ def test_safeGetInt(self):
             BMConfigParser().safeGetInt('nonexistent', 'nonexistent'), 0)
         self.assertEqual(
             BMConfigParser().safeGetInt('nonexistent', 'nonexistent', 42), 42)
+
+    def test_reset(self):
+        """safeGetInt retuns provided default for bitmessagesettings option or 0"""
+        self.config = BMConfigParser()
```
PeterSurda, line 40: maybe just
`config = BMConfigParser()`

PeterSurda (changes requested):

src/tests/test_config.py
g1itch: And then
```python
config.safeGetInt('bitmessagesettings', 'settingsversion')
```
in the following line.

I suggested `self.config = BMConfigParser()` in `.setUp()` if you want to reuse it in the rest of methods. Here it is redundant.

src/tests/test_config.py
```diff
@@ -34,3 +34,9 @@ def test_safeGetInt(self):
             BMConfigParser().safeGetInt('nonexistent', 'nonexistent'), 0)
         self.assertEqual(
             BMConfigParser().safeGetInt('nonexistent', 'nonexistent', 42), 42)
+
+    def test_reset(self):
```
PeterSurda, line 38: I don't think this is a correct test. The correct test should
1. load a test config file
2. either call `_reset` or load an empty file
3. the assert probably can remain as it is, but I would maybe put another assert in front of it, that tests if the `settingsversion` section is absent

PeterSurda (changes requested):
Please update.

src/tests/test_config.ini
```diff
@@ -0,0 +1,27 @@
...
+
+[inventory]
+storage= sqlite
+acceptmismatch= False
+
+[knownnodes]
+maxnodes= 15000
+
+[zlib]
+maxsize= 1048576
```
g1itch, line 27: Wrong place. There is a `test_pattern` dir for such files

src/bmconfigparser.py
```diff
@@ -6,7 +6,6 @@
 import shutil
 import sys  # FIXME: bad style! write more generally
 from datetime import datetime
-
```
g1itch, line 9: [PEP8 imports](https://www.python.org/dev/peps/pep-0008/#imports) requires blank line

src/bmconfigparser.py
```diff
@@ -192,6 +200,8 @@ def read(self, filenames):
                     except configparser.InterpolationError:
                         continue
 
+        def readfp(self, file_obj):
```
PeterSurda, line 203: ```def readfp(self, fp, filename=None):```

src/tests/test_config.py
```diff
@@ -1,13 +1,58 @@
...
+    configfile = StringIO('')
+
+    def setUp(self):
+        """creates a backup of BMConfigparser current state"""
+        BMConfigParser().write(self.configfile)
+        self.configfile.seek(0)
+
+    def tearDown(self):
+        """restore to the backup of BMConfigparser"""
+        BMConfigParser()._reset()
```
PeterSurda, line 54: override complaint

PeterSurda: ```#  pylint: protected-access=ignore```

src/tests/test_config.py
```diff
@@ -23,7 +68,6 @@ def test_safeGetBoolean(self):
             False
         )
         # no arg for default
-        # pylint: disable=too-many-function-args
```
PeterSurda, line 26: leave

src/tests/test_config.py
```diff
@@ -41,3 +85,15 @@ def test_safeGetFloat(self):
...
+
+    def test_reset(self):
+        """safeGetInt retuns provided default for bitmessagesettings option or 0"""
+        test_config_object = StringIO(test_config)
+        BMConfigParser().readfp(test_config_object)
+
+        self.assertEqual(
+            BMConfigParser().safeGetInt('bitmessagesettings', 'maxaddrperstreamsend'), 100)
+
+        BMConfigParser()._reset()
```
PeterSurda, line 97: again override linter

src/bmconfigparser.py
```diff
@@ -192,6 +200,8 @@ def read(self, filenames):
...
+            SafeConfigParser.read_file(self, file_obj)
```
PeterSurda, line 204: suppress linter

PeterSurda (changes requested):

src/tests/test_config.py
kdcis: updated

PeterSurda (approved):

PR #1767, opened 2021-05-27
Ported test_randomtrackingdict to Python3

kdcis:
Few changes to support python 3

### DIFF
--- a/src/randomtrackingdict.py
+++ b/src/randomtrackingdict.py
@@ -4,7 +4,7 @@ Track randomize ordered dict
 from threading import RLock
 from time import time
 
-import helper_random
+from pybitmessage import helper_random
 
 
 class RandomTrackingDict(object):
--- a/src/randomtrackingdict.py
+++ b/src/randomtrackingdict.py
@@ -4,7 +4,10 @@ Track randomize ordered dict
 from threading import RLock
 from time import time
 
-from pybitmessage import helper_random
+try:
+    import helper_random
+except ImportError:
+    from . import helper_random
 
 
 class RandomTrackingDict(object):
--- a/src/tests/test_randomtrackingdict.py
+++ b/src/tests/test_randomtrackingdict.py
@@ -1,5 +1,5 @@
 """
-Tests for RandomTrackingDict class
+Tests for RandomTrackingDict Class
 """
 import random
 import unittest
--- a/src/tests/test_randomtrackingdict.py
+++ b/src/tests/test_randomtrackingdict.py
@@ -5,9 +5,6 @@ import random
 import unittest
 
 from time import time
-from .common import skip_python3
-
-skip_python3()
 
 
 class TestRandomTrackingDict(unittest.TestCase):


g1itch (approved):
Yes, it's trivial.

g1itch:
Would be great to also have a test in the `core.TestCore`. Because it looks like it breaks portable mode.

kdcis:
> Would be great to also have a test in the `core.TestCore`. Because it looks like it breaks portable mode.

Well, I'm not sure if `core.TestCore` is ported to python 3. So I'll have a look and try doing that.

g1itch:
> > Would be great to also have a test in the `core.TestCore`. Because it looks like it breaks portable mode.
> 
> Well, I'm not sure if `core.TestCore` is ported to python 3. So I'll have a look and try doing that.

While you are porting some stuff to python3 it's important not to break current state (python2 with related imports) which is working and used by other people. If I merge this (I was going to, but decided to check twice), tomorrow we'll read the complaints in `[chan] bitmessage` for sure.

Currently test mode is the sole (but still not good enough) way to automatically check the current state - complete running application. There will be also a separate test for windows bundle using test-mode: 5cff5a7.

Absolute imports doesn't work because of circular imports in `network` and `messagetypes`. You can observe it in my [branch setup](https://github.com/g1itch/PyBitmessage/commits/setup).

g1itch:
That's what I mean: https://travis-ci.org/github/g1itch/PyBitmessage/jobs/773042961

kdcis:
> That's what I mean: https://travis-ci.org/github/g1itch/PyBitmessage/jobs/773042961
> 
> So, maybe
> 
> ```python
> try:
>     import helper_random
> except ImportError:
>     from . import helper_random
> ```

Okay, Got it. Thanks for the explanation.

g1itch (changes requested):
Not so easy. It works only when installed.

g1itch:
OK, now it pass: https://travis-ci.org/github/g1itch/PyBitmessage/jobs/773056976

g1itch (approved):
OK

PeterSurda:
We need to add tests that run in an environment without setuptools, or we need to decide to drop support for this mode.

PeterSurda (approved):

g1itch:
> We need to add tests that run in an environment without setuptools, or we need to decide to drop support for this mode.

Nope, you can't decide to drop portable mode. Not only because it's popularity among users, but also because windows bundle and, I hope, upcoming AppImage rely on it. Unfortunately I could not find the sources for buildbot tasks on git.bitmessage.org, so  I crafted another Travis job based on unfinished deploy stage: [ci-bundles](https://github.com/g1itch/PyBitmessage/tree/ci-bundles). I also started to write some tests for plugin in my `tests` branch.

g1itch:
Oops, git says "remote: error: Commits must have valid signatures." I cannot merge it.

PeterSurda:
> Nope, you can't decide to drop portable mode. Not only because it's popularity among users, but also because windows bundle and, I hope, upcoming AppImage rely on it.

I understand there is work associated with it, my point was just that we broadly have two choices. The current status, keeping it untested, isn't really a choice.

> Unfortunately I could not find the sources for buildbot tasks on git.bitmessage.org, so I crafted another Travis job based on unfinished deploy stage: [ci-bundles](https://github.com/g1itch/PyBitmessage/tree/ci-bundles). I also started to write some tests for plugin in my `tests` branch.

1. Create an account on git.bitmessage.org
2. Give me the login, and I'll give you access to the repo. It's a private repo, although it doesn't contain any passwords or keys (these are handled by a secret provider) but for the time being I'm not making it public
3. There are actually several repos, one is for buildbot config file, and another repos are for scripts, dockerfiles, and so on.
4. Kindly note it isn't deployed automatically yet, I have to do it manually, but at least it's all in version control

kdcis:
> Oops, git says "remote: error: Commits must have valid signatures." I cannot merge it.

Sorry for that. Please try now, I've signed the commits.

Issue #1450, opened 2019-03-30
Official AppImage


tradewatcher:
Hi,

can you provide an AppImage for easy use, portability and cross-platform support?

I was able to compile it with guide but its horrible to get it running without problems in portable mode.
Take a look at AppImage its really awesome.

PeterSurda:
Not sure why you need portable mode on Linux. While it probably works, it isn't being tested. Using the BITMESSAGE_HOME variable instead is recommended. Ideally, you'd install bitmessage with a package manager that comes with your disto, but we're not there yet.

tradewatcher:
I'm a beginner on Linux and it took me ~1 hour to compile, run, search files and make it portable. 

*) Easy for beginner
*) Automatic Portable feature from AppImage
*) 1 File - faster transfer
*) Cross platform support
*) Independence from package manager
*) No extra installation and sharing of dependencies (Greetings from Spectre..)
*) Assisting "AppImage" improve the Linux ecosystem and make it easier for newcomer
*) Good chances for better support of AppImage on Win10 if the ecosystem is growing.

Why portable on Linux? Encrypted usb drive with software and clean Kali/Tails distro improve security and usability.


Helpful Links to create an AppImage:
https://github.com/AppImage/AppImageKit
https://github.com/AppImage/AppImageKit/releases (AppImage of Kit)
https://discourse.appimage.org/t/how-to-create-an-appimage/155 (Help Board)

tradewatcher:
AppImage is portable if you create a folder "FILENAME.AppImage.home" and/or "FILENAME.AppImage.config"

It takes much less effort to create an AppImage compared to ex. snap packages.

mr-tron:
>  A single thing you still need to compile is C PoW.
and install python2 and qt4, which can be a problem in modern OS.

g1itch:
> > A single thing you still need to compile is C PoW.
> 
> and install python2 and qt4, which can be a problem in modern OS.

When I wrote it, python2 was still available in most distros.

There is a [branch for Qt5](https://github.com/g1itch/PyBitmessage/tree/qt5-wip).

PR #1769, opened 2021-05-27
Added AppImage build script for PyBitmessage

kdcis:


g1itch:
Could you build the appimage on travis?

g1itch:
Please don't include binary. It's build should be part of release process.

packages/AppImage/install.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+wget -O "appimagetool" "https://github.com/AppImage/AppImageKit/releases/download/continuous/appimagetool-x86_64.AppImage"
+chmod a+x appimagetool
+mv appimagetool /usr/bin/
```
g1itch, line 5: this requires root

kdcis: Yes, is that a problem?. If yes I can change so it won't require root.

g1itch: I'm not sure. Let's check: https://github.com/g1itch/PyBitmessage/commits/ci-wine-further

g1itch:
I see some issue with script: https://travis-ci.org/github/g1itch/PyBitmessage/jobs/773056006, need to check it manually as well.

Also mark the last commit as closing #1450 when finish.

g1itch:
OK, I found possible fix for AppRun: efefa4f. Bad news: this appimage still requires PyQt4 installed system wide.

packages/AppImage/install.sh
g1itch: It should work without root: a67ad09

kdcis: OKay, I'll make the changes.

g1itch:
Please squash two first commits to not bloat the repo. Also we need dependencies inside of the bundle, because its purpose is to work around the removal of PyQt4 and python2.

PR #1773, opened 2021-06-02
Code quality

PeterSurda:
A couple of code quality fixes.


src/network/udp.py
```diff
@@ -21,7 +21,7 @@ class UDPSocket(BMProto):  # pylint: disable=too-many-instance-attributes
     port = 8444
 
     def __init__(self, host=None, sock=None, announcing=False):
-        # pylint: disable=bad-super-call
+        # pylint: disable=bad-super-call, access-member-before-definition
         super(BMProto, self).__init__(sock=sock)
```
g1itch, line 25:
```python
super(UDPSocket, self).__init__(sock=sock)
```
?

src/network/udp.py
```diff
@@ -97,22 +97,30 @@ def bm_command_addr(self):
             time.time()
         return True
 
+    # override the protocol do ignore the following commands on UDP
+    # we do this mainly for security reasons
```
g1itch, line 101: This is more of inheritance flow. Maybe `UDPSocket` should inherit from some more basic class?

src/network/udp.py
```diff
@@ -21,7 +21,7 @@ class UDPSocket(BMProto):  # pylint: disable=too-many-instance-attributes
     port = 8444
 
     def __init__(self, host=None, sock=None, announcing=False):
-        # pylint: disable=bad-super-call
+        # pylint: disable=bad-super-call, access-member-before-definition
```
g1itch, line 24: This is a bug: https://github.com/PyCQA/pylint/issues/2315

src/network/udp.py
PeterSurda: Perhaps.

src/network/udp.py
Jeroenvb3: The linked bug has been fixed. The ignore shouldn't be needed anymore.

PR #1776, opened 2021-06-07
Dry run for windows exe in wine

g1itch:
Hello!

I added dry-run step to buildscripts/winbuild.sh as you suggested.

### DIFF
--- a/buildscripts/winbuild.sh
+++ b/buildscripts/winbuild.sh
@@ -167,6 +167,14 @@ function build_exe(){
 	wine pyinstaller bitmessagemain.spec
 }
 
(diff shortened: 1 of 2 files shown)


g1itch:
Interesting, "FATAL ERROR: cannot create temporary directory!" with 64bit exe.

g1itch:
https://bugs.winehq.org/show_bug.cgi?id=49334

g1itch:
It works if I downgrade pyinstaller to 3.5, but that issue is caused by security fix in 3.6 ):

PeterSurda (approved):

PR #1779, opened 2021-06-11
Bundle some plugins into windows exe

g1itch:
Hello!

I found a way to bundle plugins for Windows. With these changes `menu_qrcode` works already, `proxyconfig_stem` needs more work (no Unix sockets).

### DIFF
--- a/buildscripts/winbuild.sh
+++ b/buildscripts/winbuild.sh
@@ -108,11 +108,12 @@ function install_pyinstaller()
 	fi
 }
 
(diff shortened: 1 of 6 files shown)


g1itch:
With these changes you can copy the exe into the dir, where you have extracted Windows Expert Bundle or other tor distribution and switch proxy type to 'stem'.

PeterSurda (approved):
Haven't tested but looks ok.

PR #1780, opened 2021-06-14
Fix random failures of test_onionservicesonly()

g1itch:
Hello!

I did my homework with `TestCore.test_onionservicesonly()`. With these changes it should not fail randomly anymore.

Failure example: https://travis-ci.org/github/Bitmessage/PyBitmessage/jobs/774315030


PeterSurda (approved):
Looks ok

PR #1785, opened 2021-07-06
Added bitmessagekivy sample test case

navjotcis:
I have added test_setting screen test 
I have added a mock test for running bitmessagekivy test cases.
I have added docker `run-kivy-tests-in-docker.sh` script for running bitmessage kivy test cases in Travis.

### DIFF
--- a/kivy-requirements.txt
+++ b/kivy-requirements.txt
@@ -1,0 +1,5 @@
+kivy-garden.qrcode
+-e git+https://github.com/kivymd/KivyMD#egg=kivymd
+opencv-python
+pyzbar
+telenium
--- a/src/bitmessagekivy/mpybit.py
+++ b/src/bitmessagekivy/mpybit.py
@@ -1,0 +1,28 @@
+"""
+    Dummy implementation for kivy Desktop and android(mobile) interface
+"""
+# pylint: disable=too-few-public-methods
+
+from kivy.app import App
+from kivy.uix.label import Label
+
+
+class NavigateApp(App):
+    """Navigation Layout of class"""
+
+    def build(self):
+        """Method builds the widget"""
+        # pylint: disable=no-self-use
+        return Label(text="Hello World !")
+
+    def clickNavDrawer(self):
+        """method for clicking navigation drawer"""
+        pass
+
+    def addingtoaddressbook(self):
+        """method for clicking address book popup"""
+        pass
+
+
+if __name__ == '__main__':
+    NavigateApp().run()
--- a/src/bitmessagekivy/tests/telenium_process.py
+++ b/src/bitmessagekivy/tests/telenium_process.py
@@ -1,0 +1,83 @@
+"""
+    Base class for telenium test cases which run kivy app as background process
+"""
+
+import os
+import shutil
+import tempfile
+from time import time, sleep
+
+from telenium.tests import TeleniumTestCase
+
+_files = (
+    'keys.dat', 'debug.log', 'messages.dat', 'knownnodes.dat',
+    '.api_started', 'unittest.lock'
+)
+
+tmp_db_file = (
+    'keys.dat', 'messages.dat'
+)
+
+
+def cleanup(files=_files):
+    """Cleanup application files"""
+    for pfile in files:
+        try:
+            os.remove(os.path.join(tempfile.gettempdir(), pfile))
+        except OSError:
+            pass
+
+
+def populate_test_data():
+    """Set temp data in tmp directory"""
+    for file_name in tmp_db_file:
+        old_source_file = os.path.join(
+            os.path.abspath(os.path.dirname(__file__)), 'sampleData', file_name)
+        new_destination_file = os.path.join(os.environ['BITMESSAGE_HOME'], file_name)
+        shutil.copyfile(old_source_file, new_destination_file)
+
+
+class TeleniumTestProcess(TeleniumTestCase):
+    """Setting Screen Functionality Testing"""
+    cmd_entrypoint = [os.path.join(os.path.abspath(os.getcwd()), 'src', 'mock', 'kivy_main.py')]
+
+    @classmethod
+    def setUpClass(cls):
+        """Setupclass is for setting temp environment"""
+        os.environ["BITMESSAGE_HOME"] = tempfile.gettempdir()
+        populate_test_data()
+        super(TeleniumTestProcess, cls).setUpClass()
+
+    @classmethod
+    def tearDownClass(cls):
+        """Ensures that pybitmessage stopped and removes files"""
+        # pylint: disable=no-member
+        try:
+            cls.cli.app_quit()
+        except:
+            pass
+
+        try:
+            cls.process.kill()
+        except:
+            pass
+        cleanup()
+
+    def drag(self, xpath1, xpath2):
+        """this method is for dragging"""
+        self.cli.drag(xpath1, xpath2, 1)
+        self.cli.sleep(0.3)
+
+    def assertCheckScrollDown(self, selector, timeout=-1):
+        """this method is for checking scroll"""
+        start = time()
+        while True:
+            scroll_distance = self.cli.getattr(selector, 'scroll_y')
+            if scroll_distance > 0.0:
+                self.assertGreaterEqual(scroll_distance, 0.0)
+                return True
+            if timeout == -1:
+                return False
+            if timeout > 0 and time() - start > timeout:
+                raise Exception("Timeout")
+            sleep(0.1)
--- a/src/bitmessagekivy/tests/test_setting_screen.py
+++ b/src/bitmessagekivy/tests/test_setting_screen.py
@@ -1,0 +1,28 @@
+# pylint: disable=too-few-public-methods
+
+from .telenium_process import TeleniumTestProcess
+from .common import skip_screen_checks
+
+
+class SettingScreen(TeleniumTestProcess):
+    """Setting Screen Functionality Testing"""
+
+    @skip_screen_checks
+    def test_setting_screen(self):
+        """Show Setting Screen"""
+        print("=====================Test -Show Setting Screen=====================")
+        self.cli.sleep(3)
+        # this is for checking current screen
+        self.assertExists("//Inbox[@name~=\"inbox\"]", timeout=2)
+        # this is for opening Nav drawer
+        self.cli.wait_click('//MDActionTopAppBarButton[@icon=\"menu\"]', timeout=2)
+        # checking state of Nav drawer
+        self.assertExists("//MDNavigationDrawer[@state~=\"open\"]", timeout=2)
+        # this is for scrolling Nav drawer
+        self.drag("//NavigationItem[@text=\"Sent\"]", "//NavigationItem[@text=\"Inbox\"]")
+        # assert for checking scroll funcation
+        self.assertCheckScrollDown('//ContentNavigationDrawer//ScrollView[0]', timeout=3)
+        # this is for opening setting screen
+        self.cli.wait_click('//NavigationItem[@text=\"Settings\"]', timeout=1)
+        # Checking current screen
+        self.assertExists("//Setting[@name~=\"set\"]", timeout=2)
--- a/src/mock/class_addressGenerator.py
+++ b/src/mock/class_addressGenerator.py
@@ -1,0 +1,96 @@
+"""
+A thread for creating addresses
+"""
+
+import logging
+import random
+import threading
+
+from pybitmessage import state
+from pybitmessage import queues
+
+from pybitmessage.bmconfigparser import BMConfigParser
+
+# from network.threads import StoppableThread
+
+
+fake_addresses = {
+    'BM-2cUgQGcTLWAkC6dNsv2Bc8XB3Y1GEesVLV': {
+        'privsigningkey': '5KWXwYq1oJMzghUSJaJoWPn8VdeBbhDN8zFot1cBd6ezKKReqBd',
+        'privencryptionkey': '5JaeFJs8iPcQT3N8676r3gHKvJ5mTWXy1VLhGCEDqRs4vpvpxV8'
+    },
+    'BM-2cUd2dm8MVMokruMTcGhhteTpyRZCAMhnA': {
+        'privsigningkey': '5JnJ79nkcwjo4Aj7iG8sFMkzYoQqWfpUjTcitTuFJZ1YKHZz98J',
+        'privencryptionkey': '5JXgNzTRouFLqSRFJvuHMDHCYPBvTeMPBiHt4Jeb6smNjhUNTYq'
+    },
+    'BM-2cWyvL54WytfALrJHZqbsDHca5QkrtByAW': {
+        'privsigningkey': '5KVE4gLmcfYVicLdgyD4GmnbBTFSnY7Yj2UCuytQqgBBsfwDhpi',
+        'privencryptionkey': '5JTw48CGm5CP8fyJUJQMq8HQANQMHDHp2ETUe1dgm6EFpT1egD7'
+    },
+    'BM-2cTE65PK9Y4AQEkCZbazV86pcQACocnRXd': {
+        'privsigningkey': '5KCuyReHx9MB4m5hhEyCWcLEXqc8rxhD1T2VWk8CicPFc8B6LaZ',
+        'privencryptionkey': '5KBRpwXdX3n2tP7f583SbFgfzgs6Jemx7qfYqhdH7B1Vhe2jqY6'
+    },
+    'BM-2cX5z1EgmJ87f2oKAwXdv4VQtEVwr2V3BG': {
+        'privsigningkey': '5K5UK7qED7F1uWCVsehudQrszLyMZxFVnP6vN2VDQAjtn5qnyRK',
+        'privencryptionkey': '5J5coocoJBX6hy5DFTWKtyEgPmADpSwfQTazMpU7QPeART6oMAu'
+    }
+}
+
+
+class StoppableThread(threading.Thread):
+    """Base class for application threads with stopThread method"""
+    name = None
+    logger = logging.getLogger('default')
+
+    def __init__(self, name=None):
+        if name:
+            self.name = name
+        super(StoppableThread, self).__init__(name=self.name)
+        self.stop = threading.Event()
+        self._stopped = False
+        random.seed()
+        self.logger.info('Init thread %s', self.name)
+
+    def stopThread(self):
+        """Stop the thread"""
+        self._stopped = True
+        self.stop.set()
+
+
+class FakeAddressGenerator(StoppableThread):
+    """A thread for creating fake addresses"""
+    name = "addressGenerator"
+    address_list = list(fake_addresses.keys())
+
+    def stopThread(self):
+        try:
+            queues.addressGeneratorQueue.put(("stopThread", "data"))
+        except:
+            pass
+        super(FakeAddressGenerator, self).stopThread()
+
+    def run(self):
+        """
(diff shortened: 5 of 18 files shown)


src/state.py
```diff
@@ -68,3 +68,5 @@
 
 thisapp = None
 """Singleton instance"""
+
+isKivyworking = False
```
PeterSurda, line 72: rename

src/mock/class_addressGenerator.py
```diff
@@ -0,0 +1,112 @@
...
+import random
+from pybitmessage import state
+import threading
+
+from pybitmessage.bmconfigparser import BMConfigParser
+
+# from network.threads import StoppableThread
+
+
+fake_addresses = {
```
PeterSurda, line 20: this looks ok now

src/mock/class_addressGenerator.py
```diff
@@ -0,0 +1,112 @@
...
+    def run(self):
+        """
+        Process the requests for addresses generation
+        from `.queues.addressGeneratorQueue`
+        """
+        while state.shutdown == 0:
+            queueValue = addressGeneratorQueue.get()
+            streamNumber = 1
+            try:
+                if len(BMConfigParser().addresses()) > 0:
```
PeterSurda, line 88: I don't like they way this looks.

src/bitmessagekivy/tests/telenium_process.py
```diff
@@ -0,0 +1,71 @@
...
+
+        try:
+            cls.process.kill()
+        except:
+            pass
+
+    def click_on(self, xpath, seconds=0.3):
+        """this methos is used for on_click event with time"""
+        self.cli.click_on(xpath)
+        self.cli.sleep(seconds)
```
PeterSurda, line 66: no

src/bitmessagekivy/tests/telenium_process.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def drag(self, xpath1, xpath2):
+        """this method is for dragging screen"""
+        self.cli.drag(xpath1, xpath2, 1)
+        self.cli.sleep(0.3)
```
PeterSurda, line 71: no

src/bitmessagekivy/tests/test_setting_screen.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class SettingScreen(TeleniumTestProcess):
+    """Setting Screen Functionality Testing"""
+
+    @skip_screen_checks
+    def test_setting_screen(self):
+        """Show Setting Screen"""
+        print("=====================Test -Show Setting Screen=====================")
+        self.cli.sleep(3)
```
PeterSurda, line 14: no, wait for something useful

src/bitmessagekivy/tests/test_setting_screen.py
```diff
@@ -0,0 +1,27 @@
...
+        # this is for opening Nav drawer
+        self.click_on('//MDActionTopAppBarButton[@icon=\"menu\"]')
+        # checking state of Nav drawer
+        self.assertEqual(self.cli.getattr('//MDNavigationDrawer', 'state'), 'open')
```
PeterSurda, line 18: Find a better way to do this.

src/bitmessagekivy/tests/test_setting_screen.py
```diff
@@ -0,0 +1,27 @@
...
+        # this is for scrolling Nav drawer
+        self.drag("//NavigationItem[@text=\"Sent\"]", "//NavigationItem[@text=\"Inbox\"]")
+        # assert for checking scroll funcation
+        scroll_distance = self.cli.getattr('//ContentNavigationDrawer//ScrollView[0]', 'scroll_y')
+        self.assertLessEqual(scroll_distance, -0.0)
```
PeterSurda, line 23: find a better way to do this

PeterSurda (changes requested):

src/mock/class_addressGenerator.py
navjotcis: ok

src/mock/class_addressGenerator.py
navjotcis: I have changed it

src/state.py
navjotcis: now i am renaming it with kivyAppWorking

src/bitmessagekivy/tests/test_setting_screen.py
navjotcis: Need to wait here for app to load

src/bitmessagekivy/tests/telenium_process.py
navjotcis: here it's necessary to  hold for 0.3 seconds

src/bitmessagekivy/tests/telenium_process.py
```diff
@@ -0,0 +1,80 @@
...
+        """Setupclass is for setting temp environment"""
+        os.environ["BITMESSAGE_HOME"] = tempfile.gettempdir()
+        set_temp_data()
+        super(TeleniumTestProcess, cls).setUpClass()
+
+    @classmethod
+    def tearDownClass(cls):
+        """Ensures that pybitmessage stopped and removes files"""
+        # pylint: disable=no-member
+        cleanup()
```
PeterSurda, line 53: `cleanup()` should go **after** the `quit`/`kill`.

src/bitmessagekivy/tests/telenium_process.py
PeterSurda: I'll accept this for the time being as I see it's difficult to avoid. Try to think about an assert that can check for this but if you can't then it's ok.

src/bitmessagekivy/tests/telenium_process.py
```diff
@@ -0,0 +1,80 @@
...
+    def drag(self, xpath1, xpath2):
+        """this method is for dragging"""
+        self.cli.drag(xpath1, xpath2, 1.5)
+        self.cli.sleep(0.3)
+
+    def assertCheckScrollDown(self, scroll_distance, destination, timeout=-1):
+        """this method is for checking scroll"""
+        start = time()
+        while True:
+            if scroll_distance <= destination:
```
PeterSurda, line 73: It looks like here `scroll_distance` is a value, rather than a reference, which make the test less useful.

src/bitmessagekivy/tests/test_setting_screen.py
PeterSurda: Ok I'll accept it for now.

src/bitmessagekivy/tests/test_setting_screen.py
```diff
@@ -0,0 +1,29 @@
...
+        self.assertExists("//Inbox[@name~=\"inbox\"]", timeout=2)
+        # this is for opening Nav drawer
+        self.cli.wait_click('//MDActionTopAppBarButton[@icon=\"menu\"]', timeout=2)
+        # checking state of Nav drawer
+        self.assertExists("//MDNavigationDrawer[@state~=\"open\"]", timeout=2)
+        # this is for scrolling Nav drawer
+        self.drag("//NavigationItem[@text=\"Sent\"]", "//NavigationItem[@text=\"Inbox\"]")
+        # assert for checking scroll funcation
+        scroll_distance = self.cli.getattr('//ContentNavigationDrawer//ScrollView[0]', 'scroll_y')
+        self.assertCheckScrollDown(scroll_distance, -0.0, timeout=1)
```
PeterSurda, line 25: Here, `scroll_distance` should be a reference as otherwise the test is not very helpful.

src/mock/class_addressGenerator.py
```diff
@@ -0,0 +1,109 @@
...
+        'privencryptionkey': '5J5coocoJBX6hy5DFTWKtyEgPmADpSwfQTazMpU7QPeART6oMAu'
+    }
+}
+
+
+UISignalQueue = Queue.Queue()
+addressGeneratorQueue = Queue.Queue()
+
+
+class StoppableThread(threading.Thread):
```
PeterSurda, line 48: Can't we import `StoppableThread` from the original location?

src/mock/class_addressGenerator.py
```diff
@@ -0,0 +1,109 @@
...
+    },
+    'BM-2cX5z1EgmJ87f2oKAwXdv4VQtEVwr2V3BG': {
+        'privsigningkey': '5K5UK7qED7F1uWCVsehudQrszLyMZxFVnP6vN2VDQAjtn5qnyRK',
+        'privencryptionkey': '5J5coocoJBX6hy5DFTWKtyEgPmADpSwfQTazMpU7QPeART6oMAu'
+    }
+}
+
+
+UISignalQueue = Queue.Queue()
+addressGeneratorQueue = Queue.Queue()
```
PeterSurda, line 45: This may not work correctly down the line as the `UISignalQueue` and `addressGeneratorQueue` are global variables. Instead, `state.py` and `queues.py` should be made compatible with python3 and imported here.

src/mock/class_addressGenerator.py
```diff
@@ -0,0 +1,109 @@
...
+    address_list = list(fake_addresses.keys())
+
+    def stopThread(self):
+        try:
+            addressGeneratorQueue.put(("stopThread", "data"))
+        except:
+            pass
+        super(FakeAddressGenerator, self).stopThread()
+
+    def run(self):
```
PeterSurda, line 80: This section is ok.

src/mock/class_addressGenerator.py
```diff
@@ -0,0 +1,109 @@
...
+                BMConfigParser().add_section(address)
+                BMConfigParser().set(address, 'label', label)
+                BMConfigParser().set(address, 'enabled', 'true')
+                BMConfigParser().set(
+                    address, 'privsigningkey', fake_addresses[address]['privsigningkey'])
+                BMConfigParser().set(
+                    address, 'privencryptionkey', fake_addresses[address]['privencryptionkey'])
+                BMConfigParser().save()
+
+                UISignalQueue.put((
```
PeterSurda, line 101: We can probably skip using `UISignalQueue` in this mock class for the time being. Maybe it will be needed later but at the moment it isn't.

src/mock/main.py
```diff
@@ -0,0 +1,29 @@
+"""This module is for thread start."""
```
PeterSurda, line 1: This file should have a different name. For example, `kivy_main.py`. Also the docstring should say something like `Mock kivy app with mock threads`.

src/mock/main.py
```diff
@@ -0,0 +1,29 @@
...
+def main():
+    """main method for starting threads"""
+    if state.enableObjProc:
+        # Start the address generation thread
+        addressGeneratorThread = FakeAddressGenerator()
+        # close the main program even if there are threads left
+        addressGeneratorThread.daemon = True
+        addressGeneratorThread.start()
+
+        # sqlLookup = sqlThread()
```
PeterSurda, line 18: this can be removed, once we have a mock sql thread we can re-add.

src/state.py
```diff
@@ -68,3 +68,5 @@
 
 thisapp = None
 """Singleton instance"""
+
+kivyAppWorking = False
```
PeterSurda, line 72: `enableKivy` and put it into the section with other `enable....`

PeterSurda (changes requested):
Please make the requested changes, it's close to being able to merge, the major issues have been resolved.

src/bitmessagekivy/mpybit.py
```diff
@@ -0,0 +1,28 @@
+"""
+    Bitmessage Desktop and android(mobile) interface
```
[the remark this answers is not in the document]
navjotcis: ok

src/bitmessagekivy/tests/telenium_process.py
navjotcis: Yes I know scroll_distance is the value of the y-axis which starts from 0.0-1.0 so I am checking if scroll values are less than or equal to 0.0 then it means that someone has scrolled the nav drawer if the condition is satisfied then the test passes otherwise it given and fail result

src/mock/class_addressGenerator.py
navjotcis: we can import it but we need to deal with lots of import issues and currently python3 porting work is in progress so if I will make changes then it can cause issues so I am making StoppableThread in a separate file for a temporary basis

src/mock/class_addressGenerator.py
navjotcis: ok

src/mock/class_addressGenerator.py
navjotcis: Ok, For now, I am removing it but it will be required in the test_create_random_address.py test case because here I am checking  created address in spinner or my address list so this will be required in that case

src/mock/main.py
navjotcis: ok

src/bitmessagekivy/tests/telenium_process.py
PeterSurda: I think what you said is correct, the problem is that you retrieve the value only once, before calling this method. In other words, the loop doesn't check if the value has changed. The loop is superfluous as it is repeating exactly the same thing. You should pass a reference, e.g. a function, or a telenium selector, not a value.

src/bitmessagekivy/tests/test_setting_screen.py
PeterSurda: Here, you retrieve `scroll_distance`, which I presume is a float, and then pass it to the assert. Instead, it should pass some sort of reference.

src/bitmessagekivy/tests/telenium_process.py
PeterSurda: In other words, `scroll_distance` is a float and the value is assigned before this method is called. It doesn't change while the loop is running.

src/mock/class_addressGenerator.py
PeterSurda: I was worried about that one. How much work is it to make the original `StoppableThread` compatible with python3, can someone else help you maybe?

src/mock/class_addressGenerator.py
PeterSurda: Ok.

kivy-requirements.txt
```diff
@@ -0,0 +1,34 @@
+certifi==2020.12.5
```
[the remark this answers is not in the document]
navjotcis: ok

src/bitmessagekivy/tests/telenium_process.py
navjotcis: ok

src/mock/class_addressGenerator.py
navjotcis: We need to change the import format of the network, storage, and pyelliptic modules, and maybe there is more but when I was working on it I kept getting issues in every files one by one.

src/bitmessagekivy/tests/telenium_process.py
```diff
@@ -0,0 +1,85 @@
...
+
+tmp_db_file = (
+    'keys.dat', 'messages.dat'
+)
+
+
+def cleanup(home=None, files=_files):
+    """Cleanup application files"""
+    if not home:
+        home = tempfile.gettempdir()
```
PeterSurda, line 25: this is useless.

src/bitmessagekivy/tests/telenium_process.py
```diff
@@ -0,0 +1,85 @@
...
+        try:
+            cls.cli.app_quit()
+        except:
+            pass
+
+        try:
+            cls.process.kill()
+        except:
+            pass
+        cleanup()
```
PeterSurda, line 66: doesn't pass directory

src/bitmessagekivy/tests/telenium_process.py
```diff
@@ -0,0 +1,85 @@
...
+    if not home:
+        home = tempfile.gettempdir()
+    for pfile in files:
+        try:
+            os.remove(os.path.join(home, pfile))
+        except OSError:
+            pass
+
+
+def set_temp_data():
```
PeterSurda, line 33: I would call this differetly, e.g. `populate_test_data`

src/bitmessagekivy/tests/telenium_process.py
```diff
@@ -0,0 +1,85 @@
...
+    """Set temp data in tmp directory"""
+    for file_name in tmp_db_file:
+        old_source_file = os.path.join(
+            os.path.abspath(os.path.dirname(__file__)), 'sampleData', file_name)
+        new_destination_file = os.path.join(os.environ['BITMESSAGE_HOME'], file_name)
+        shutil.copyfile(old_source_file, new_destination_file)
+
+
+class TeleniumTestProcess(TeleniumTestCase):
```
PeterSurda, line 42: I would make the two function from above, `cleanup` and `set_temp_data`, methods of this function, to avoid passing around directory.

src/mock/kivy_main.py
```diff
@@ -0,0 +1,22 @@
+"""Mock kivy app with mock threads."""
+
+from pybitmessage import state
+from pybitmessage.bitmessagekivy.mpybit import NavigateApp
+from class_addressGenerator import FakeAddressGenerator
+
+
+def main():
+    """main method for starting threads"""
+    if state.enableObjProc:
```
PeterSurda, line 10: you don't need this test, just always create the thread.

PeterSurda (changes requested):
In addition to the individual comments, review the docstrings. The docstrings should provide such information that a developer who wants to work on the project can read the automatically generated documentation from the docstrings and understand it sufficiently to work on it.

PeterSurda (approved):

g1itch:
36 M of binary data, OMG

g1itch:
Also I cannot understand: if there is a `bitmessagekivy` package, to which the tests belong, why all these files (requirements, Dockerfile, Travis recipe, test script) are in the top level dir, not inside of the `src/bitmessagekivy/` where the `bitmessagekivy` setup script is supposed to be?

navjotcis:
we are keeping all the files(requirements, Dockerfile, Travis recipe, test script) outside the bitmessagekivy because bitmessageqt is also following the same way of managing docker script and test cases so I think we should follow the same structure.

g1itch:
> we are keeping all the files(requirements, Dockerfile, Travis recipe, test script) outside the bitmessagekivy because bitmessageqt is also following the same way of managing docker script and test cases so I think we should follow the same structure.

I don't think so. In my opinion `bitmessageqt` may have it's own test suite and  setup script, obviously with it's own dependencies. Dockerfile is nonsense for a GUI application, Travis recipe since 17 Jun is just a lazy appeal to buildbot and should probably be replaced by proper buildbot task in the master.cfg.

All that files just litter the top level dir. There are `packages` and `buildscripts` dirs for the scripts and alternative dockerfiles.

PeterSurda:
I don't have a particularly strong opinion one way or another. The new Dockerfile and travis file are for running the new tests (which require a different environment than the old tests) locally on your machine, so that you don't depend on an external CI.

I'm not particularly excited about including a big binary blob for the database, but I'm even less happy about complex test setups which are now required for some tests. This one may be big, but is less complex, and replacing it with a smaller one later is easier.

g1itch:
There is already -f in the [run-kivy-tests-in-docker.sh](../blob/v0.6/run-kivy-tests-in-docker.sh) so it might easily be `packages/Dockerfile.kivy-travis`. To not depend on external CI I use tox. It works way better than rebuilding a container each time. And I can even comment core tests and make a speed run.

PR #1788, opened 2021-07-21
Compatibility tests

g1itch:
Hello!

Let's introduce some simple compatibility tests, based on a sample data and well known values. I wish them to cover the implementations of all the functionality, defined in the protocol. So that an alternative implementation can take the samples to check it's compatibility.

It's also helpful for the ongoing python3 porting. Because most of the developers are not writing any tests. And I've already found [a feature](https://github.com/g1itch/PyBitmessage/runs/3116252583?check_suite_focus=true) in `highlevelcrypto` module, which I don't yet understand, but I think that such incompatibilities is better to reveal soon. The test shows that not all the subset of `pyelliptic` is used in `highlevelcrypto` in compatible way with python3.

### DIFF
--- a/src/protocol.py
+++ b/src/protocol.py
@@ -22,6 +22,7 @@ from bmconfigparser import BMConfigParser
 from debug import logger
 from fallback import RIPEMD160Hash
 from helper_sql import sqlExecute
+from network.node import Peer
 from version import softwareVersion
 
 # Service flags
@@ -54,6 +55,9 @@ OBJECT_BROADCAST = 3
 OBJECT_ONIONPEER = 0x746f72
 OBJECT_I2P = 0x493250
 OBJECT_ADDR = 0x61646472
+
+#: protocol specification says max 1000 addresses in one addr command
+MAX_ADDR_COUNT = 1000
 
 eightBytesOfRandomDataUsedToDetectConnectionsToSelf = pack(
     '>Q', random.randrange(1, 18446744073709551615))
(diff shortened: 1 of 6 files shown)


g1itch:
And of course the check's gonna fail.

g1itch:
> The test shows that not all the subset of `pyelliptic` is used in `highlevelcrypto` in compatible way with python3.

It's even worse: apparently the `pyelliptic` is not python3 compatible. It's just not covered with tests.

g1itch:
> And I've already found [a feature](https://github.com/g1itch/PyBitmessage/runs/3116252583?check_suite_focus=true) in `highlevelcrypto` module, which I don't yet understand, but I think that such incompatibilities is better to reveal soon.

solved it in the recent code: https://github.com/g1itch/PyBitmessage/runs/3182720003#step:8:70

src/protocol.py
```diff
@@ -22,8 +22,14 @@
 from debug import logger
 from fallback import RIPEMD160Hash
 from helper_sql import sqlExecute
+from network.node import Peer
+from randomtrackingdict import RandomTrackingDict
```
g1itch, line 26: Interesting. Unused import is simply ignored here.

PeterSurda:
Maybe this shouldn't be called "Compatiblity tests" as most of it is network subsystem refactoring. Could you perhaps put the (smaller) protocol refactoring into a separate PR, and just keep this for the network subsystem? Then the protocol refactoring can be merged earlier and faster.

PeterSurda:
maybe also "random" refactoring separately.

g1itch:
> Maybe this shouldn't be called "Compatiblity tests" as most of it is network subsystem refactoring. Could you perhaps put the (smaller) protocol refactoring into a separate PR, and just keep this for the network subsystem? Then the protocol refactoring can be merged earlier and faster.

I see, the rest of this branch is currently a complicated changeset merged from different branches with pretty controversial changes mixed with a few tests. I'm slowly reading the other part of the implementation in search of a pieces, which I can isolate clearly and cover by tests. Unsuccessfully so far.

Maybe I'll rebase the branch, if I found a way to write cleaner set of tests for the significant portion of the protocol. But the overall idea of those "Compatiblity tests" is covering the [protocol structures](https://pybitmessage-test.readthedocs.io/en/doc/protocol.html#common-structures), excluding those related to crypto operations (the #1796 part).

g1itch:
> maybe also "random" refactoring separately.

Yes, the random is probably closer to the refactoring of the crypto operations. But this particular approach is a bit problematic, so I don't merge those changes into the crypto branch and so far neither I see a way to isolate the related changes into a separate PR. So I'm going to rename the branch and replace it by the further changes when they are ready.

Issue #1710, opened 2021-01-09
SQL unit tests


PeterSurda:
- there should be a mechanism to do unit test for SQL calls outside of the core tests. Maybe have a setup method that spawns a sql thread
- the migrations between the versions need to have their own unit test, maybe these steps should be split into separate functions
- there should be some sample SQL dumps that are loaded before tests

PR #1794, opened 2021-07-28
Test refactor sqlthread

cis-muzahid:
## Repository contributions to the PyBitmessage project

- You can get paid for merged commits if you register at [Tip4Commit](https://tip4commit.com/github/Bitmessage/PyBitmessage)

### Code

- Try to refer to github issue tracker or other permanent sources of discussion about the issue.
- It is clear from the diff *what* you have done, it may be less clear *why* you have done it so explain why this change is necessary rather than what it does

### Documentation

- If there has been a change to the code, there's a good possibility there should be a corresponding change to the documentation 
- If you can't run `fab build_docs` successfully, ask for someone to run it against your branch

### Tests

- If there has been a change to the code, there's a good possibility there should be a corresponding change to the tests
- If you can't run `fab tests` successfully, ask for someone to run it against your branch

## Translations

- For helping with translations, please use [Transifex](https://www.transifex.com/bitmessage-project/pybitmessage/).
- There is no need to submit pull requests for translations.
- For translating technical terms it is recommended to consult the [Microsoft Language Portal](https://www.microsoft.com/Language/en-US/Default.aspx).

### Gitiquette

- Make the pull request against the ["v0.6" branch](https://github.com/Bitmessage/PyBitmessage/tree/v0.6)
- Use references to tickets, e.g. `addresses #123` or `fixes #234` in your commit messages
- Try to use a good editor that removes trailing whitespace, highlights potential python issues and uses unix line endings
- If for some reason you don't want to use github, you can submit the patch using Bitmessage to the "bitmessage" chan, or to one of the developers.


src/class_sqlThread.py
```diff
@@ -29,22 +18,144 @@
...
+    from .bmconfigparser import BMConfigParser
+    from .debug import logger
+    from .addresses import encodeAddress
 
+# pylint: disable=attribute-defined-outside-init,protected-access
+root_path = os.path.dirname(os.path.dirname(__file__))
 
-class sqlThread(threading.Thread):
+
+def connection_build():
```
PeterSurda, line 37: this should contain an argument that allows to use an in memory database instead of the file.

src/class_sqlThread.py
```diff
@@ -29,22 +18,144 @@
...
+    cur = conn.cursor()
+    return conn, cur
+
+
+class UpgradeDB(object):
+    """
+        Upgrade Db with respect to versions
+    """
+
+    conn, cur = connection_build()
```
PeterSurda, line 52: here allow to call it optionally with a parameter for in memory database

src/class_sqlThread.py
```diff
@@ -29,22 +18,144 @@
...
 
-    def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements
+    def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements,
+        # Redefinition-of-parameters-type-from-tuple-to-str, R0204, line-too-long, E501
         """Process SQL queries from `.helper_sql.sqlSubmitQueue`"""
         helper_sql.sql_available = True
-        self.conn = sqlite3.connect(state.appdata + 'messages.dat')
-        self.conn.text_factory = str
-        self.cur = self.conn.cursor()
+        self.conn, self.cur = connection_build()
```
PeterSurda, line 158: this should be in parent class

src/class_sqlThread.py
```diff
@@ -102,34 +213,7 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
-            parameters = ''
-            self.cur.execute(item, parameters)
-
-            item = '''ALTER TABLE sent ADD encodingtype int DEFAULT '2' '''
-            parameters = ''
-            self.cur.execute(item, parameters)
-            self.conn.commit()
-
-            settingsversion = 4
+        settingsversion = self.earlier_setting_version(settingsversion)
```
PeterSurda, line 216: this should be in parent class

src/class_sqlThread.py
```diff
@@ -141,8 +225,34 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
         # file so that when we make changes to the database, the database
         # version we are on can stay embedded in the messages.dat file. Let us
         # check to see if the settings table exists yet.
+
```
PeterSurda, line 228: this should all be in parent class

src/class_sqlThread.py
```diff
@@ -465,10 +349,14 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
             else:
                 logger.error(err)
 
-        # Let us check to see the last time we vaccumed the messages.dat file.
-        # If it has been more than a month let's do it now.
+    def check_vaccumed(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements,
```
PeterSurda, line 352: move to parent

src/class_sqlThread.py
```diff
@@ -635,8 +523,45 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
                 helper_sql.sqlReturnQueue.put((self.cur.fetchall(), rowcount))
                 # helper_sql.sqlSubmitQueue.task_done()
 
+    def earlier_setting_version(self, settingsversion):
```
PeterSurda, line 526: move to parent

src/class_sqlThread.py
```diff
@@ -29,22 +18,144 @@
...
+        self.__running.set()  # Set running to True
+
+    def pause(self):
+        self.__flag.clear()  # Set to False to block the thread
+
+    def resume(self):
+        self.__flag.set()  # Set to True, let the thread stop blocking
 
-    def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements
+    def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements,
```
PeterSurda, line 154: only this should be in this class, the other functionality should be refactored into methods and put into parent.

src/class_sqlThread.py
```diff
@@ -29,22 +18,144 @@
...
+    def increment_settings_version(self, level):
+        """
+            Update version with one level
+        """
+        item = '''update settings set value=? WHERE key='version';'''
+        parameters = (level + 1,)
+        self.cur.execute(item, parameters)
+
+
+class sqlThread(threading.Thread, UpgradeDB):
```
PeterSurda, line 134: maybe an `UpgradeDB` object would be an attribute of the `sqlThread` rarther than `sqlThread` inheriting it. That way we isolate the DB stuff and can test it without threading, and the thread would need to use the same interfaces as the test.

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: pass arguments with DB file name

src/class_sqlThread.py
cis-muzahid: Inherit connection vars from UpgradeDB class

src/class_sqlThread.py
cis-muzahid: Changes done,
Refactored code and kept in Upgrade DB,

src/class_sqlThread.py
cis-muzahid: done

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
```diff
@@ -15,128 +15,122 @@
...
+    conn, cur = connection_build('messages.dat')
 
-        # call create_function for encode address
-        self.create_function()
+    def __init__(self):
+        self._current_level = None
+        self.max_level = 11
+        # self.conn, self.cur = connection_build('messages.dat')
+
+    def __get_current_settings_version(self):
```
PeterSurda, line 54: ```@getter something```

src/class_sqlThread.py
```diff
@@ -15,128 +15,122 @@
...
+        Upgrade Db with respect to versions
+    """
+    conn, cur = connection_build('messages.dat')
 
-        # call create_function for encode address
-        self.create_function()
+    def __init__(self):
+        self._current_level = None
+        self.max_level = 11
+        # self.conn, self.cur = connection_build('messages.dat')
```
PeterSurda, line 52: make another class which overrides this to point to in-memory database, this can then be used for tests. You can call it something like `class MockDB(UpgradeDB)`

src/class_sqlThread.py
```diff
@@ -15,128 +15,122 @@
...
 
+# pylint: disable=attribute-defined-outside-init,protected-access
+root_path = os.path.dirname(os.path.dirname(__file__))
 
-class sqlThread(threading.Thread):
-    """A thread for all SQL operations"""
 
-    def __init__(self):
-        threading.Thread.__init__(self, name="SQL")
+def connection_build(file_name):
```
PeterSurda, line 33: Put this inside `UpgradeDB`

src/class_sqlThread.py
```diff
@@ -15,128 +15,122 @@
...
-        # call create_function for encode address
-        self.create_function()
+    def __init__(self):
+        self._current_level = None
+        self.max_level = 11
+        # self.conn, self.cur = connection_build('messages.dat')
+
+    def __get_current_settings_version(self):
+        """
+            Upgrade Db with respect to their versions
```
PeterSurda, line 56: update docstring

src/class_sqlThread.py
```diff
@@ -15,128 +15,122 @@
...
+        """
+        item = '''SELECT value FROM settings WHERE key='version';'''
+        parameters = ()
+        self.cur.execute(item, parameters)
+        return int(self.cur.fetchall()[0][0])
 
+    def _upgrade_one_level_sql_statement(self, file_name):
+        """
+            Execute SQL files and queries
```
PeterSurda, line 65: update docstring

src/class_sqlThread.py
```diff
@@ -15,128 +15,122 @@
...
-            self.cur.execute('''INSERT INTO settings VALUES('lastvacuumtime',?)''', (
-                int(time.time()),))
-            self.cur.execute(
-                '''CREATE TABLE objectprocessorqueue'''
-                ''' (objecttype int, data blob, UNIQUE(objecttype, data) ON CONFLICT REPLACE)''')
-            self.conn.commit()
-            logger.info('Created messages database file')
+            with open(os.path.join(root_path, "src/sql/init_version_{}.sql".format(file_name))) as sql_file:
+                sql_as_string = sql_file.read()
+            self.cur.executescript(sql_as_string)
```
PeterSurda, line 70: we should at least differenciate between SQL exceptions and other exceptions (e.g. file not found)

src/class_sqlThread.py
```diff
@@ -15,128 +15,122 @@
...
         except Exception as err:
             if str(err) == 'table inbox already exists':
-                logger.debug('Database file already exists.')
```
PeterSurda, line 86: this check is only for testing if the database already has been initialised. It wasn't used to check upgrade levels.

src/class_sqlThread.py
```diff
@@ -15,128 +15,122 @@
...
-            self.conn.commit()
-
-            settingsversion = 4
-
-        BMConfigParser().set(
-            'bitmessagesettings', 'settingsversion', str(settingsversion))
-        BMConfigParser().save()
+    @property
+    def sql_schema_version(self):
+        self._current_level = self.__get_current_settings_version()
```
PeterSurda, line 81: this should return the version, not store

src/class_sqlThread.py
```diff
@@ -15,128 +15,122 @@
...
+
+    @sql_schema_version.setter
+    def sql_schema_version(self, cur, conn, current_level):
+        """
+            Update version with one level
+        """
+        item = '''update settings set value=? WHERE key='version';'''
+        parameters = (current_level + 1,)
+        self.cur.execute(item, parameters)
+        self._current_level = self.__get_current_settings_version()
```
PeterSurda, line 91: should not store

src/class_sqlThread.py
```diff
@@ -15,128 +15,122 @@
...
+    @property
+    def sql_schema_version(self):
+        self._current_level = self.__get_current_settings_version()
+
+    @sql_schema_version.setter
+    def sql_schema_version(self, cur, conn, current_level):
+        """
+            Update version with one level
+        """
+        item = '''update settings set value=? WHERE key='version';'''
```
PeterSurda, line 88: use atomic increment

```
'''UPDATE settings SET value=value+1 WHERE key='version';'''
````

But check for types (string vs. int).

File: src/class_sqlThread.py [hunk 20]
```diff
@@ -15,128 +15,122 @@
...
+        self.conn = conn
+        self.cur = cur
+        if current_level:
+            self._current_level = current_level
+        else:
+            self._current_level = self.__get_current_settings_version()
+        self.max_level = 11
+
+        # call upgrading level in loop
+        for l in range(self._current_level, self.max_level):
```
PeterSurda, line 109:
```
while self.settings_version < self.max_level:
    self._upgrade_one_level_sql_statement(self.settings_version)
    self.settings_version = 1
```

src/class_sqlThread.py
```diff
@@ -15,128 +15,122 @@
...
+            if int(l) == 1:
+                continue
+            self._upgrade_one_level_sql_statement(l)
+            # self.increment_settings_version(l)
+            self._current_level = l
+
+    def increment_settings_version(self, level):
```
PeterSurda, line 116: no need for this

src/class_sqlThread.py
```diff
@@ -15,128 +15,122 @@
...
+        """
+            Update version with one level
+        """
+        item = '''update settings set value=? WHERE key='version';'''
+        parameters = (level + 1,)
+        self.cur.execute(item, parameters)
 
-        helper_startup.updateConfig()
+    def embaded_version(self):
```
PeterSurda, line 124: typo

PeterSurda: should be called something like `init_database`

src/class_sqlThread.py
PeterSurda: different name, to show it's related to `BMConfigParser` version rather than sql version

src/class_sqlThread.py
```diff
@@ -629,11 +401,158 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+
+        # call create_function for encode address
+        self.create_function(self.conn)
+
+        self.initiate_database(self.cur, self.conn)
+
+        # If the settings version is equal to 2 or 3 then the
+        # sqlThread will modify the pubkeys table and change
+        # the settings version to 4.
+        settingsversion = BMConfigParser().getint(
```
PeterSurda, line 470: write getter + setter for this

src/class_sqlThread.py
```diff
@@ -629,11 +401,158 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+        """Process SQL queries from `.helper_sql.sqlSubmitQueue`"""
+        helper_sql.sql_available = True
+        self.conn, self.cur = UpgradeDB().conn, UpgradeDB().cur
+
+        self.cur.execute('PRAGMA secure_delete = true')
+
+        # call create_function for encode address
+        self.create_function(self.conn)
+
+        self.initiate_database(self.cur, self.conn)
```
PeterSurda, line 465: Something like this to skip init when running tests:
```
if not self.instanceof(MockDB):
   self.initiate_database()
```

src/debug.py
```diff
@@ -41,9 +42,12 @@
...
 
 from six.moves import configparser
-
-import helper_startup
-import state
+try:
+    import helper_startup
+    import state
+except ImportError:
+    from . import helper_startup
```
PeterSurda, line 49: puth these lines together

PeterSurda (changes requested):

src/class_sqlThread.py
cis-muzahid: Already using getter for the same

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Change name to ```upgrade_config_parser_setting_version```

src/class_sqlThread.py
cis-muzahid: Moved in to ```__init__```

src/class_sqlThread.py
cis-muzahid: Already using getter for the same

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Differed between
```except IOError as err:``` and ```except Exception as err:```
Not add others exceptions

File: src/class_sqlThread.py (hunk 16 above)
Comment (cis-muzahid) [reply src/class_sqlThread.py:86]: Removed this line

File: src/class_sqlThread.py (hunk 17 above)
Comment (cis-muzahid) [reply src/class_sqlThread.py:81]: return value

File: src/debug.py (hunk 25 above)
Comment (g1itch) [reply src/debug.py:49]: You don't need this at all. Just don't use debug in `sqlThread` and make it like other threads to use own logger, gotten in the class definition, look: https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/network/threads.py#L9

I reverted it once already: b75585f2

File: src/class_sqlThread.py (hunk 16 above)
Comment (g1itch) [reply src/class_sqlThread.py:86]: Why don't you add the exception type?

**Comment by g1itch** [comment]
Your `sql` directories are not python packages and should not contain the empty `__init__` modules. If you want to install `SQL` scripts with the app, put them into the package data: https://github.com/Bitmessage/PyBitmessage/blob/v0.6/setup.py#L137
This doesn't apply to tests, they are not installed, and you may find the files based on `os.path.abspath(__file__)`

File: src/tests/sql/init_version_5.sql [hunk 26]
```diff
@@ -0,0 +1,10 @@
+DROP TABLE objectprocessorqueue;
+
+CREATE TABLE IF NOT EXISTS `knownnodes` (
```
g1itch, line 3: Cool (; I haven't seen this used
```
$ sqlite3 .config/PyBitmessage/messages.dat
SQLite version 3.34.1 2021-01-20 14:10:07
Enter ".help" for usage hints.
sqlite> select * from knownnodes;
Error: no such table: knownnodes
```

src/tests/sql/init_version_5.sql
```diff
@@ -0,0 +1,10 @@
+DROP TABLE objectprocessorqueue;
```
g1itch, line 1: Why drop here, if you are going to create it later?

src/tests/sql/init_version_7.sql
```diff
@@ -0,0 +1,45 @@
+DROP TABLE `inventory`;
+
+CREATE TABLE IF NOT EXISTS `inventory` (
+  `hash` blob NOT NULL,
+  `objecttype` int DEFAULT NULL,
+  `streamnumber` int NOT NULL,
+  `payload` blob DEFAULT NULL,
+  `integer` integer NOT NULL,
+  UNIQUE(hash) ON CONFLICT REPLACE
+) ;
```
g1itch, line 10: I don't see a difference from the previous version - in the `init_version_6.sql`

src/debug.py
PeterSurda: This PR is still a WIP, you don't need to look at it yet, except in some cases like these imports which are not my strength.

src/tests/sql/init_version_7.sql
PeterSurda: I haven't reviewed the SQL files yet, I'm focusing on getting the structure of classes and tests right, it's already at like 3rd iteration.

src/debug.py
g1itch: I receive a notification from this PR, because it was not originally a draft, it was me, who made it draft.

PeterSurda:
@g1itch This PR should among other things separate the SQL execution from threads, so that you can run the tests without launching the SQL thread. There is a lot of work and that's why it's taking long.

g1itch:
> @g1itch This PR should among other things separate the SQL execution from threads, so that you can run the tests without launching the SQL thread. There is a lot of work and that's why it's taking long.

I see. I have no problem with the speed yet. I'm glad to see many line of code deleted in this PR. But I'd prefer to report the strange things, I can see in the code before it got merged. Because I believe we can broke it harder than before. More tests will help to avoid such case.

g1itch:
By the way, the core test suite reports an interesting issue here: https://buildbot.bitmessage.org/#builders/25/builds/1515

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: managed in earlier logic used in memory for test mocking

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: returned that value

src/class_sqlThread.py
cis-muzahid: convert into this str(int(current_level) + 1)

src/class_sqlThread.py
cis-muzahid: Changes are done for that and, but for some query I applied same query  script.

src/class_sqlThread.py
cis-muzahid: Applied 
```
        while self._current_level < self.max_level:
            print(self._current_level)
            self._upgrade_one_level_sql_statement(int(self._current_level))
            self._current_level = self._current_level + 1```

File: src/class_sqlThread.py (hunk 21 above)
Comment (cis-muzahid) [reply src/class_sqlThread.py:133]: Removed that

File: src/class_sqlThread.py [hunk 29]
```diff
@@ -26,117 +26,134 @@
...
+    """
+        Upgrade Db with respect to versions
+    """
 
+    def __init__(self):
+        self._current_level = None
+        self.max_level = 11
+        self.conn, self.cur = self.connection_build('messages.dat')
+
+    def connection_build(self, file_name=None):
```
PeterSurda, line 43:
```
def connection_build(self, file_name='messages.dat'):
```

src/class_sqlThread.py
```diff
@@ -26,117 +26,134 @@
...
+            sys.stderr.write(
+                'ERROR trying to upgrade database. Error message: %s\n' % str(err))
+            os._exit(0)
+
+    @property
+    def sql_schema_version(self):
+        return self.__get_current_settings_version()
+
+    @sql_schema_version.setter
+    def sql_schema_version(self, current_level):
```
PeterSurda, line 103: instead of `current_level` use `ignore`

src/class_sqlThread.py
```diff
@@ -26,117 +26,134 @@
...
+        """
+            Update version with one level
+        """
+        item = '''update settings set value=? WHERE key='version';'''
+        parameters = (current_level + 1,)
+        self.cur.execute(item, parameters)
+        self._current_level = self.__get_current_settings_version()
```
PeterSurda, line 110: this shouldn't be needed

src/class_sqlThread.py
```diff
@@ -629,11 +402,169 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+
+        # People running earlier versions of PyBitmessage do not have the
+        # usedpersonally field in their pubkeys table. Let's add it.
+        if settingsversion == 2:
+            item = '''ALTER TABLE pubkeys ADD usedpersonally text DEFAULT 'no' '''
+            parameters = ()
+            self.cur.execute(item, parameters)
+            self.conn.commit()
+
+            settingsversion = 3
```
PeterSurda, line 418: separate getter setter for settingsversion

src/class_sqlThread.py
```diff
@@ -629,11 +402,169 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+    def set_config_settings_version(self, settingsversion):
+        BMConfigParser().set(
+            'bitmessagesettings', 'settingsversion', str(settingsversion))
+        BMConfigParser().save()
+
+    def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements,
+        # Redefinition-of-parameters-type-from-tuple-to-str, R0204, line-too-long, E501
+        """Process SQL queries from `.helper_sql.sqlSubmitQueue`"""
+        helper_sql.sql_available = True
+        self.conn, self.cur = UpgradeDB().conn, UpgradeDB().cur
```
PeterSurda, line 468: insetad just create one object and save it, like

```
self.db = UpgradeDB()
```
and then just reference
`self.db.cur.execute(blabla)`

src/class_sqlThread.py
```diff
@@ -629,11 +402,169 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+
+        self.cur.execute('PRAGMA secure_delete = true')
+
+        # call create_function for encode address
+        self.create_function(self.conn)
```
PeterSurda, line 473: `self.db.create_function(blabla)`

src/class_sqlThread.py
```diff
@@ -629,11 +402,169 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+
+        self.initiate_database(self.cur, self.conn)
```
PeterSurda, line 475: `self.db.initialize_schema()`

src/class_sqlThread.py
```diff
@@ -629,11 +402,169 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+        # BMConfigParser().save()
+
+        helper_startup.updateConfig()
+
+        # From now on, let us keep a 'version' embedded in the messages.dat
+        # file so that when we make changes to the database, the database
+        # version we are on can stay embedded in the messages.dat file. Let us
+        # check to see if the settings table exists yet.
+
+        self.embaded_version()
```
PeterSurda, line 499: typo

src/class_sqlThread.py
```diff
@@ -26,117 +26,134 @@
...
-        self.create_function()
+class UpgradeDB(object):
+    """
+        Upgrade Db with respect to versions
+    """
 
+    def __init__(self):
+        self._current_level = None
+        self.max_level = 11
+        self.conn, self.cur = self.connection_build('messages.dat')
```
PeterSurda, line 41: just
`self._connection_build('messages.dat')`

PeterSurda (changes requested):

src/class_sqlThread.py
cis-muzahid: change into 
```'''update settings set value=value+1 WHERE key='version';'''```

File: src/class_sqlThread.py (hunk 23 above)
Comment (cis-muzahid) [reply src/class_sqlThread.py]: Applied setter and getter

File: src/class_sqlThread.py (hunk 30 above)
Comment (cis-muzahid) [reply src/class_sqlThread.py:103]: removed current_level

File: src/class_sqlThread.py (hunk 31 above)
Comment (cis-muzahid) [reply src/class_sqlThread.py:117]: Now its 
```    @sql_schema_version.setter
    def sql_schema_version(self):
        """
            Update version with one level
        """
        item = '''update settings set value=value+1 WHERE key='version';'''
        self.cur.execute(item)
        self._current_level = self.__get_current_settings_version()
```

src/class_sqlThread.py
cis-muzahid: passed while initialise

src/class_sqlThread.py
```diff
@@ -26,117 +26,134 @@
...
+        """
+        if file_name == "memory":
+            conn = sqlite3.connect(':memory:')
+        elif file_name is None:
+            conn = sqlite3.connect(state.appdata + "messages.dat")
+        else:
+            conn = sqlite3.connect(state.appdata + file_name)
+        conn.text_factory = str
+        cur = conn.cursor()
+        return conn, cur
```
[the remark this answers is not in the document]
cis-muzahid: Done

src/class_sqlThread.py
```diff
@@ -148,277 +164,29 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+        #     '''update sent set status='msgsent' where status='sentmessage'  ''')
+        # self.cur.execute(
+        #     '''update sent set status='doingpubkeypow' where status='findingpubkey'  ''')
+        # self.cur.execute(
+        #     '''update sent set status='broadcastqueued' where status='broadcastpending'  ''')
+        # self.conn.commit()
+
+    def add_new_option(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements
+        """
+            Add new option
```
PeterSurda, line 187: write a meanignful docstring

PeterSurda (changes requested):

src/class_sqlThread.py
cis-muzahid: applied changes method name to ```__connection_build_internal``` and pass file_name

src/class_sqlThread.py
cis-muzahid: Its changed into 
```        item = '''update settings set value=? WHERE key='version';'''
        self.cur.execute(item, self.__get_current_settings_version() + 1)
```
Because its not taking ```value+1```

src/class_sqlThread.py
cis-muzahid: Changed into ```initialize_schema```

src/class_sqlThread.py
cis-muzahid: Changed

src/class_sqlThread.py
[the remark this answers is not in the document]
cis-muzahid:
```    def __init__(self, db_name="messages.dat"):
        self._current_level = None
        self.max_level = 11
        self.conn = None
        self.cur = None
        self.__connection_build_internal(db_name)

    def __connection_build_internal(self, file_name="messages.dat"):
        """
            Stablish SQL connection
        """

        if file_name == "memory":
            self.conn = sqlite3.connect(':memory:')
        else:
            self.conn = sqlite3.connect(state.appdata + file_name)
        self.conn.text_factory = str
        self.cur = self.conn.cursor()
```

src/class_sqlThread.py
```diff
@@ -26,117 +26,133 @@
...
-        self.conn.text_factory = str
-        self.cur = self.conn.cursor()
-
-        self.cur.execute('PRAGMA secure_delete = true')
-
-        # call create_function for encode address
-        self.create_function()
-
+# pylint: disable=attribute-defined-outside-init,protected-access
+root_path = os.path.dirname(os.path.dirname(__file__))
```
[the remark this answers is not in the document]
cis-muzahid: Done

src/class_sqlThread.py
```diff
@@ -26,117 +26,133 @@
...
+
+class UpgradeDB(object):
+    """
+        Upgrade Db with respect to versions
+    """
+
+    def __init__(self, db_name="messages.dat"):
+        self._current_level = None
+        self.max_level = 11
+        self.conn, self.cur = self._connection_build(db_name)
```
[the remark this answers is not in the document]
cis-muzahid: Done

src/class_sqlThread.py
```diff
@@ -26,117 +26,133 @@
...
+
+    @staticmethod
```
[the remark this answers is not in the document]
cis-muzahid: Removed static method

src/class_sqlThread.py
```diff
@@ -26,117 +26,133 @@
...
+            Stablish SQL connection
+        """
+
+        if file_name == "memory":
+            conn = sqlite3.connect(':memory:')
+        else:
+            conn = sqlite3.connect(state.appdata + file_name)
+        conn.text_factory = str
+        cur = conn.cursor()
+        return conn, cur
```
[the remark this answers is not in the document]
cis-muzahid: Done

src/class_sqlThread.py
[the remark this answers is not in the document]
cis-muzahid: Done

src/class_sqlThread.py
```diff
@@ -26,117 +26,133 @@
...
+        Upgrade Db with respect to versions
+    """
+
+    def __init__(self, db_name="messages.dat"):
+        self._current_level = None
+        self.max_level = 11
+        self.conn, self.cur = self._connection_build(db_name)
+
+    @staticmethod
+    def _connection_build(file_name="messages.dat"):
```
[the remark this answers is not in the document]
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: changed to ```upgrade_schema_if_old_version```

src/class_sqlThread.py
```diff
@@ -148,277 +164,29 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+        #     '''update sent set status='doingmsgpow' where status='doingpow'  ''')
+        # self.cur.execute(
+        #     '''update sent set status='msgsent' where status='sentmessage'  ''')
+        # self.cur.execute(
+        #     '''update sent set status='doingpubkeypow' where status='findingpubkey'  ''')
+        # self.cur.execute(
+        #     '''update sent set status='broadcastqueued' where status='broadcastpending'  ''')
+        # self.conn.commit()
+
+    def add_new_option(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements
```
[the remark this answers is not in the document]
cis-muzahid: Done

src/class_sqlThread.py
```diff
@@ -26,117 +26,145 @@
...
+        self.max_level = 11
+        self.conn = None
+        self.cur = None
+        print("=-=-=-=-=")
+        print(db_name)
+        print("=-=-=-=-=")
+        self._connection_build_internal(db_name)
+        self.root_path = os.path.dirname(os.path.dirname(__file__))
+
+    def _connection_build_internal(self, file_name="messages.dat"):
```
PeterSurda, line 49:
```
def __connection_build_internal(self, file_name='messages.dat', memory=False)
    if memory:
         self.conn = sqlite3.connect(':memory:')
    else
         self.conn = sqlite3.connect(os.path.join(state.appdata, file_name))
```

src/class_sqlThread.py
```diff
@@ -26,117 +26,145 @@
...
+        # self.conn = conn
+        # self.cur = cur
+        # if current_level:
+        #     self._current_level = current_level
+        # else:
+        #     self._current_level = self.__get_current_settings_version()
+        # if max_level:
+        #     self.max_level = max_level
+
+        while self.__get_current_settings_version() < self.max_level:
```
PeterSurda, line 132: use getter

PeterSurda:
```
increment = True

while self.settings_version < self.max_level:
   self._upgrade()
   self.settings_version = increment
```

and create an empty sql script for level 1 (one that doesn't change anything), like `SELECT NULL`

src/class_sqlThread.py
```diff
@@ -629,11 +411,256 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+            self.cur.execute(
+                '''CREATE TABLE inventory (hash blob, objecttype int, streamnumber int, payload blob,'''
+                ''' expirestime integer, tag blob, UNIQUE(hash) ON CONFLICT REPLACE)''')
+            self.cur.execute(
+                '''INSERT INTO subscriptions VALUES'''
+                '''('Bitmessage new releases/announcements','BM-GtovgYdgs7qXPkoYaRgrLFuFKz1SFpsw',1)''')
+            self.cur.execute(
+                '''CREATE TABLE settings (key blob, value blob, UNIQUE(key) ON CONFLICT REPLACE)''')
+            self.cur.execute('''INSERT INTO settings VALUES('version','11')''')
+            self.cur.execute('''INSERT INTO settings VALUES('lastvacuumtime',?)''', (
```
PeterSurda, line 479: make this into a separate method

src/tests/test_sqlthread.py
```diff
@@ -5,40 +5,273 @@
...
+    """
+        Filter column from schema
+    """
+    for x in schema:
+        for y in x:
+            if y == column:
+                yield y
+
+
 class TestSqlThread(unittest.TestCase):
```
PeterSurda, line 24: should be called something like `TestUpgradeDB`

src/class_sqlThread.py
```diff
@@ -26,117 +26,145 @@
...
+        self._current_level = None
+        self.max_level = 11
+        self.conn = None
+        self.cur = None
+        print("=-=-=-=-=")
+        print(db_name)
+        print("=-=-=-=-=")
+        self._connection_build_internal(db_name)
+        self.root_path = os.path.dirname(os.path.dirname(__file__))
+
```
PeterSurda, line 48:
```
def _connection_build(self, db_name='messages.dat'):
    self.__connection_build_internal(file_name=db_name)
```

src/class_sqlThread.py
```diff
@@ -629,11 +411,256 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+    #         self.db.conn.create_function("enaddr", 3, func=encodeAddress, deterministic=True)
+    #     except (TypeError, sqlite3.NotSupportedError) as err:
+    #         logger.debug(
+    #             "Got error while pass deterministic in sqlite create function {}, Passing 3 params".format(err))
+    #         self.db.conn.create_function("enaddr", 3, encodeAddress)
+
+
+class TestDB:
+    def __init__(self):
+        self.db = UpgradeDB('memory')
```
PeterSurda, line 657: no

src/class_sqlThread.py
```diff
@@ -629,11 +411,256 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+    #     """
+    #     try:
+    #         self.db.conn.create_function("enaddr", 3, func=encodeAddress, deterministic=True)
+    #     except (TypeError, sqlite3.NotSupportedError) as err:
+    #         logger.debug(
+    #             "Got error while pass deterministic in sqlite create function {}, Passing 3 params".format(err))
+    #         self.db.conn.create_function("enaddr", 3, encodeAddress)
+
+
+class TestDB:
```
PeterSurda, line 655:
```
class TestDB(UpgradeDB)
```

src/class_sqlThread.py
```diff
@@ -629,11 +411,256 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+    def __init__(self):
+        self.db = UpgradeDB('memory')
+        #     UpgradeDB.__init__(self, "memory")
+
+    def _connection_build(self):
+        # self.db = UpgradeDB('memory')
+        # self._connection_build_internal('memory')
+        return self.db.conn, self.db.cur
```
PeterSurda, line 663:
```
def _connection_build(self, db_name=None)
    self.__connection_build_internal(memory=True)
```

src/tests/test_sqlthread.py
```diff
@@ -5,40 +5,273 @@
...
+        self.assertEqual(result, ['tag'], "Data not migrated for version 3")
+
+    @version
+    def test_sql_thread_version_4(self):
+        """
+            Test with version 4
+        """
+
+        # Assertion
+        self.cur.execute('''select * from inventory where objecttype = 'pubkey';''')
```
PeterSurda, line 176:
```
self.db = TestDB()
res = self.db.execute()
self.assertwhatever
```

src/tests/test_sqlthread.py
```diff
@@ -5,40 +5,273 @@
...
+            # Update versions DB mocking
+            # tables = list(self.cur.execute("select name from sqlite_master where type is 'table'"))
+            # print("-=-=-=-=-=-=-")
+            # print("tables")
+            # print(tables)
+            # print("-=-=-=-=-=-=-")
+            # for q in tables:
+            #     if q:
+            #         print(q)
+
```
PeterSurda, line 106: maybe here in the decorator create the `self.db` object

PeterSurda (changes requested):

src/sql/init_version_2.sql
```diff
@@ -0,0 +1,55 @@
+ --
+ -- Temp Table structure for table `inventory_backup`
+ --
+
+ CREATE TEMP TABLE `inventory_backup` (
+   `hash` blob NOT NULL,
+   `objecttype` text DEFAULT NULL,
+   `streamnumber` int NOT NULL,
+   `receivedtime` int NOT NULL,
```
PeterSurda, line 9: `receivedtime` and `payload` are swapped

src/sql/init_version_2.sql
```diff
@@ -0,0 +1,55 @@
...
+ --
+ --  Table structure for table `inventory`
+ --
+
+
+ CREATE TABLE `inventory` (
+   `hash` blob NOT NULL,
+   `objecttype` text DEFAULT NULL,
+   `streamnumber` int NOT NULL,
+   `receivedtime` int NOT NULL,
```
PeterSurda, line 39: `receivedtime` and `payload` are swapped

PeterSurda (changes requested):

src/tests/test_sqlthread.py
cis-muzahid: Done

cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
cis-muzahid: Changed approach

src/class_sqlThread.py
cis-muzahid: Done

src/class_sqlThread.py
```diff
@@ -26,117 +26,139 @@
...
+            self.conn = sqlite3.connect(':memory:')
+        else:
+            self.conn = sqlite3.connect(state.appdata + file_name)
         self.conn.text_factory = str
         self.cur = self.conn.cursor()
 
-        self.cur.execute('PRAGMA secure_delete = true')
-
-        # call create_function for encode address
-        self.create_function()
```
PeterSurda, line 46: To make life easier, this should remain.

src/tests/test_sqlthread.py
```diff
@@ -5,40 +5,271 @@
...
+
+class TestBitmessageDB(unittest.TestCase):  # pylint: disable=protected-access
     """Test case for SQL thread"""
 
+    # query file path
+    __name__ = None
+    root_path = os.path.dirname(os.path.dirname(__file__))
+    # up = UpgradeDB('memory')
+    test_db = TestDB()
```
PeterSurda, line 31: remove this

src/tests/test_sqlthread.py
```diff
@@ -5,40 +5,271 @@
...
+    conn, cur = test_db._connection_build()  # pylint: disable=protected-access
```
PeterSurda, line 32: remove this

src/tests/test_sqlthread.py
```diff
@@ -5,40 +5,271 @@
...
+
     @classmethod
     def setUpClass(cls):
```
PeterSurda, line 35: remove this method

src/tests/test_sqlthread.py
```diff
@@ -5,40 +5,271 @@
...
+        #         cls.cur.execute("drop table if exists %s" % q)
+        # print("flashing db end")
+        pass
+
+    @classmethod
+    def setUp(cls):
+        """
+            Drop all tables before each test case start
+        """
+        tables = list(cls.cur.execute("select name from sqlite_master where type is 'table'"))
```
PeterSurda, line 53:
```
cls.test_db = TestDB()
```

src/tests/test_sqlthread.py
```diff
@@ -5,40 +5,271 @@
...
+        for q in tables:
+            if q:
+                cls.cur.execute("drop table if exists %s" % q)
+        cls.test_db.create_sql_function()
+        cls.test_db.initialize_schema()
+        cls.cur.execute('''INSERT INTO settings VALUES('version','2')''')
+
+    @classmethod
+    def tearDown(cls):
```
PeterSurda, line 62: delete his method

src/tests/test_sqlthread.py
```diff
@@ -5,40 +5,271 @@
...
+        pass
 
     @classmethod
     def tearDownClass(cls):
```
PeterSurda, line 66: delete this method

PeterSurda (changes requested):

PeterSurda (changes requested):
There are still "DEFAULT" and "NOT NULL" declarations in some places where they are missing in the old ones.

src/class_sqlThread.py
```diff
@@ -629,11 +367,119 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+
+    def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements,
+        # Redefinition-of-parameters-type-from-tuple-to-str, R0204, line-too-long, E501
+        """Process SQL queries from `.helper_sql.sqlSubmitQueue`"""
+
+        self.db = BitmessageDB()
+
+        helper_sql.sql_available = True
+
+        self.db.initialize_schema()
```
PeterSurda, line 446: if table inbox not already exists, then initialize_schema

PeterSurda:
Fixes #1710

PeterSurda (changes requested):
The comments in the `sql/init_version_*.sql` are not helpful. They just repeat the statements in different words. Instead, the original comments should be kept.

src/tests/test_sqlthread.py
```diff
@@ -5,40 +5,237 @@
...
-        sqlLookup = sqlThread()
-        sqlLookup.daemon = True
-        sqlLookup.start()
-        sql_ready.wait()
+    def setUp(cls):
+        """
+            Drop all tables before each test case start
+        """
+        cls.test_db = TestDB()
+        cls.conn, cls.cur = cls.test_db._connection_build()  # pylint: disable=protected-access
```
PeterSurda, line 41: why do you need this? how about just using `cls.test_db.conn` and `cls.test_db.cur`?

PeterSurda:
`check_vacuumed` should end at line 229. Line 231 should be moved to parent method, and the `while` loop that follows should have its own function and also be called from parent.

src/class_sqlThread.py
PeterSurda: `vacuum` not `vaccum`

setup.py
```diff
@@ -137,7 +137,7 @@ def run(self):
         package_data={'': [
             'bitmessageqt/*.ui', 'bitmsghash/*.cl', 'sslkeys/*.pem',
             'translations/*.ts', 'translations/*.qm',
-            'images/*.png', 'images/*.ico', 'images/*.icns'
+            'images/*.png', 'images/*.ico', 'images/*.icns', 'sql/*'
```
PeterSurda, line 140: `sql/*.py`, `sql/*.sql` so that it doesn't pick up things like `__init__.pyc`

src/class_sqlThread.py
```diff
@@ -629,11 +367,124 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+        # Bitmessage users or modify the SQLite database? Add it right
+        # above this line!
+
+        self.db.check_columns_can_store_binary_null()
+
+        # Let us check to see the last time we vaccumed the messages.dat file.
+        # If it has been more than a month let's do it now.
+
+        self.db.check_vacuum()
+
```
PeterSurda, line 479:
```
# Now we're ready to process queue
```

src/tests/sql/init_version_5.sql
PeterSurda: I think the table was deleted in a later version.

src/class_sqlThread.py
```diff
@@ -490,150 +233,253 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+        self.db.upgrade_to_latest()
+
+        self.db.check_columns_can_store_binary_null()
+
+        self.db.check_vacuum()
+
+        helper_sql.sql_ready.set()
+
+        while True:
```
PeterSurda, line 475: something like this

```
while self.loop_queue():
    pass
```

src/class_sqlThread.py
```diff
@@ -490,150 +233,253 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+                                "Disk full"),
+                            _translate(
+                                "MainWindow",
+                                'Alert: Your disk or data storage volume is full. Bitmessage will now exit.'),
+                            True)))
+                    os._exit(0)
+        elif item == 'exit':
+            self.db.conn.close()
+            logger.info('sqlThread exiting gracefully.')
+            return
```
PeterSurda, line 336:
```
return False
```

PeterSurda (changes requested):

src/tests/core.py
```diff
@@ -409,6 +409,22 @@ def test_adding_two_same_case_sensitive_addresses(self):
...
+
+        sqlExecute('create table if not exists testtbl (id integer)')
+        tables = list(sqlQuery("select name from sqlite_master where type is 'table'"))
+        res = [item for item in tables if 'testtbl' in item]
+        self.assertEqual(res[0][0], 'testtbl')
+
+        queryreturn = sqlExecute("INSERT INTO testtbl VALUES(101);")
+        self.assertEqual(queryreturn, 1)
+
+        queryreturn = sqlQuery(''' SELECT * FROM testtbl ''')
```
g1itch, line 423: "SELECT * FROM testtbl"

src/tests/core.py
```diff
@@ -409,6 +409,22 @@ def test_adding_two_same_case_sensitive_addresses(self):
         self.delete_address_from_addressbook(address1)
         self.delete_address_from_addressbook(address2)
 
+    def test_sqlscripts(self):
+        """ Testing sql statements"""
```
g1itch, line 413: What it tests?

g1itch (commented):
Barely readable ); The `tests/sql` seems duplicating `sql`

g1itch (commented):
I see no point in another review since my previous comments are mostly ignored.

src/class_sqlThread.py
PeterSurda: There are two exception hanlders, one for IOError (disk failure, disk full), and everything else. We may want to handle these situations differently, and we don't know what all kinds of exceptions can be thrown, so I recommended doing it this way.

src/class_sqlThread.py
```diff
@@ -17,126 +18,112 @@
...
-
         self.cur.execute('PRAGMA secure_delete = true')
 
-        # call create_function for encode address
-        self.create_function()
+    def __get_current_settings_version(self):
+        """
+            Get current setting Version
+        """
+        item = '''SELECT value FROM settings WHERE key='version';'''
```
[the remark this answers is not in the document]
PeterSurda: There was a problem, we found out what it is, will be fixed in next commit.

src/class_sqlThread.py
```diff
@@ -490,150 +210,252 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+                            _translate(
+                                "MainWindow",
+                                'Alert: Your disk or data storage volume is full. Bitmessage will now exit.'),
+                            True)))
+                    os._exit(0)
+        else:
+            parameters = helper_sql.sqlSubmitQueue.get()
+            rowcount = 0
+            try:
+                self.db.cur.execute(item, parameters)
```
PeterSurda, line 388: `item` is a bad name. Use `query`.

src/class_sqlThread.py
```diff
@@ -490,150 +224,254 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+                    logger.fatal(
+                        'Major error occurred when trying to execute a SQL statement within the sqlThread.'
+                        ' Please tell Atheros about this error message or post it in the forum!'
+                        ' Error occurred while trying to execute statement: "%s"  Here are the parameters;'
+                        ' you might want to censor this data with asterisks (***)'
+                        ' as it can contain private information: %s.'
+                        ' Here is the actual error message thrown by the sqlThread: %s',
+                        str(item),
+                        str(repr(parameters)),
+                        str(err))
```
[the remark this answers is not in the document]
PeterSurda: I tend to agree. Something like `traceback.format_exc` would be better.

src/tests/sql/init_version_5.sql
PeterSurda: you're right, if it's removed and there is an error, it means there is a bug somewhere

src/tests/sql/init_version_5.sql
PeterSurda: It looks like some weird leftover of deleted code. Since the current upgrade code deletes it, we should keep it, otherwise we'd have to test against ancient versions.

src/tests/sql/init_version_7.sql
PeterSurda: I looked through the tests and instructed to redesign how they perform setup.

setup.py
```diff
@@ -137,7 +137,7 @@ def run(self):
         package_data={'': [
             'bitmessageqt/*.ui', 'bitmsghash/*.cl', 'sslkeys/*.pem',
             'translations/*.ts', 'translations/*.qm',
-            'images/*.png', 'images/*.ico', 'images/*.icns'
+            'images/*.png', 'images/*.ico', 'images/*.icns', 'sql/*.py', 'sql/*.sql'
```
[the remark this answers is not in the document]
PeterSurda: The file was removed, but the reference in `setup.py` still persists. @cis-muzahid please delete this reference, and also delete the file `src/tests/sql/__init__.py`, that is also unnecessary.

src/class_sqlThread.py
```diff
@@ -490,150 +224,254 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+                    queues.UISignalQueue.put((
+                        'alert', (
+                            _translate(
+                                "MainWindow",
+                                "Disk full"),
+                            _translate(
+                                "MainWindow",
+                                'Alert: Your disk or data storage volume is full. Bitmessage will now exit.'),
+                            True)))
+                    os._exit(0)
```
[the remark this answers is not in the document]
PeterSurda: I agree

PeterSurda (changes requested):
There are still a couple of unresolved conversations.

PeterSurda (changes requested):
The PR shouldn't remove the `*.sql` files. It should ignore them.

PeterSurda (commented):
**unstage** the sql files

PR #1795, opened 2021-07-28
Fix python3 warnings in test_blindsig

g1itch:
As I wrote before it's as easy as replacing `str()` by `bytes()`.


PeterSurda (approved):
:+1:

Issue #886, opened 2016-07-16
Clarification on backport of pyelliptic requested + external dependency unbundling


ng-0:
As I work for a GNU system which unbundles where possible, I'd like to get some more
clarification on the backporting of pyelliptic you do. Can you explain your changes or point
out in the code which changes are made?

At your choice, could you even move the external dependencies
into git-submodules (other repositories) so it can be up to package
maintainers to pull them in?

This would improve the work of us package maintainers in cutting down the lines of code, especially for package systems like GNU Guix or systems like Gentoo.
https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies
http://dustycloud.org/misc/talks/guix/chicagolug_2015/guix_talk.html (could include reasons and understanding, there are very likely more insights in https://www.gnu.org/software/guix/help/#papers and https://www.gnu.org/software/guix/help/#talks (same page).

PeterSurda:
I don't know what the difference is, you'd have to diff it yourself. You could contact Atheros and ask him.

ng-0:
Thanks for the reply, Peter. I'll just CC @Atheros1 here as I find the github bugtracker not ideal.

But, this is not the only purpose of this issue, the second being bundled dependencies.

bmng-dev:
Newer versions of PyElliptic break the Bitmessage protocol. Bitmessage still uses ECDSA-with-SHA1 signatures and length prefixed coordinate representation for ephemeral public keys while recent versions of PyElliptic only use ECDSA-with-SHA2 signatures and uncompressed point compression form representation for ephemeral public keys. The bundled PyElliptic has an improved OpenSSL library locator.

The bundled Socksipy has been modified to support Tor's SOCKS Extensions.

rekado:
Do you plan to send your improvements to the respective upstream projects?

bmng-dev:
Me personally? No, I didn't make the modifications.

PeterSurda:
I modified the socksipy, I added support for the RESOLVE extension used by Tor. This is used for improving bootstrapping. I can try to push it upstream.

PyElliptic was probably modified by Atheros (I wasn't involved in the development at that time so I don't know).

PeterSurda:
[...]

PR #1796, opened 2021-07-28
Refactor using of crypto functions

g1itch:
In this branch I'm gonna cover by tests all the crypto functions used in code and do a refactoring. Basic points:

 * only `highlevelcrypto` should import from `pyelliptic`
 * `highlevelcrypto` should not depend on the app state e.g. module `state` or `BMConfigparser()` instance (the same will be done for `protocol` in #1788)
 * `highlevelcrypto` should include hashes - double sha512, ripemd160
 * finally I'm gonna detach `pyelliptic` into the separate repo to show a possible way of addressing #886

### DIFF
--- a/src/highlevelcrypto.py
+++ b/src/highlevelcrypto.py
@@ -8,6 +8,7 @@ High level cryptographic functions based on `.pyelliptic` OpenSSL bindings.
 """
 
 import hashlib
+import os
 from binascii import hexlify
 
 import pyelliptic
@@ -57,6 +58,16 @@ def encodeWalletImportFormat(privKey):
     privKey = b'\x80' + privKey
     checksum = hashlib.sha256(hashlib.sha256(privKey).digest()).digest()[0:4]
     return a.changebase(privKey + checksum, 256, 58)
+
+
+# Random
+
+def randomBytes(n):
+    """Get n random bytes"""
+    try:
+        return os.urandom(n)
+    except NotImplementedError:
+        return OpenSSL.rand(n)
 
 
 def makeCryptor(privkey):
--- a/src/highlevelcrypto.py
+++ b/src/highlevelcrypto.py
@@ -30,6 +30,35 @@ def calculateInventoryHash(data):
     return double_sha512(data)[:32]
 
 
+# WIF (uses arithmetic ):
+def decodeWalletImportFormat(WIFstring):
+    """
+    Convert private key from base58 that's used in the config file to
+    8-bit binary string.
+    """
+    fullString = a.changebase(WIFstring, 58, 256)
+    privkey = fullString[:-4]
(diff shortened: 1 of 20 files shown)


src/tests/test_addresses.py
```diff
@@ -33,3 +35,26 @@ def test_encode(self):
         self.assertEqual(
             sample_address,
             addresses.encodeAddress(2, 1, unhexlify(sample_ripe)))
+
+    def test_wif(self):
+        """Decode WIFs of [chan] bitmessage and check the keys"""
+        self.assertEqual(
+            sample_wif_privsigningkey,
+            highlevelcrypto.decodeWalletImportFormat(
+                b'5K42shDERM5g7Kbi3JT5vsAWpXMqRhWZpX835M2pdSoqQQpJMYm'))
```
PeterSurda, line 44: these binary data should also be a variable in samples.

src/tests/test_crypto.py
```diff
@@ -62,6 +64,32 @@ def _hashdigest(data):
...
 
+    # def test_sign(self):
+    #     """Check the signature of the sample_msg created with sample key"""
+    #     self.assertEqual(
+    #         highlevelcrypto.sign(sample_msg, sample_privsigningkey), sample_sig)
+
+    def test_double_sha512(self):
+        """Reproduce the example on page 1 of the Specification"""
+        self.assertEqual(
+            highlevelcrypto.double_sha512(b'hello'), sample_double_sha512)
```
PeterSurda, line 75: ideally `b'hello'` should also be a variable in samples

PeterSurda:
overall nice work, although I haven't investigated it in depth

src/tests/test_crypto.py
g1itch: It's from here: https://pybitmessage-test.readthedocs.io/en/doc/protocol.html#hashes

g1itch:
> overall nice work, although I haven't investigated it in depth

It seems that I misplaced (or misnamed) the signatures. I was going to regenerate them. And there is also a lot of unfinished changes here, mentioned in the PR description.

src/highlevelcrypto.py
```diff
@@ -7,17 +7,85 @@
...
+    priv = randomBytes(32)
+    pub = pointMult(priv)
+    return priv, pub
+
+
+def deterministic_keys(passphrase, nonce):
+    """Generate keys from *passphrase* and *nonce* (encoded as varint)"""
+    priv = hashlib.sha512(passphrase + nonce).digest()[:32]
+    pub = pointMult(priv)
+    return priv, pub
```
g1itch, line 86: I'm not sure if these functions should return the pair or only the private key.

PR #1797, opened 2021-08-02
Python3 port for test_process.py

kdcis:
- Enabled Python3

- Fix warning: Unclosed File Error

### DIFF
--- a/src/tests/test_process.py
+++ b/src/tests/test_process.py
@@ -87,7 +87,8 @@ class TestProcessProto(unittest.TestCase):
     def _get_readline(cls, pfile):
         pfile = os.path.join(cls.home, pfile)
         try:
-            return open(pfile, 'rb').readline().strip()
+            with open(pfile, 'rb') as p:
+                return p.readline().strip()
         except (OSError, IOError):
             pass
 


src/tests/test_process.py
```diff
@@ -87,7 +84,8 @@ def setUp(self):
     def _get_readline(cls, pfile):
         pfile = os.path.join(cls.home, pfile)
         try:
-            return open(pfile, 'rb').readline().strip()
```
PeterSurda, line 90: good

src/pybitmessage
```diff
@@ -8,4 +8,7 @@ script_file = os.path.join(dist.location, dist.key, 'bitmessagemain.py')
 new_globals = globals()
 new_globals.update(__file__=script_file)
 
-execfile(script_file, new_globals)
```
PeterSurda, line 11: This is probably ok, but other parts of the code refuse to run in python3, or may be not compatible. If you remove the refusals, you'll get more errors and those need to be fixed. However, maybe it makes sense to fix those first before allowing this script to run in python3 as that may be a lot of work.

PeterSurda (changes requested):

g1itch:
It's a bad idea to edit this workaround artifact. The main problem is the circular import in the `network` package and not sorted side effects of such imports as `debug`, `helper_startup`. That all issues are preventing us from using the right script kwarg:  https://github.com/Bitmessage/PyBitmessage/blob/v0.6/setup.py#L171

And also you've already informed about `TestProcessProto` and the blind tests: they are going to be finished last because are using the whole pybitmessage daemon installed in the system.

g1itch:
The main idea of the fake script was to use `pathmagic` in setuptools generated script, or rather something close enough to it, not to reinvent the wheel, but this particular one is taken from gentoo `python-exec` product and it was a miracle that it works everywhere.

PeterSurda:
Only keep the bottom most change, revert everything else.

kdcis:
> Only keep the bottom most change, revert everything else.

Updated.

PeterSurda (changes requested):
Remove first hunk, only keep second hunk.

kdcis:
> Remove first hunk, only keep second hunk.

Updated accordingly.

PeterSurda (approved):

PR #1801, opened 2021-08-11
Ignored bare except warnings

kdcis:
Ignored bare except warnings


PeterSurda:
let's add an ignore commit instead of `BaseException`.

kdcis:
> let's add an ignore commit instead of `BaseException`.

updated ignore bare except commit.

PeterSurda (approved):

PeterSurda (approved):

PR #1805, opened 2021-08-13
Updated code quality binary operator line change & ignore base except warning in protocol.py

kdcis:
Updated binary operator line change
Ignored base except warning

### DIFF
--- a/src/protocol.py
+++ b/src/protocol.py
@@ -450,6 +450,7 @@ def decryptAndCheckPubkeyPayload(data, address):
         try:
             decryptedData = cryptorObject.decrypt(encryptedData)
         except:  # noqa:E722
+            # FIXME: use a proper exception after `pyelliptic.ecc` is refactored.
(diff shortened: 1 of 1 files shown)


src/protocol.py
```diff
@@ -445,7 +445,7 @@ def decryptAndCheckPubkeyPayload(data, address):
             return 'failed'
         try:
             decryptedData = cryptorObject.decrypt(encryptedData)
-        except:
+        except:  # noqa:E722
```
PeterSurda, line 448: Write a "FIXME" that we need to have a test for this.

PeterSurda (changes requested):

src/protocol.py
PeterSurda:
```
  # FIXME: use a proper exception after `pyelliptic.ecc` is refactored.
```

kdcis: Updated

PeterSurda (approved):

PR #1806, opened 2021-08-13
Refactoring crypto base changes

g1itch:
Hello!

This is a meaningful subset of the changes in my branches `compatibility`, `crypto`, `flake8*`. Starting the refactoring from separation of protocol tests, crypto tests and the pyelliptic tests (added more).


src/protocol.py
```diff
@@ -202,15 +202,19 @@ def checkIPv6Address(host, hostStandardFormat, private=False):
...
+    if host == b'\x00' * 15 + b'\x01':
         if not private:
             logger.debug('Ignoring loopback address: %s', hostStandardFormat)
         return False
-    if host[0] == '\xFE' and (ord(host[1]) & 0xc0) == 0x80:
+    try:
+        host = [ord(c) for c in host[:2]]
+    except TypeError:  # python3 has ints already
+        pass
+    if host[0:1] == b'\xFE' and host[1] & 0xc0 == 0x80:
```
PeterSurda, line 213: I tested it, and the comparison between `int` and `bytes` doesn't work, at least in python2. It should probably be
```
if host[0] == 0xfe and host[1] & 0xc0 == 0x80:
```

src/tests/test_crypto.py
```diff
@@ -76,38 +59,16 @@ def _hashdigest(data):
...
-
-    def test_address(self):
-        """Create address and check the result"""
-        from pybitmessage import addresses
-        from pybitmessage.fallback import RIPEMD160Hash
-
-        sha = hashlib.new('sha512')
-        sha.update(sample_pubsigningkey + sample_pubencryptionkey)
-        ripe_hash = RIPEMD160Hash(sha.digest()).digest()
-        self.assertEqual(ripe_hash, unhexlify(sample_ripe))
```
PeterSurda, line 106: I think this test is missing in the new code.

src/tests/test_protocol.py
```diff
@@ -2,25 +2,97 @@
...
+            protocol.network_group('8.8.8.8'))
+
+        test_ip = '0102:0304:0506:0708:090A:0B0C:0D0E:0F10'
+        self.assertEqual(
+            b'\x01\x02\x03\x04\x05\x06\x07\x08\x09\x0A\x0B\x0C',
+            protocol.network_group(test_ip))
+
+        test_ip = 'bootstrap8444.bitmessage.org'
+        self.assertEqual(
+            'bootstrap8444.bitmessage.org',
```
PeterSurda, line 87: perhaps
```
test_ip,
```

src/tests/test_protocol.py
```diff
@@ -2,25 +2,97 @@
...
+            protocol.network_group(test_ip))
+
+        test_ip = 'quzwelsuziwqgpt2.onion'
+        self.assertEqual(
+            test_ip,
+            protocol.network_group(test_ip))
+
+        test_ip = None
+        self.assertEqual(
+            None,
```
PeterSurda, line 97: perhaps also
```
test_ip,
```

PeterSurda (changes requested):
Please make the changes and review comments.

src/protocol.py
g1itch: Oh, I see `host[1]`. In py2 it will be bytes. Perhaps I missed some sample in the `test_checkIPv4Address()`

src/tests/test_crypto.py
g1itch: This part looks like `TestCrypto.test_hash_string()` above. The rest part is in `TestAddresses` `.test_decode()` and `test_encode()`.

src/protocol.py
g1itch: yes, in py2 `host[0]` will be int

g1itch: OK, why don't just write `if host[:2] == b'\xFE\x80'` and what about IPv4 254.128.0.0?

g1itch: And also 169.254.0.0/16

g1itch: Also `test_checkIPv6Address()` should probably fail because the first sample is reserved for documentation:

https://www.ripe.net/participate/member-support/lir-basics/ipv6_reference_card.pdf

src/tests/test_crypto.py
PeterSurda: Ok I missed that, looks ok.

PeterSurda (approved):

PR #1807, opened 2021-08-17
Updated code quality to handle bare except warning changes in class_singleCleaner.py

kdcis:
Added KeyError & RuntimeError for `del state.neededPubkeys[address]`
Added BMConfigParser.safeGetFloat method & test case for it
Added bare except ignore warning

### DIFF
--- a/src/bmconfigparser.py
+++ b/src/bmconfigparser.py
@@ -142,6 +142,15 @@ class BMConfigParser(SafeConfigParser):
                 ValueError, AttributeError):
             return default
 
+    def safeGetFloat(self, section, field, default=0.0):
+        """Return value as float, default on exceptions,
+        0.0 if default missing"""
+        try:
+            return self.getfloat(section, field)
+        except (configparser.NoSectionError, configparser.NoOptionError,
+                ValueError, AttributeError):
+            return default
+
     def safeGet(self, section, option, default=None):
         """Return value as is, default on exceptions, None if default missing"""
         try:
--- a/src/class_singleCleaner.py
+++ b/src/class_singleCleaner.py
@@ -50,14 +50,14 @@ class singleCleaner(StoppableThread):
         timeWeLastClearedInventoryAndPubkeysTables = 0
         try:
             state.maximumLengthOfTimeToBotherResendingMessages = (
-                float(BMConfigParser().get(
-                    'bitmessagesettings', 'stopresendingafterxdays'))
+                BMConfigParser().getfloat(
+                    'bitmessagesettings', 'stopresendingafterxdays')
                 * 24 * 60 * 60
             ) + (
-                float(BMConfigParser().get(
-                    'bitmessagesettings', 'stopresendingafterxmonths'))
+                BMConfigParser().getfloat(
+                    'bitmessagesettings', 'stopresendingafterxmonths')
                 * (60 * 60 * 24 * 365) / 12)
-        except:
+        except:  # noqa:E722
             # Either the user hasn't set stopresendingafterxdays and
             # stopresendingafterxmonths yet or the options are missing
             # from the config file.
@@ -156,8 +156,11 @@ class singleCleaner(StoppableThread):
             # is already present and will not do the POW and send the message
             # because it assumes that it has already done it recently.
             del state.neededPubkeys[address]
-        except:
+        except KeyError:
             pass
+        except RuntimeError:
+            self.logger.warning(
+                "Can't remove %s from neededPubkeys, requesting pubkey will be delayed", address, exc_info=True)
 
         queues.UISignalQueue.put((
             'updateStatusBar',
--- a/src/tests/test_config.py
+++ b/src/tests/test_config.py
@@ -34,3 +34,10 @@ class TestConfig(unittest.TestCase):
             BMConfigParser().safeGetInt('nonexistent', 'nonexistent'), 0)
         self.assertEqual(
             BMConfigParser().safeGetInt('nonexistent', 'nonexistent', 42), 42)
+
+    def test_safeGetFloat(self):
+        """safeGetFloat retuns provided default for nonexistent option or 0.0"""
+        self.assertEqual(
+            BMConfigParser().safeGetFloat('nonexistent', 'nonexistent'), 0.0)
+        self.assertEqual(
+            BMConfigParser().safeGetFloat('nonexistent', 'nonexistent', 42.0), 42.0)


src/class_singleCleaner.py
```diff
@@ -171,7 +171,7 @@ def resendPubkeyRequest(self, address):
             # is already present and will not do the POW and send the message
             # because it assumes that it has already done it recently.
             del state.neededPubkeys[address]
-        except:
+        except:  # noqa:E722
```
PeterSurda, line 174: ```except KeyError:```

src/class_singleCleaner.py
```diff
@@ -57,7 +57,7 @@ def run(self):  # pylint: disable=too-many-branches
                 float(BMConfigParser().get(
                     'bitmessagesettings', 'stopresendingafterxmonths'))
                 * (60 * 60 * 24 * 365) / 12)
-        except:
+        except:  # noqa:E722
```
PeterSurda, line 60: add a method `BMConfigParser.safeGetFloat`, also add a default value into `BMConfigDefaults` and then there is no need for error handler because it will handle missing as well as invalid values.

src/class_singleCleaner.py
PeterSurda: **This is just an explanation**

If we somehow manage to trigger a `RuntimeError`, this will raise an exception, but ignoring it will also cause trouble down the line as the key won't be requested. But because this function is called inside custom garbage collection, it shouldn't happen. If it happens, we'll have BM crashing with `RuntimeError` and we'll know what's happening, as opposed to silently not requesting a key.

PeterSurda: Add a separate handler for RuntimeError, which will print a log entry with "warning" severity, saying that "Can't remove key in \_\_file\_\_ \_\_class\_\_ \_\_line\_\_, this shouldn't happen but if it does it prevents a pubkey from being requested".

PeterSurda (changes requested):

src/class_singleCleaner.py
kdcis: Updated.

src/class_singleCleaner.py
kdcis: Updated.

src/bmconfigparser.py
```diff
@@ -142,6 +144,15 @@ def safeGetInt(self, section, field, default=0):
...
 
+    def safeGetFloat(self, section, field, default=0.0):
+        """Return value as float, default on exceptions,
+        0.0 if default missing"""
+        try:
+            return self.getfloat(section, field)
+        except (configparser.NoSectionError, configparser.NoOptionError,
+                ValueError, AttributeError):
+            return default
+
```
g1itch, line 155: You need a test for this in `test_config`

src/class_singleCleaner.py
```diff
@@ -48,20 +48,15 @@ class singleCleaner(StoppableThread):
...
-                    'bitmessagesettings', 'stopresendingafterxmonths'))
-                * (60 * 60 * 24 * 365) / 12)
-        except:
-            # Either the user hasn't set stopresendingafterxdays and
-            # stopresendingafterxmonths yet or the options are missing
-            # from the config file.
-            state.maximumLengthOfTimeToBotherResendingMessages = float('inf')
+
+        state.maximumLengthOfTimeToBotherResendingMessages = (
+            float(BMConfigParser().safeGetFloat(
```
g1itch, line 53: Why do you cast float to float?

src/class_singleCleaner.py
```diff
@@ -171,8 +166,13 @@ def resendPubkeyRequest(self, address):
             # is already present and will not do the POW and send the message
             # because it assumes that it has already done it recently.
             del state.neededPubkeys[address]
-        except:
+        except KeyError:
             pass
+        except RuntimeError:
+            self.logger.warning(
+                'Can\'t remove key in __file__ __class__ __line__,'
```
g1itch, line 173: Very strange message

src/bmconfigparser.py
PeterSurda: I agree

src/class_singleCleaner.py
PeterSurda: tested, additional cast not needed, please remove

src/class_singleCleaner.py
```diff
@@ -48,20 +48,15 @@ class singleCleaner(StoppableThread):
...
-            # stopresendingafterxmonths yet or the options are missing
-            # from the config file.
-            state.maximumLengthOfTimeToBotherResendingMessages = float('inf')
+
+        state.maximumLengthOfTimeToBotherResendingMessages = (
+            float(BMConfigParser().safeGetFloat(
+                'bitmessagesettings', 'stopresendingafterxdays'))
+            * 24 * 60 * 60
+        ) + (
+            float(BMConfigParser().safeGetFloat(
```
PeterSurda, line 57: here too

src/class_singleCleaner.py
PeterSurda: ```self.logger.warning("Can't remove %s from neededPubkeys, requesting pubkey will be delayed", address)```

PeterSurda (changes requested):

src/class_singleCleaner.py
g1itch: Oh, perhaps you can just add `exc_info=True` to the `logger.warning()` call in the later except clause? Than you will have much more info than  `__file__` `__class__` `__line__`

kdcis: Updated

src/bmconfigparser.py
kdcis: Added.

src/class_singleCleaner.py
kdcis: Updated

src/class_singleCleaner.py
kdcis: Updated

src/bmconfigparser.py
```diff
@@ -30,7 +30,9 @@
         "maxuploadrate": 0,
         "apiinterface": "127.0.0.1",
         "apiport": 8442,
-        "udp": "True"
+        "udp": "True",
+        "stopresendingafterxdays": 0.0,
```
PeterSurda, line 34: should be `float(inf)`

PeterSurda, line 34: delete again

src/class_singleCleaner.py
```diff
@@ -48,20 +48,15 @@ class singleCleaner(StoppableThread):
     def run(self):  # pylint: disable=too-many-branches
         gc.disable()
         timeWeLastClearedInventoryAndPubkeysTables = 0
-        try:
-            state.maximumLengthOfTimeToBotherResendingMessages = (
-                float(BMConfigParser().get(
```
PeterSurda, line 53: move the float (instead of casting, do `BMConfigParser().getfloat()`

PeterSurda (changes requested):

src/class_singleCleaner.py
kdcis: Updated

src/bmconfigparser.py
kdcis: reverted

PeterSurda (approved):

PR #1819, opened 2021-08-23
Updated code quality imports placement & bare except warning ignore comments changes in depends.py

kdcis:
Updated imports placement
Added bare except warning ignore comments


src/depends.py
```diff
@@ -3,7 +3,13 @@
 and suggest how it may be installed
 """
 
+import logging
+import os
+import re
+import subprocess
 import sys
+import pyelliptic.openssl
```
PeterSurda, line 11: this isn't ported to python3

PeterSurda (changes requested):
revert pyelliptic, rest of changes can stay, and we'll retry the test

PeterSurda (approved):

src/depends.py
```diff
@@ -3,7 +3,11 @@
 and suggest how it may be installed
 """
 
+import os
+import re
+import subprocess
```
PeterSurda, line 8: introduces in 2.4

src/depends.py
```diff
@@ -3,7 +3,11 @@
...
 import sys
+from importlib import import_module
```
PeterSurda, line 10: introduced in 2.7

PeterSurda (changes requested):
move the 2 imports below the version test

PR #1830, opened 2021-08-26
Formatting l10n module

g1itch:
Another portion of formatting for the code grooming.

### DIFF
--- a/src/l10n.py
+++ b/src/l10n.py
@@ -1,21 +1,31 @@
-"""
-Localization
-"""
+"""Localization helpers"""
+
 import logging
 import os
+import re
+import sys
 import time
 
 from bmconfigparser import BMConfigParser
 
 logger = logging.getLogger('default')
 
-
 DEFAULT_ENCODING = 'ISO8859-1'
 DEFAULT_LANGUAGE = 'en_US'
 DEFAULT_TIME_FORMAT = '%Y-%m-%d %H:%M:%S'
 
-encoding = DEFAULT_ENCODING
-language = DEFAULT_LANGUAGE
+try:
+    import locale
+    encoding = locale.getpreferredencoding(True) or DEFAULT_ENCODING
+    language = (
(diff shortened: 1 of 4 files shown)


src/l10n.py
```diff
@@ -40,55 +50,51 @@
...
+        time.strftime(
+            time_format, (0, 0, 0, 11, 0, 0, 0, 0, 0)).decode(encoding)
+        time.strftime(
+            time_format, (0, 0, 0, 13, 0, 0, 0, 0, 0)).decode(encoding)
         # Check DST
-        unicode(time.strftime(time_format, (0, 0, 0, 0, 0, 0, 0, 0, 1)), encoding)
-    except:
+        time.strftime(
+            time_format, (0, 0, 0, 0, 0, 0, 0, 0, 1)).decode(encoding)
+    except Exception:  # TODO: write tests and determine exception types
```
g1itch, line 80: should be `UnicodeDecodeError` but again, I cannot write a test

src/l10n.py
```diff
@@ -40,55 +50,51 @@
...
     # it looks like some stuff isn't initialised yet when this is called the
     # first time and its init gets the locale settings from the environment
     os.environ["LC_ALL"] = newlocale
 
 
-def formatTimestamp(timestamp=None, as_unicode=True):
+def formatTimestamp(timestamp=None):
     """Return a formatted timestamp"""
     # For some reason some timestamps are strings so we need to sanitize.
     if timestamp is not None and not isinstance(timestamp, int):
```
g1itch, line 100: Here is probably also `ValueError`

g1itch:
More I read this module, a more code seems redundant and written just in case ):

This will be squashed into 2 commits if there will be no objections. The `range` from `six.moves` seems excessive here (#1837); again, nobody reads any PR ):

PeterSurda (approved):
Looks ok, I hope to have windows workers on buildbot soon so that the tests can run there too.

PR #1835, opened 2021-08-27
Updated code quality ignore bare except warning & replaced unicode by decode function in class_smtpServer.py

kdcis:
Added ignore bare except warning
Replaced unicode by decode


src/class_smtpServer.py
```diff
@@ -55,7 +55,7 @@ def smtp_AUTH(self, arg):
                 self.push('235 2.7.0 Authentication successful')
             else:
                 raise Exception("Auth fail")
```
PeterSurda, line 57: maybe a custom exception

PeterSurda (changes requested):
custom exception

PeterSurda (approved):

PR #1836, opened 2021-08-27
Updated code quality replaced unicode by decode function in api.py

kdcis:
Replaced unicode by decode function


g1itch (approved):
obvious

PR #1837, opened 2021-08-27
Updated code quality range by six.moves.range functions in l10n.py

kdcis:
Replaced range by six.moves.range functions

### DIFF
--- a/src/l10n.py
+++ b/src/l10n.py
@@ -100,7 +100,7 @@ def formatTimestamp(timestamp=None):
     if timestamp is not None and not isinstance(timestamp, int):
         try:
             timestamp = int(timestamp)
-        except:
+        except:  # noqa:E722
             timestamp = None
 
     # timestamp can't be less than 0.
--- a/src/l10n.py
+++ b/src/l10n.py
@@ -6,8 +6,9 @@ import re
 import sys
 import time
 
-from bmconfigparser import BMConfigParser
 from six.moves import range
+
+from bmconfigparser import BMConfigParser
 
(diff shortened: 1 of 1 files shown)


PeterSurda:
resolve merge conflicts and check if there's anything left at all in the PR.

g1itch:
> resolve merge conflicts and check if there's anything left at all in the PR.

It's about using the `range()` or `six.moves.range()`. You decide if it's needed here. The code is executed once during the program start.

src/l10n.py
```diff
@@ -7,6 +7,7 @@
 import time
 
 from bmconfigparser import BMConfigParser
+from six.moves import range
```
g1itch, line 10: isort

src/l10n.py
```diff
@@ -100,7 +101,7 @@ def formatTimestamp(timestamp=None):
     if timestamp is not None and not isinstance(timestamp, int):
         try:
             timestamp = int(timestamp)
-        except:
+        except:  # noqa:E722
```
PeterSurda, line 104: `except (ValueError, TypeError)`

PeterSurda (changes requested):

src/l10n.py
kdcis: Updated

src/l10n.py
kdcis: Updated

PeterSurda (approved):

PeterSurda (approved):

PR #1839, opened 2021-09-02
The effects of fc19c54

g1itch:
Hi!

I'm trying to find the possible effects of fc19c54.
I need the buildbot checks with PyQt4 installed to see more, because in my workflow it crashes silently.

### DIFF
--- a/src/class_singleWorker.py
+++ b/src/class_singleWorker.py
@@ -155,7 +155,8 @@ class singleWorker(StoppableThread):
                 try:
                     self.sendBroadcast()
                 except:  # noqa:E722
-                    self.logger.warning("sendBroadcast didn't work")
+                    self.logger.warning(
+                        "sendBroadcast didn't work", exc_info=True)
             elif command == 'doPOWForMyV2Pubkey':
                 try:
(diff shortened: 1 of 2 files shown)


PeterSurda:
The problem appears to be a missing `continue` at `src/class_singleWorker.py` (line 571 in current head).

g1itch:
> The problem appears to be a missing `continue` at `src/class_singleWorker.py` (line 571 in current head).

There are also unneeded nested exception handlers.

PeterSurda:
> There are also unneeded nested exception handlers.

Yes I already gave instructions to reduce these as well.

PS. maybe keep this PR, we do need a test for a broadcast from an address without a key. But you don't have any asserts, maybe there should be an additional status value for such a case that you can check.

g1itch:
This was just a quick hack for an illustration. For the proper test I think a separate test case is needed, outside of the core tests, to not introduce more random failures like in those dontconnect and udp tests.

PeterSurda:
Yea you're right, it's difficult to add tests after weird code has been in production for a long time.

PR #1840, opened 2021-09-04
Minimal useragent validation

g1itch:
Hello!

This may seem a bit controversial patch, but the Protocol [specifies](https://pybitmessage-test.readthedocs.io/en/doc/useragent.html) the exact format for the user agent which turns out to be a [BIP14](https://github.com/bitcoin/bips/blob/master/bip-0014.mediawiki) blueprint, and it in turn refers to RFC1945.

### DIFF
--- a/src/network/bmproto.py
+++ b/src/network/bmproto.py
@@ -5,6 +5,7 @@ Class BMProto defines bitmessage's network protocol workflow.
 import base64
 import hashlib
 import logging
+import re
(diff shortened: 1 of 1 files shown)


PeterSurda (approved):
Reviewing the references it looks the only place where this makes an impact is in the network status tab in the Qt interface, the debug log would still show the actual string. That makes sense to me.

PR #1842, opened 2021-09-15
Updated code quality added bare except warning ignore comment

kdcis:
Added bare except warning ignore comment

### DIFF
--- a/src/upnp.py
+++ b/src/upnp.py
@@ -207,10 +207,7 @@ class uPnPThread(StoppableThread):
 
     def __init__(self):
         super(uPnPThread, self).__init__(name="uPnPThread")
-        try:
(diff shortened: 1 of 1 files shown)


src/upnp.py
```diff
@@ -209,7 +209,7 @@ def __init__(self):
         super(uPnPThread, self).__init__(name="uPnPThread")
         try:
             self.extPort = BMConfigParser().getint('bitmessagesettings', 'extport')
```
PeterSurda, line 211: `safeGetInt`

PeterSurda (changes requested):

src/upnp.py
kdcis: Updated

PeterSurda (approved):

PR #1844, opened 2021-09-16
Added ignore comments for bare except warnings for addresses, class_smtpDeliver, depends, helper_msgcoding, helper_startup, highlevelcrypto, protocol & bitmessagecurses/__init__ files

kdcis:
Added ignore comments for bare except warnings


src/addresses.py
```diff
@@ -45,7 +45,7 @@ def decodeBase58(string, alphabet=ALPHABET):
         for char in string:
             num *= base
             num += alphabet.index(char)
-    except:  # ValueError
+    except:  # noqa:E722  # ValueError
```
PeterSurda, line 48: `ValueError`

src/helper_startup.py
```diff
@@ -55,7 +55,7 @@ def loadConfig():
             logger.info('Loading config files from same directory as program.')
             needToCreateKeysFile = False
             state.appdata = paths.lookupExeFolder()
-        except:
```
PeterSurda, line 58: rewrite without try/except, instead `if not config.safeGet('bitmessagesettings', 'settingsversion'):`
also handle permissions exceptions and read when execuring `.read`

PeterSurda (changes requested):

src/helper_startup.py
PeterSurda: looks like permission exceptions are handled internally, no need for any `try/except` then

PeterSurda (changes requested):

PeterSurda (approved):

PR #1847, opened 2021-09-28
Updated import placement & added bare except ignore comment

kdcis:
Updated import placement
Added bare except ignore comment


src/plugins/notification_notify2.py
```diff
@@ -4,8 +4,8 @@
 """
 
 import gi
-gi.require_version('Notify', '0.7')
 from gi.repository import Notify
+gi.require_version('Notify', '0.7')
```
PeterSurda, line 8: I don't know about this, @g1itch what do you think?

g1itch: It should be strictly before import. Otherwise it's pointless. The notify2 plugin is rarely used in the modern installations and doesn't work in appimage. I keep wondering why you always edit those unmaintained modules with no tests.

PeterSurda: Code quality need to be fixed one way or another so that the tests finally pass. If I wait for tests for every change it will take a decade to get anything done. I focus the tests in more important parts of the code.

g1itch: I mean, if you want to make any progress, maybe it's better to concentrate on the main package, not on the unused or extension code. You could simply exclude those `bitmessagecurses`, `bitmessagecli` and other unmaintained stuff from the lint checks.

Also here in Ukraine, the position "Quality Assurance" implies writing a tests in the first place, so when you mention code quality, the tests are always come in mind. And it's simply strange to edit the code you don't understand (and maybe nobody understands how it's supposed to work because it's stale, unmaintained or unused) without the test coverage. It randomly breaks and you are gonna spend much time debugging it in the future or end up dropping it, confirming the waste of time.

PeterSurda: These are good points, however I can't find sufficiently qualified developers to comprehend the code on their own, and I can't train what I don't know myself (i.e. I would need to find time to understand it first, and I don't have the time). If you know someone suitable for such a position, send me their contact details.

kdcis: Reverted & added ignore comment.

PeterSurda (approved):

PR #1849, opened 2021-09-29
Deleted api_client.py file

kdcis:
Deleted api_client.py file


g1itch:
@PeterSurda do you think this is still needed after the API refactoring and with updated doc (https://pybitmessage-test.readthedocs.io/en/latest/autodoc/pybitmessage.api.html) BTW, the official doc is broken for some reason: https://pybitmessage.readthedocs.io/en/latest/

g1itch:
Also it's weird to print "Uncomment this line.."  because the user who run the script wont see any line.

PeterSurda:
@kdcis I agree with g1itch, maybe add a couple more examples into `src.api.py`, the example there is kind of short and people may not understand right away how to supply arguments and what the result data looks like.

PR #1852, opened 2021-09-30
Updated import placement, print statement, bare except & long line changes

kdcis:
Updated import placement
Updated print statement brackets
Ignored bare except warnings
Updated long line changes


dev/msgtest.py
```diff
@@ -3,25 +3,25 @@
 from pprint import pprint
 import sys
 import traceback
+import messagetypes
```
PeterSurda, line 6: Make this into a unit test, maybe add a couple more test cases. This file can be removed completely after the tests are there.

dev/powinterrupttest.py
```diff
@@ -11,7 +11,7 @@
 
 def signal_handler(signal, frame):
     global shutdown
-    print "Got signal %i in %s/%s" % (signal, current_process().name, current_thread().name)
+    print("Got signal %i in %s/%s" % (signal, current_process().name, current_thread().name))
```
PeterSurda, line 14: This can also be made into a unit test, but in a separate PR.

PeterSurda (changes requested):
make `msgtest.py` into a unit test, otherwise ok.

g1itch:
@PeterSurda is this dir included into the lint checks? As for me, it shouldn't be. I agree that it should be converted to a set of full-fledged unit tests. I tried to do that with `bloomfiltertest` a couple of years ago and came into a conclusion that there is no suitable implementation )`: https://github.com/g1itch/PyBitmessage/tree/tests-bloom

PeterSurda:
@g1itch It isn't included in the checks. I gave instructions to later rewrite these as proper unit tests.

g1itch:
> It isn't included in the checks.

Than the only reason to touch this is inability to find a piece of work, I guess.

src/tests/test_msg.py
```diff
@@ -0,0 +1,43 @@
...
+from pybitmessage import messagetypes
+
+sample_data = {"": "message", "subject": "subject", "body": "body"}
+invalid_data = {"": "message", "subject": b"\x01\x02\x03", "body": b"\x01\x02\x03\x04"}
+
+
+class TestMessageTypes(unittest.TestCase):
+    """A test case for messagetypes"""
+
+    @classmethod
```
PeterSurda, line 15: This can be removed as it just calls parent.

src/messagetypes/message.py
```diff
@@ -1,6 +1,6 @@
 import logging
 
-from messagetypes import MsgBase
+from pybitmessage.messagetypes import MsgBase
```
PeterSurda, line 3: This needs to be tested outside of installed environment.

PeterSurda: Unit test works, but could use more testing.

PeterSurda (changes requested):
Make the requestd changes

src/messagetypes/message.py
kdcis: Tested & working good.

PeterSurda (approved):

PeterSurda (changes requested):
This should be split into 2 PRs.

One is the `print` fixes and removing obsolete files.
One is the messagetypes related stuff

Alternatively, if it's easier, you can split it into 2 commits instead of 2 PRs.

kdcis:
> This should be split into 2 PRs.
> 
> One is the `print` fixes and removing obsolete files. One is the messagetypes related stuff
> 
> Alternatively, if it's easier, you can split it into 2 commits instead of 2 PRs.

Updated commits accordingly, Please check

PeterSurda (approved):

PR #1875, opened 2021-11-02
A working subset of #1788

g1itch:
Hello!

This is a subset of #1788, the tests that are ready (+4 tests, including those for base58 and varint) and corresponding changes in the `protocol` module.

### DIFF
--- a/src/tests/test_addresses.py
+++ b/src/tests/test_addresses.py
@@ -46,11 +46,15 @@ class TestAddresses(unittest.TestCase):
 
     def test_base58(self):
         """Check Base58 encoding and decoding"""
+        self.assertEqual(addresses.decodeBase58('1'), 0)
+        self.assertEqual(addresses.decodeBase58('!'), 0)
         self.assertEqual(
             addresses.decodeBase58(sample_addr4), sample_daddr4_512)
         self.assertEqual(
             addresses.decodeBase58(sample_addr3), sample_daddr3_512)
 
+        self.assertEqual(addresses.encodeBase58(0), '1')
+        self.assertEqual(addresses.encodeBase58(-1), None)
         self.assertEqual(
             sample_addr4, addresses.encodeBase58(sample_daddr4_512))
         self.assertEqual(
--- a/src/tests/test_addresses.py
+++ b/src/tests/test_addresses.py
@@ -4,7 +4,13 @@ from binascii import unhexlify
 
 from pybitmessage import addresses
 
-from .samples import sample_address, sample_ripe
+from .samples import (
+    sample_address, sample_daddr3_512, sample_daddr4_512,
+    sample_deterministic_addr4, sample_deterministic_addr3,
+    sample_deterministic_ripe, sample_ripe)
+
+sample_addr3 = sample_deterministic_addr3.split('-')[1]
+sample_addr4 = sample_deterministic_addr4.split('-')[1]
 
 
 class TestAddresses(unittest.TestCase):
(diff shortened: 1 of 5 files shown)


g1itch:
There is also a test for base58 in #1796: 7f213f6

src/tests/test_addresses.py
```diff
@@ -33,3 +33,21 @@ def test_encode(self):
         self.assertEqual(
             sample_address,
             addresses.encodeAddress(2, 1, unhexlify(sample_ripe)))
+
+    def test_base58(self):
+        """Check Base58 encoding and decoding"""
+        self.assertEqual(
```
PeterSurda, line 39: - these strings and numbers should be in `samples.py`
- to improve coverage, encode needs to test `0`, and decode needs to test an invalid character

src/tests/test_packets.py
```diff
@@ -0,0 +1,68 @@
...
+        self.assertEqual(data, unhexlify('feffffffff'))
+        data = addresses.encodeVarint(4294967296)
+        self.assertEqual(data, unhexlify('ff0000000100000000'))
+        data = addresses.encodeVarint(18446744073709551615)
+        self.assertEqual(data, unhexlify('ffffffffffffffffff'))
+
+        with self.assertRaises(addresses.varintEncodeError):
+            addresses.encodeVarint(18446744073709551616)
+
+        value, lenght = addresses.decodeVarint(b'\xfeaddr')
```
PeterSurda, line 38: `length`, not `lenght`

PeterSurda (changes requested):
Mostly ok, just minor cleanup pls.

src/tests/test_addresses.py
```diff
@@ -17,19 +23,40 @@ def test_decode(self):
...
+        self.assertEqual(addresses.decodeBase58('1'), 0)
+        self.assertEqual(addresses.decodeBase58('!'), 0)
+        self.assertEqual(
+            addresses.decodeBase58(sample_addr4), sample_daddr4_512)
+        self.assertEqual(
+            addresses.decodeBase58(sample_addr3), sample_daddr3_512)
+
+        self.assertEqual(addresses.encodeBase58(1), '2')
+        # infinite loop!
+        # self.assertEqual(addresses.encodeBase58(-1), '')
```
g1itch, line 58: @PeterSurda what do you think `addresses.encodeBase58()` should do in such case? Raise an exception, return None, or drop the sign?

src/tests/test_addresses.py
```diff
@@ -17,19 +23,40 @@ def test_decode(self):
...
             addresses.encodeAddress(2, 1, unhexlify(sample_ripe)))
+        ripe = unhexlify(sample_deterministic_ripe)
+        self.assertEqual(
+            addresses.encodeAddress(3, 1, ripe),
+            'BM-%s' % addresses.encodeBase58(sample_daddr3_512))
+
+    def test_base58(self):
+        """Check Base58 encoding and decoding"""
+        self.assertEqual(addresses.decodeBase58('1'), 0)
+        self.assertEqual(addresses.decodeBase58('!'), 0)
```
g1itch, line 50: This is the invalid character.

src/tests/test_addresses.py
PeterSurda: Good catch, I'd go with `return None`.

src/tests/test_addresses.py
```diff
@@ -17,19 +23,40 @@ def test_decode(self):
...
+        self.assertEqual(
+            addresses.decodeBase58(sample_addr4), sample_daddr4_512)
+        self.assertEqual(
+            addresses.decodeBase58(sample_addr3), sample_daddr3_512)
+
+        self.assertEqual(addresses.encodeBase58(1), '2')
```
PeterSurda, line 56: still missing test for `0`:
`self.assertEqual(addresses.encodeBase58(0), '1')`

PeterSurda (changes requested):
Still `encodeBase58(0)` is missing. Fixing negative can be fixed either in this PR or a separate one.

src/tests/test_addresses.py
g1itch: It failed when I wrote it first time, maybe I misspelled something

g1itch:
> Fixing negative can be fixed either in this PR or a separate one.

It's a simple mechanical fix. I prefer to merge complete tests, not stubs.

PeterSurda (approved):

PR #1883, opened 2021-11-15
Added mock code

kdcis:
Added mock code for class_objectProcessor, class_singleWorker, inventory, connectionpool & stats


g1itch:
Wrong placement. There shouldn't be a `mock` package, because the end user doesn't need it. Please move into tests.

src/mock/class_objectProcessor.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    def __init__(self):
+        threading.Thread.__init__(self, name="objectProcessor")
+        random.seed()
+        # It may be the case that the last time Bitmessage was running,
+        # the user closed it before it finished processing everything in the
+        # objectProcessorQueue. Assuming that Bitmessage wasn't closed
+        # forcefully, it should have saved the data in the queue into the
+        # objectprocessorqueue table. Let's pull it out.
+        sql_ready.wait()
```
g1itch, line 31: it will never happen

src/mock/class_objectProcessor.py
```diff
@@ -0,0 +1,50 @@
...
+        queryreturn = sqlQuery(
```
g1itch, line 32: This is not a mock class because it uses `sqlQuery()` from `helper_sql`, thus depends on the `sqlThread`

PeterSurda: #1794 contains a test sql class (runs in-memory database and allows custom schema initialization), that will be used here as well.

PeterSurda:
> Wrong placement. There shouldn't be a `mock` package, because the end user doesn't need it. Please move into tests.

There are some technical difficulties with such an approach at the moment, needs a bit more work to fix it.

src/mock/class_objectProcessor.py
PeterSurda: For the time being we can skip this, yes.

src/tests/mock/bitmessagemock.py
```diff
@@ -0,0 +1,3 @@
+def main():
+    """Mock main function"""
+    pass
```
PeterSurda, line 3: here it can start launching the mock threads.

src/tests/mock/class_objectProcessor.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class MockObjectProcessor(threading.Thread):
+    """
+    The objectProcessor thread, of which there is only one, receives network
+    objects (msg, broadcast, pubkey, getpubkey) from the receiveDataThreads.
+    """
+    def __init__(self):
+        threading.Thread.__init__(self, name="objectProcessor")
+        random.seed()
```
PeterSurda, line 25: 25 - 41 - remove

PeterSurda (changes requested):

PR #1884, opened 2021-11-17
Use tox for testing inside a container

g1itch:
Hi!

I'm rewriting the `packages/Dockerfile.travis` and the corresponding script to test using tox.

### DIFF
--- a/packages/docker/Dockerfile.bionic
+++ b/packages/docker/Dockerfile.bionic
@@ -17,8 +17,8 @@ RUN apt-get install -yq --no-install-suggests --no-install-recommends \
     # Testing environment
     language-pack-en \
     qtbase5-dev qt5-default tor tox xvfb \
-    # dpkg
-    debhelper dh-apparmor dh-python fakeroot python-stdeb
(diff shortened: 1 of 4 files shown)


g1itch:
I'm still missing something important: surprisingly this reproduces my old issue with shutdown tests in Fedora container.

g1itch:
> I decided to unify the docker files (at least within one distro version) and use multi stage build. Maybe you can look at `Dockerfile.bionic` and create a new stage called "tox" (or "test").

I'm not familiar with the multistage builds. Maybe it's a good Idea. Also I still haven't configured a local buildbot instance to propose some changes in the `master.cfg`. So I decided to start with this test script. What I wanted to show here is that you can have 3 VMs (or containers): xenial?, bionic, focal and use them for all the checks replacing the CMD by `tox`, `tox -e py27-doc` or a script for building a package; then remove all the code related to travis or custom package installation from the `master.cfg`.

Also I don't like the current state of the `Dockerfile.bionic`, it contains many unneeded packages and produces 2.5 GB image.

Issue #1898, opened 2021-12-10
Error on send message via API


teruteru128:
If you send a message to an address registered in the address book via API, an error will be returned.

```log
2021-12-11 02:50:58,436 - ERROR - too many values to unpack
Traceback (most recent call last):
  File "api.py", line 1514, in _dispatch
  File "api.py", line 1487, in _handle_request
  File "api.py", line 1137, in HandleSendMessage
ValueError: too many values to unpack
```

OS: Windows 10 21H1
PyBitmessage: 0.6.3.2 20211207

g1itch:
Could I see your code?

teruteru128:
The first thing you can prepare is written in C, is that okay?

g1itch:
> The first thing you can prepare is written in C, is that okay?

I'm not afraid (;

The line number of your traceback seems to point to decoding of base64 in subject argument

teruteru128:
https://github.com/teruteru128/study/blob/develop/src/bmsendmsg.c
https://github.com/teruteru128/libstudy/blob/master/src/bmapi.c

teruteru128:
Hmm ... should I submit something made with python?
I think it's probably not difficult because it depends on whether there is a corresponding record in the address book using the XML-RPC API.

g1itch:
> The line number of your traceback seems to point to decoding of base64 in subject argument

Oh, I see it's really about the label: https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/api.py#L1137

g1itch:
> Hmm ... should I submit something made with python? I think it's probably not difficult because it depends on whether there is a corresponding record in the address book using the XML-RPC API.

Yes, if you can reproduce it in python. Don't you get this exception when using JSON-RPC? It seems not related.

g1itch:
Perhaps try to remove the comma in the line. Does it help?

Пт, 10 груд. 2021 20:23 користувач Teruteru ***@***.***> пише:

teruteru128:
Um ... I'm sorry, I'm using PyBitmessage distributed as a binary on Windows, so I can't modify the source code and test it right away.

teruteru128:
However, the indication of removing the comma seems correct.

g1itch:
OK, so I have to write a test and merge the fix or at least build a new
continuous release. Maybe tomorrow.

Пт, 10 груд. 2021 20:46 користувач Teruteru ***@***.***> пише:

[...]

PR #1899, opened 2021-12-11
Fixing recently discovered API bug

g1itch:
Hi!

Here will be a test and fix for the API bug in 'sendMessage' command reported in #1898.

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -1134,7 +1134,7 @@ class BMRPCDispatcher(object):
         queryreturn = sqlQuery(
             "SELECT label FROM addressbook WHERE address=?", toAddress)
         try:
-            toLabel, = queryreturn[0][0]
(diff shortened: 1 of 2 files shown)


g1itch:
Reproduced in https://buildbot.bitmessage.org/#builders/24/builds/4901

PeterSurda (approved):

PeterSurda:
Ideally the section which retrieves label from the database would be its own function or method that can be tested separately but this is also good.

g1itch:
> Ideally the section which retrieves label from the database would be its own function or method that can be tested separately but this is also good.

It's a proper commendation, but I think we may use both tests and it's better to be addressed as separate PR. This additional check is just for reproducing the issue in the current code and prevent a regression.

PR #1926, opened 2022-01-25
Override the pylint checks

shekhar-cis:
Disabled `unsubscriptable-object`, `redefined-builtin` and `too-many-return-statements`.

### DIFF
--- a/src/bmconfigparser.py
+++ b/src/bmconfigparser.py
@@ -69,6 +69,7 @@ class BMConfigParser(SafeConfigParser):
             raise ValueError("Invalid value %s" % value)
         return SafeConfigParser.set(self, section, option, value)
 
+    # pylint: disable=redefined-builtin, too-many-return-statements
     def get(self, section, option, raw=False, vars=None):
         if sys.version_info[0] == 3:
             # pylint: disable=arguments-differ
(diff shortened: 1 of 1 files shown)


PeterSurda (approved):

src/bmconfigparser.py
```diff
@@ -187,7 +188,7 @@ def read(self, filenames):
                     try:
                         if not self.validate(
                                 section, option,
-                                self[section][option]
+                                self[section][option]  # pylint: disable=unsubscriptable-object
```
g1itch, line 191: This one might be rewritten using `.get()`. Suppressing pylint warnings is a bad habit.

src/bmconfigparser.py
```diff
@@ -69,6 +69,7 @@ def set(self, section, option, value=None):
             raise ValueError("Invalid value %s" % value)
         return SafeConfigParser.set(self, section, option, value)
 
+    # pylint: disable=redefined-builtin, too-many-return-statements
     def get(self, section, option, raw=False, vars=None):
```
g1itch, line 73: This method shouldn't be written in the first place. It's an unneeded crutch. `redefined-builtin` is not good. Pylint shows `too-many-return-statements` warning because it's poorly written, through a brute force. The same applies to `arguments-differ`. Better to remove it completely and refactor the `BMConfigParser` in a proper way.

PeterSurda:
Thank you @g1itch I'll add this to the queue. I just needed warning suppressed for now.

PR #1927, opened 2022-01-28
Refactor `BMConfigParser` as a module variable

shekhar-cis:
Refactor `BMConfigParser`
- now as a module variable instead of singleton
- can run unit tests
- can be used in mock more easily

### DIFF
--- a/packages/pyinstaller/bitmessagemain.spec
+++ b/packages/pyinstaller/bitmessagemain.spec
@@ -74,7 +74,7 @@ a.datas += [
 
 # append the translations directory
 a.datas += addTranslations()
-
+a.datas += [('default.ini', os.path.join(srcPath, 'default.ini'), 'DATA')]
 
 excluded_binaries = [
     'QtOpenGL4.dll',
--- a/src/bitmessageqt/settings.py
+++ b/src/bitmessageqt/settings.py
@@ -16,7 +16,7 @@ import paths
 import queues
 import state
 import widgets
-from bmconfigparser import BMConfigParser
+from bmconfigparser import config
 from helper_sql import sqlExecute, sqlStoredProcedure
 from helper_startup import start_proxyconfig
 from network import knownnodes, AnnounceThread
@@ -24,11 +24,11 @@ from network.asyncore_pollchoose import set_rates
 from tr import _translate
 
 
-def getSOCKSProxyType(config):
+def getSOCKSProxyType(config_):
     """Get user socksproxytype setting from *config*"""
     try:
         result = ConfigParser.SafeConfigParser.get(
-            config, 'bitmessagesettings', 'socksproxytype')
+            config_, 'bitmessagesettings', 'socksproxytype')
     except (ConfigParser.NoSectionError, ConfigParser.NoOptionError):
         return None
     else:
@@ -45,7 +45,7 @@ class SettingsDialog(QtGui.QDialog):
 
         self.parent = parent
         self.firstrun = firstrun
-        self.config = BMConfigParser()
+        self.config = config
         self.net_restart_needed = False
         self.timer = QtCore.QTimer()
 
@@ -80,7 +80,7 @@ class SettingsDialog(QtGui.QDialog):
             )
         QtGui.QWidget.resize(self, QtGui.QWidget.sizeHint(self))
 
-    def adjust_from_config(self, config):
+    def adjust_from_config(self, config_):
         """Adjust all widgets state according to config settings"""
         # pylint: disable=too-many-branches,too-many-statements
         if not self.parent.tray.isSystemTrayAvailable():
@@ -89,31 +89,31 @@ class SettingsDialog(QtGui.QDialog):
                 "MainWindow", "Tray (not available in your system)"))
             for setting in (
                     'minimizetotray', 'trayonclose', 'startintray'):
-                config.set('bitmessagesettings', setting, 'false')
+                config_.set('bitmessagesettings', setting, 'false')
         else:
             self.checkBoxMinimizeToTray.setChecked(
-                config.getboolean('bitmessagesettings', 'minimizetotray'))
+                config_.getboolean('bitmessagesettings', 'minimizetotray'))
             self.checkBoxTrayOnClose.setChecked(
-                config.safeGetBoolean('bitmessagesettings', 'trayonclose'))
+                config_.safeGetBoolean('bitmessagesettings', 'trayonclose'))
             self.checkBoxStartInTray.setChecked(
-                config.getboolean('bitmessagesettings', 'startintray'))
+                config_.getboolean('bitmessagesettings', 'startintray'))
 
         self.checkBoxHideTrayConnectionNotifications.setChecked(
-            config.getboolean(
+            config_.getboolean(
                 'bitmessagesettings', 'hidetrayconnectionnotifications'))
         self.checkBoxShowTrayNotifications.setChecked(
-            config.getboolean('bitmessagesettings', 'showtraynotifications'))
+            config_.getboolean('bitmessagesettings', 'showtraynotifications'))
 
         self.checkBoxStartOnLogon.setChecked(
-            config.getboolean('bitmessagesettings', 'startonlogon'))
+            config_.getboolean('bitmessagesettings', 'startonlogon'))
 
         self.checkBoxWillinglySendToMobile.setChecked(
-            config.safeGetBoolean(
+            config_.safeGetBoolean(
                 'bitmessagesettings', 'willinglysendtomobile'))
         self.checkBoxUseIdenticons.setChecked(
-            config.safeGetBoolean('bitmessagesettings', 'useidenticons'))
+            config_.safeGetBoolean('bitmessagesettings', 'useidenticons'))
         self.checkBoxReplyBelow.setChecked(
-            config.safeGetBoolean('bitmessagesettings', 'replybelow'))
+            config_.safeGetBoolean('bitmessagesettings', 'replybelow'))
 
         if state.appdata == paths.lookupExeFolder():
             self.checkBoxPortableMode.setChecked(True)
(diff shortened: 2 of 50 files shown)


src/bmconfigparser.py
```diff
@@ -214,22 +123,8 @@ def addresses():
         def read(self, filenames):
             """Read config and populate defaults"""
             self._reset()
+            SafeConfigParser.read(self, os.path.join(os.path.dirname(__file__), 'default.ini'))
             SafeConfigParser.read(self, filenames)
```
PeterSurda, line 127:
```
if filenames:
    SafeConfigParser.read(self.filenames)
```

src/bmconfigparser.py
```diff
@@ -214,22 +123,8 @@ def addresses():
         def read(self, filenames):
```
PeterSurda, line 123:
```
def read(self, filenames=None)
```

src/tests/test_config.py
```diff
@@ -93,5 +94,6 @@ def test_reset(self):
             BMConfigParser().safeGetInt('bitmessagesettings', 'maxaddrperstreamsend'), 100)
```
PeterSurda, line 94: split into 2 tests.

one, tests `_reset`, after reset, the `settingsversion` section must be absent.

second, let's call it `test_defaults`
```
config = BMConfigParser()
config.addsection("bitmessagesettings")
config.set("bitmessagesettings", "maxaddrperstreamsend", 100)
config.read() # this resets plus populates with defaults
assert(config.get("bitmessagesettings", "maxaddrperstreamsend"), 500)
```

src/helper_startup.py
```diff
@@ -39,6 +39,7 @@
 def loadConfig():
     """Load the config"""
     config = BMConfigParser()
+    config.read(os.path.join(os.path.dirname(__file__), 'default.ini'))
```
PeterSurda, line 42: no need this line

src/helper_startup.py
```diff
@@ -70,44 +71,24 @@ def loadConfig():
```
PeterSurda, line 71:
```
config.read() # this populates defaults
```

src/helper_startup.py
```diff
@@ -70,44 +71,24 @@ def loadConfig():
 
         # This appears to be the first time running the program; there is
         # no config file (or it cannot be accessed). Create config file.
-        config.add_section('bitmessagesettings')
-        config.set('bitmessagesettings', 'settingsversion', '10')
```
PeterSurda, line 74: we need this (`settingsversion`)

src/inventory.py
```diff
@@ -14,7 +14,8 @@ class Inventory():
     to manage the inventory.
     """
     def __init__(self):
-        self._moduleName = BMConfigParser().safeGet("inventory", "storage")
+        self.config = BMConfigParser()
+        self._moduleName = self.config.safeGet("inventory", "storage")
```
PeterSurda, line 18: shouldn't be necessary. If it causes problems anyway, it needs to be investigated in more detail

PeterSurda (changes requested):

src/class_sqlThread.py
```diff
@@ -93,6 +94,9 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
         # If the settings version is equal to 2 or 3 then the
         # sqlThread will modify the pubkeys table and change
         # the settings version to 4.
+        if state.appdata:
```
PeterSurda, line 97: ```config_ready.wait()```

src/bmconfigparser.py
```diff
@@ -5,6 +5,7 @@
 import os
 import shutil
 import sys  # FIXME: bad style! write more generally
+import threading
```
PeterSurda, line 8:
```
from threading import Event
```

src/bmconfigparser.py
```diff
@@ -270,3 +169,6 @@ def validate_bitmessagesettings_maxoutboundconnections(value):
         if value < 0 or value > 8:
             return False
         return True
+
+
+config = BMConfigParser()
```
g1itch, line 174: Another place with side effects is bad

PeterSurda: What alternative design do you propose? I think a module variable is an improvement singleton and I don't know other approaches.

src/default.ini
```diff
@@ -0,0 +1,46 @@
...
+apiinterface = 127.0.0.1
+apiport = 8442
+udp = True
+port = 8442
+timeformat = %%c
+blackwhitelist = black
+startonlogon = False
+showtraynotifications = True
+startintray = False
+socksproxytype = None
```
PeterSurda, line 17: probably should be lowercase, `none`

src/bmconfigparser.py
```diff
@@ -209,27 +121,14 @@ def readfp(self, fp, filename=None):
         def addresses():
             """Return a list of local bitmessage addresses (from section labels)"""
             return [
-                x for x in BMConfigParser().sections() if x.startswith('BM-')]
+                x for x in config.sections() if x.startswith('BM-')]
 
-        def read(self, filenames):
+        def read(self, filenames=None):
```
PeterSurda, line 126: is now the same for p2 and p3

src/class_singleWorker.py
```diff
@@ -403,10 +403,10 @@ def sendOutOrStoreMyV4Pubkey(self, myAddress):
         whereas in the past it directly appended it to the outgoing buffer, I think. Same with all the other methods in
         this class.
         """
-        if not BMConfigParser().has_section(myAddress):
+        if not config.has_section(myAddress):
             # The address has been deleted.
             return
-        if shared.BMConfigParser().safeGetBoolean(myAddress, 'chan'):
+        if shared.config.safeGetBoolean(myAddress, 'chan'):
```
PeterSurda, line 409: not sure why `shared`, remove

src/class_singleWorker.py
```diff
@@ -956,7 +956,7 @@ def sendMsg(self):
                 if protocol.isBitSetWithinBitfield(behaviorBitfield, 30):
                     # if we are Not willing to include the receiver's
                     # RIPE hash on the message..
-                    if not shared.BMConfigParser().safeGetBoolean(
+                    if not shared.config.safeGetBoolean(
```
PeterSurda, line 959: also why `shared`?

src/class_sqlThread.py
```diff
@@ -93,7 +94,8 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
         # If the settings version is equal to 2 or 3 then the
         # sqlThread will modify the pubkeys table and change
         # the settings version to 4.
-        settingsversion = BMConfigParser().getint(
+        config_ready.wait()
```
PeterSurda, line 97: no need to wait twice in the same method, remove

src/helper_startup.py
```diff
@@ -38,7 +39,6 @@
 
 def loadConfig():
     """Load the config"""
```
PeterSurda, line 41: add debugging in this method

src/helper_startup.py
```diff
@@ -130,11 +112,11 @@ def loadConfig():
         config.save()
     else:
         updateConfig()
+    config_ready.set()
 
 
 def updateConfig():
```
PeterSurda, line 118: also add debugging in this function

src/bmconfigparser.py
g1itch: For me this approach seems worse than the singleton antipattern. Create the instance of `BMConfigParser()` upon program start and use it everywhere passing as argument to functions or accessing it through a slot of another object.

These changes also create potential conflicts all across the codebase. And current test coverage is definitely not enough for such big changes.

PeterSurda: but where to put it then? `state.py`?

src/default.ini
```diff
@@ -0,0 +1,46 @@
...
+socksusername = None
+sockspassword = None
+keysencrypted = False
+messagesencrypted = False
+minimizeonclose = False
+dontconnect = True
+replybelow = False
+stopresendingafterxdays = None
+stopresendingafterxmonths = None
+
```
PeterSurda, line 30: Add
`opencl = None`

src/default.ini
```diff
@@ -0,0 +1,46 @@
...
+maxaddrperstreamsend = 500
+maxbootstrapconnections = 20
+maxdownloadrate = 0
+maxoutboundconnections = 8
+maxtotalconnections = 200
+maxuploadrate = 0
+apiinterface = 127.0.0.1
+apiport = 8442
+udp = True
+port = 8442
```
PeterSurda, line 11: should be `8444`

PeterSurda (changes requested):

src/bmconfigparser.py
```diff
@@ -173,33 +97,18 @@ def _reset(self):
         for x in sections:
             self.remove_section(x)
 
+    def read(self, filenames=None):
+        self._reset()
```
PeterSurda, line 101: debug backtrace
debug object address

PeterSurda: maybe also on `get` and `set`

src/bitmessageqt/settings.py
```diff
@@ -16,15 +16,15 @@
...
+from bmconfigparser import config
 from helper_sql import sqlExecute, sqlStoredProcedure
 from helper_startup import start_proxyconfig
 from network import knownnodes, AnnounceThread
 from network.asyncore_pollchoose import set_rates
 from tr import _translate
 
 
-def getSOCKSProxyType(config):
+def getSOCKSProxyType(config):  # pylint: disable=redefined-outer-name
```
PeterSurda, line 27: rename varaible (e.g. `config_`)

src/default.ini
```diff
@@ -0,0 +1,47 @@
...
+socksport = 9050
+socksauthentication = False
+socksusername = None
+sockspassword = None
+keysencrypted = False
+messagesencrypted = False
+minimizeonclose = False
+dontconnect = True
+replybelow = False
+stopresendingafterxdays = None
```
PeterSurda, line 28: these 2 should be empty string

src/default.ini
```diff
@@ -0,0 +1,47 @@
...
+timeformat = %%c
+blackwhitelist = black
+startonlogon = False
+showtraynotifications = True
+startintray = False
+socksproxytype = none
+sockshostname = localhost
+socksport = 9050
+socksauthentication = False
+socksusername = None
```
PeterSurda, line 21: shouldn't be `None`, but empty string

src/default.ini
```diff
@@ -0,0 +1,47 @@
...
+sockspassword = None
```
PeterSurda, line 22: also empty string

src/default.ini
```diff
@@ -0,0 +1,47 @@
...
+keysencrypted = False
+messagesencrypted = False
+minimizeonclose = False
+dontconnect = True
+replybelow = False
+stopresendingafterxdays = None
+stopresendingafterxmonths = None
+opencl = None
```
PeterSurda, line 30: for consistency purposes maybe also empty string

src/tests/test_config.py
```diff
@@ -42,56 +42,69 @@ class TestConfig(unittest.TestCase):
...
+            config.safeGetFloat('nonexistent', 'nonexistent', 42.0), 42.0)
 
     def test_reset(self):
         """safeGetInt retuns provided default for bitmessagesettings option or 0"""
+        config = BMConfigParser()
         test_config_object = StringIO(test_config)
-        BMConfigParser().readfp(test_config_object)
+        config.readfp(test_config_object)
+        # pylint: disable=protected-access
+        config._reset()
```
PeterSurda, line 99: after reset, needs to test that what we deleted is absent

src/tests/test_config.py
```diff
@@ -42,56 +42,69 @@ class TestConfig(unittest.TestCase):
...
 
+    def test_defaults(self):
+        """Loading defaults"""
+        config = BMConfigParser()
+        config.add_section('bitmessagesettings')
+        config.set("bitmessagesettings", "maxaddrperstreamsend", "100")
         self.assertEqual(
```
PeterSurda, line 106: probably don't need this one assert

PeterSurda (changes requested):

packages/pyinstaller/bitmessagemain.spec
```diff
@@ -94,7 +94,8 @@ a.binaries += [
     (os.path.join('sslkeys', 'cert.pem'),
          os.path.join(srcPath, 'sslkeys', 'cert.pem'), 'BINARY'),
     (os.path.join('sslkeys', 'key.pem'),
-         os.path.join(srcPath, 'sslkeys', 'key.pem'), 'BINARY')
+         os.path.join(srcPath, 'sslkeys', 'key.pem'), 'BINARY'),
+    ('default.ini', os.path.join(srcPath, 'default.ini'), 'DATA'),
```
PeterSurda, line 98: - should go to `datas`, not `binaries`
- source path missing

src/tests/test_config.py
```diff
@@ -42,56 +42,70 @@ class TestConfig(unittest.TestCase):
 
     def setUp(self):
         """creates a backup of BMConfigparser current state"""
-        BMConfigParser().write(self.configfile)
+        config = BMConfigParser()
```
PeterSurda, line 45: delete `setUp` and `tearDown`

packages/pyinstaller/bitmessagemain.spec
```diff
@@ -94,7 +94,7 @@ a.binaries += [
     (os.path.join('sslkeys', 'cert.pem'),
          os.path.join(srcPath, 'sslkeys', 'cert.pem'), 'BINARY'),
     (os.path.join('sslkeys', 'key.pem'),
-         os.path.join(srcPath, 'sslkeys', 'key.pem'), 'BINARY')
+         os.path.join(srcPath, 'sslkeys', 'key.pem'), 'BINARY'),
```
PeterSurda, line 97: remove `,`

PeterSurda (changes requested):

src/bmconfigparser.py
PeterSurda: So you basically recommend dependency injection. Which is ok, I'm just not sure that can be done easily now because of legacy structures. Using a module variable is halfway there, we can do unit tests on isolated objects, and can be refactored into dependency injection later.

PeterSurda (approved):

PeterSurda:
@g1itch I'm ready to merge this, any objections other than dependency injection? That can be done later and easier now with this refactoring.

src/bmconfigparser.py
PeterSurda: It's not just tests though, there are some additions that prevent misbehaviour, like a clean reset, consolidation of default values and threads waiting for initialization. I'm not super happy that the commit is kind of big, just the old code is difficult to modify in small chunks. The commit makes the code easier to change in the future.

PeterSurda:
> No. Maybe I should accept the death of #1389.

There's work being done on the mock backend and more flexible buildbot jobs. Once they are available, UI tests can be executed not only for kivy but also for Qt. Then #1389 can be tested more reliably and merged.

g1itch:
It looks like `SettingsDialog.net_restart_needed`  in the `bitmessageqt` is ignored in the appimage built with this changeset.

PR #1928, opened 2022-02-16
Fix BMConfigParser.setTemp() broken in a5773999

g1itch:
Hi!

I reverted some changes in `bmconfigparser` from previous commits. Particularly those, breaking the logic of `setTemp()`, which is used in proxyconfig plugin and for network restart after changing the proxy type.

Done some formatting for PEP8. In this regard, the question arose: what is hidden addresses and why it's only for python3?


g1itch:
Another interesting failure in buildbot: https://buildbot.bitmessage.org/#/builders/25/builds/3740

PeterSurda (approved):

g1itch:
What about hidden addresses? I want to rewrite this part as well: https://github.com/Bitmessage/PyBitmessage/blob/v0.6/src/bmconfigparser.py#L106

PeterSurda:
`hidden` doesn't appear to be used anywhere and probably won't, so can be deleted.

PeterSurda:
Do you need anything else from me before merging this?

g1itch:
Sorry, I got distracted a bit. Is this squashed enough?

PeterSurda (approved):

PR #1931, opened 2022-02-18
Add main.kv in kivy live

shekhar-cis:
Added main.kv: inside bitmessagekivy/main.kv,
Symlinked with kivy mock.


PeterSurda (changes requested):
- remove the commented code
- sort imports

PeterSurda (approved):

PR #1934, opened 2022-02-23
Kivy Addressbook screen

shekhar-cis:
Added `addressbook.py` kivy screen inside `bitmessagekivy/baseclass`


PeterSurda (changes requested):
can we split longer methods to smaller ones?

PR #1954, opened 2022-03-29
Kivy_state.py

shekhar-cis:
Kivy related variables are assigned in kivy_state.py, separated from state.py.

### DIFF
--- a/src/bitmessagekivy/kivy_state.py
+++ b/src/bitmessagekivy/kivy_state.py
@@ -1,0 +1,36 @@
+# pylint: disable=too-many-instance-attributes, too-few-public-methods
+
+"""
+Kivy State variables are assigned here, they are separated from state.py
+=================================
+"""
+
+
+class KivyStateVariables(object):
+    """This Class hold all the kivy state variables"""
+
+    def __init__(self):
+        self.association = ''
+        self.navinstance = None
+        self.mail_id = 0
+        self.myAddressObj = None
+        self.detailPageType = None
+        self.ackdata = None
+        self.status = None
+        self.screen_density = None
+        self.msg_counter_objs = None
+        self.check_sent_acc = None
+        self.sent_count = 0
+        self.inbox_count = 0
+        self.trash_count = 0
+        self.draft_count = 0
+        self.all_count = 0
+        self.searcing_text = ''
+        self.search_screen = ''
+        self.send_draft_mail = None
+        self.is_allmail = False
+        self.in_composer = False
+        self.availabe_credit = 0
+        self.in_sent_method = False
+        self.in_search_mode = False
+        self.imageDir = None
--- a/src/bitmessagekivy/mpybit.py
+++ b/src/bitmessagekivy/mpybit.py
@@ -6,9 +6,12 @@
 from kivy.app import App
 from kivy.uix.label import Label
 
+from pybitmessage.bitmessagekivy.kivy_state import KivyStateVariables
+
 
 class NavigateApp(App):
     """Navigation Layout of class"""
+    kivy_state_obj = KivyStateVariables()
 
     def build(self):
         """Method builds the widget"""


PeterSurda (changes requested):
Make it into a class. Create an instance of this class at the beginning (e.g. in `main.py` or something like that, and then pass it into the other objects when you create them (dependency injection).

src/kivy_state.py
```diff
@@ -0,0 +1,72 @@
...
+# Network protocols availability, initialised below
+networkProtocolAvailability = None
+
+sqlReady = False  # set to true by sqlTread when ready for processing
+
+trustedPeer = None
+Peer = collections.namedtuple('Peer', ['host', 'port'])
+
+
+def resetNetworkProtocolAvailability():
```
PeterSurda, line 17: this should be only a pure data class, so this function should go elsewhere

src/kivy_state.py
```diff
@@ -0,0 +1,72 @@
...
+    """This method helps to reset the availability of network protocol"""
+    # pylint: disable=global-statement
+    global networkProtocolAvailability
```
PeterSurda, line 20: if we need to modify this, maybe make `networkProtocolAvailability` into an object? And make `reset` a method of this object.

PeterSurda (changes requested):
In first step, make this into a class. Then in later iterations maybe we can move some of these variables into better classes, to be attributes of individual UI elements or screens or something.

src/kivy_state.py
```diff
@@ -0,0 +1,50 @@
+"""
+Kivy_state.py
+=================================
+"""
+import collections
+
+
+# Network protocols availability, initialised below
+networkProtocolAvailability = None
```
PeterSurda, line 9: make this into a class and instantiate one inside the `__init__` of `KivyStateVariables`

src/kivy_state.py
```diff
@@ -0,0 +1,50 @@
...
+
+def resetNetworkProtocolAvailability():
```
PeterSurda, line 11: this should be a method of the class mentioned above

PeterSurda (changes requested):

src/kivy_state.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+import collections
+
+
+# pylint: disable=old-style-class
+class KivyStateVariables:
+    """This Class hold all the kivy state variables"""
+
+    def __init__(self):
+        self.sqlReady = False  # set to true by sqlTread when ready for processing
```
PeterSurda, line 15: remove

src/kivy_state.py
```diff
@@ -0,0 +1,50 @@
...
+        self.trustedPeer = None
```
PeterSurda, line 16: remove

src/kivy_state.py
```diff
@@ -0,0 +1,50 @@
...
+        self.Peer = collections.namedtuple('Peer', ['host', 'port'])
```
PeterSurda, line 17: remove

src/kivy_state.py
```diff
@@ -0,0 +1,50 @@
...
+        self.searcing_text = ''
+        self.search_screen = ''
+        self.send_draft_mail = None
+        self.is_allmail = False
+        self.in_composer = False
+        self.availabe_credit = 0
+        self.in_sent_method = False
+        self.in_search_mode = False
+        self.imageDir = None
+        self.networkProtocolAvailability = None  # Network protocols availability
```
PeterSurda, line 42: remove

src/kivy_state.py
```diff
@@ -0,0 +1,50 @@
...
+        self.resetNetworkProtocolAvailability()
```
PeterSurda, line 43: remove

src/kivy_state.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def resetNetworkProtocolAvailability(self):
```
PeterSurda, line 45: remove

src/kivy_state.py
```diff
@@ -0,0 +1,50 @@
...
+        """This method helps to reset the availability of network protocol"""
+        self.networkProtocolAvailability = {'IPv4': None, 'IPv6': None, 'onion': None}
+
+
+kivy_state_obj = KivyStateVariables()
```
PeterSurda, line 50: instantiation should be in main.py or basically somewhere very early during runtime

PeterSurda (changes requested):

src/kivy_state.py
```diff
@@ -0,0 +1,37 @@
+# pylint: disable=too-many-instance-attributes, too-few-public-methods, old-style-class
+
+"""
+Kivy_state.py
```
PeterSurda, line 4: better docstring

src/kivy_state.py
```diff
@@ -0,0 +1,37 @@
...
+=================================
+"""
+
+
+# pylint: disable=old-style-class
```
PeterSurda, line 9: remove

src/kivy_state.py
```diff
@@ -0,0 +1,37 @@
+# pylint: disable=too-many-instance-attributes, too-few-public-methods, old-style-class
```
PeterSurda, line 1: remove `old-style-class`

src/kivy_state.py
```diff
@@ -0,0 +1,37 @@
...
+
+"""
+Kivy_state.py
+=================================
+"""
+
+
+# pylint: disable=old-style-class
+class KivyStateVariables:
```
PeterSurda, line 10: make this into a new style class, this is not a mixin

PeterSurda (changes requested):

PeterSurda:
Inside main or something like that, very early, create one instance of this class and save it inside the app.

PeterSurda (approved):

PR #1967, opened 2022-05-04
Add address enable/disable function

shekhar-cis:
I have added function to `enable and disable` identity in `src/addresses.py`.

### DIFF
--- a/src/addresses.py
+++ b/src/addresses.py
@@ -7,6 +7,7 @@ import logging
 from binascii import hexlify, unhexlify
 from struct import pack, unpack
 
+from bmconfigparser import BMConfigParser
 
 logger = logging.getLogger('default')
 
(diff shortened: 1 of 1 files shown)


src/addresses.py
```diff
@@ -281,3 +282,15 @@ def addBMIfNotPresent(address):
     """Prepend BM- to an address if it doesn't already have it"""
     address = str(address).strip()
     return address if address[:3] == 'BM-' else 'BM-' + address
+
+
+def disable_addresses(address):
```
PeterSurda, line 287: should be signular as it only disables one address

src/addresses.py
```diff
@@ -281,3 +282,15 @@ def addBMIfNotPresent(address):
...
+    """"To disable the Address"""
+    BMConfigParser().set(str(address), 'enabled', 'false')
+    BMConfigParser().save()
+
+
+def enable_addresses(address):
```
PeterSurda, line 293: here too

src/addresses.py
```diff
@@ -7,6 +7,7 @@
 from binascii import hexlify, unhexlify
 from struct import pack, unpack
 
+from bmconfigparser import BMConfigParser
```
PeterSurda, line 10: since this adds an import, it may be better to put the functions into `BMConfigParser` directly and make them methods

PeterSurda (changes requested):

src/addresses.py
shekhar-cis: Alright, i'll create a new PR to do update in `BMConfigparser`.

PR #1973, opened 2022-05-13
Test: enable/disable the identity

shekhar-cis:
Added test for enabling and disabling the address.

### DIFF
--- a/src/tests/test_config.py
+++ b/src/tests/test_config.py
@@ -1,8 +1,9 @@
 """
 Various tests for config
 """
+import logging
 import unittest
-
+from six.moves import configparser
 from six import StringIO
 from pybitmessage.bmconfigparser import BMConfigParser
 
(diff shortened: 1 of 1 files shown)


src/tests/test_config.py
```diff
@@ -32,7 +33,17 @@
 maxnodes = 15000
 
 [zlib]
-maxsize = 1048576"""
+maxsize = 1048576
+
+[BM-2cUkVw5kgWfNjydbCHW7sewmG6fbkTxcih]
```
PeterSurda, line 38: put this section into a separate constant

src/tests/test_config.py
```diff
@@ -104,3 +115,31 @@ def test_defaults(self):
         self.assertEqual(
             self.config.safeGetInt(
                 'bitmessagesettings', 'maxaddrperstreamsend'), 500)
+
+    def test_disable_identity(self):
+        """Test Disable the identity"""
+        test_config_object = StringIO(test_config)
```
PeterSurda, line 121: here just load the new constant

src/tests/test_config.py
PeterSurda: and make a separate constant which contains a disabled address

src/tests/test_config.py
```diff
@@ -104,3 +115,31 @@ def test_defaults(self):
         self.assertEqual(
             self.config.safeGetInt(
                 'bitmessagesettings', 'maxaddrperstreamsend'), 500)
+
+    def test_disable_identity(self):
```
PeterSurda, line 119: both two new methods should do 3 things:
- change
- change already changed
- changed non-existing

No need for `try`/`except`. There is `AssertRaises` for that. No need for `if` either.

src/tests/test_config.py
PeterSurda, line 121: maybe write a decorator (or maybe a new class) to do the init, and then write 6 methods

```
def setUp():
    self.enabled = Config()
    self.enabled.load(enabled_constant)
    self.disabled = Config()
    self.disabled.load(disabled_constant)
    self.nonexistent = Config()
```

PeterSurda: individual unit tests would be something like this:

```
self.enabled.enable(aldksfjal)
assertEquals(self.enabled.get(blabla), True)
```
```
assertRaises(self.nonexistent.enable(blablbla))
```

PeterSurda (changes requested):

PR #1977, opened 2022-05-30
Add kivy QR-code Screen

shekhar-cis:
Hi,
    I have added a QR-code screen to generate qrcode for the saved addresses.

### DIFF
--- a/src/bitmessagekivy/baseclass/qrcode.py
+++ b/src/bitmessagekivy/baseclass/qrcode.py
@@ -1,0 +1,31 @@
+# pylint: disable=import-error, no-name-in-module, too-few-public-methods
+
+"""
+Generate QRcode of saved addresses in addressbook.
+"""
+
+from kivy.app import App
+from kivy.uix.screenmanager import Screen
+from kivy.properties import StringProperty
+from kivy_garden.qrcode import QRCodeWidget
+from debug import logger
+
+
+class ShowQRCode(Screen):
+    """ShowQRCode Screen class for kivy Ui"""
+    address = StringProperty()
+
+    def __init__(self, *args, **kwargs):
+        """Instantiate kivy state variable"""
+        super(ShowQRCode, self).__init__(*args, **kwargs)
+        self.kivy_running_app = App.get_running_app()
+
+    def qrdisplay(self, instance, address):
+        """Method used for showing QR Code"""
+        self.ids.qr.clear_widgets()
+        self.kivy_running_app.set_toolbar_for_QrCode()
+        self.address = address  # used for label
+        self.ids.qr.add_widget(QRCodeWidget(data=self.address))
+        self.ids.qr.children[0].show_border = False
+        instance.parent.parent.parent.dismiss()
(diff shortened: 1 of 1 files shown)


src/bitmessagekivy/baseclass/qrcode.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def qrdisplay(self, instasnce, address):
+        """Method used for showing QR Code"""
+        self.ids.qr.clear_widgets()
+        state.kivyapp.set_toolbar_for_QrCode()
+        self.address = address
+        self.ids.qr.add_widget(QRCodeWidget(data=address))
+        self.ids.qr.children[0].show_border = False
+        instasnce.parent.parent.parent.dismiss()
+        toast('Show QR code')
```
PeterSurda, line 26: remove `toast`, or change it to `log.debug` or something like that

src/bitmessagekivy/baseclass/qrcode.py
```diff
@@ -0,0 +1,26 @@
...
+
+class ShowQRCode(Screen):
+    """ShowQRCode Screen class for kivy Ui"""
+    address = StringProperty()
+
+    def qrdisplay(self, instasnce, address):
+        """Method used for showing QR Code"""
+        self.ids.qr.clear_widgets()
+        state.kivyapp.set_toolbar_for_QrCode()
+        self.address = address
```
PeterSurda, line 22: why are we using `self.address`?

PeterSurda (changes requested):

src/bitmessagekivy/baseclass/qrcode.py
shekhar-cis: Yes we can remove it, it is not using here.

src/bitmessagekivy/baseclass/qrcode.py
```diff
@@ -0,0 +1,26 @@
...
+        self.ids.qr.add_widget(QRCodeWidget(data=address))
```
PeterSurda, line 23: ```self.ids.qr.add_widget(QRCodeWidget(data=self.address))```

otherwise it's a bit confusing as to why we're using both `address` and `self.address`

src/bitmessagekivy/baseclass/qrcode.py
```diff
@@ -0,0 +1,26 @@
...
+
+class ShowQRCode(Screen):
+    """ShowQRCode Screen class for kivy Ui"""
+    address = StringProperty()
+
+    def qrdisplay(self, instasnce, address):
+        """Method used for showing QR Code"""
+        self.ids.qr.clear_widgets()
+        state.kivyapp.set_toolbar_for_QrCode()
+        self.address = address
```
PeterSurda, line 22: I would add a comment here, e.g. `#  used for label`

PeterSurda (changes requested):

PeterSurda (changes requested):
typo

PeterSurda (approved):

PR #1997, opened 2022-07-11
Fix telenium_process

shekhar-cis:


PeterSurda (changes requested):
- unstage addressbook file
- fix the assert as it is
- name of the PR is misleading

PeterSurda (approved):

PR #1999, opened 2022-07-13
Refactor sqlthread and add create test cases

shekhar-cis:


src/class_sqlThread.py
```diff
@@ -17,128 +17,100 @@
...
+        self.cur.execute(query, parameters)
+        return int(self.cur.fetchall()[0][0])
+
+    def _upgrade_one_level_sql_statement(self, file_name):
+        """
+            Upgrade database versions with applying sql scripts
+        """
+        self.initialise_sql("init_version_{}".format(file_name))
+
+    def initialise_sql(self, file_name):
```
PeterSurda, line 71: the PR should use consistent language dialect. In some places in this PR verbs end with `-se` (UK) and other places with `-ze` (US). I don't think the project has a standard for this but I think it should be `-ze`.

src/class_sqlThread.py
```diff
@@ -487,154 +206,258 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+            upgrade config parser setting version.
+            If the settings version is equal to 2 or 3 then the
+            sqlThread will modify the pubkeys table and change
+            the settings version to 4.
+        """
+        while self.sql_config_settings_version < self.max_setting_level:
+            self.db.upgrade_config_parser_setting_version(self.sql_config_settings_version)
+            self.sql_config_settings_version = self.sql_config_settings_version
+
+    def loop_queue(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements
```
PeterSurda, line 284: this should be split into multiple methods

src/class_sqlThread.py
```diff
@@ -17,128 +17,100 @@
...
+        """
+        return self.__get_current_settings_version()
+
+    @sql_schema_version.setter
+    def sql_schema_version(self, setter):
+        """
+            Update version with one level
+        """
+        if setter:
+            query = "UPDATE settings SET value=CAST(value AS INT) + 1 WHERE key = 'version'"
```
PeterSurda, line 96: we should have a test of what happens if there is legacy string in the database when retrieving the value

src/class_sqlThread.py
```diff
@@ -17,128 +17,100 @@
...
-                int(time.time()),))
-            self.cur.execute(
-                '''CREATE TABLE objectprocessorqueue'''
-                ''' (objecttype int, data blob, UNIQUE(objecttype, data) ON CONFLICT REPLACE)''')
-            self.conn.commit()
-            logger.info('Created messages database file')
+            with open(os.path.join(self.root_path, "pybitmessage/sql/{}.sql".format(file_name))) as sql_file:
+                sql_as_string = sql_file.read()
+            self.cur.executescript(sql_as_string)
+        except IOError as err:
```
PeterSurda, line 76: It may be helpful to have a common error handler or at least common error messages for errors. Or, perhaps a decorator would be helpful? It would wrap the whole method in a `try`/`except`, assuming we can refactor the rest of the code so that all the SQL operations can be done by dedicated methods.

PeterSurda (changes requested):

src/tests/test_sqlthread.py
```diff
@@ -2,43 +2,223 @@
...
-        query = sqlQuery('SELECT enaddr(4, 1, "21122112211221122112")')
-        self.assertEqual(
-            query[0][-1], encoded_str, "test case fail for create_function")
+    @version
+    def test_bm_db_version_type(self):
+        """
+            Test version type
+        """
+        self.test_db.cur.execute('''INSERT INTO settings VALUES('version','test_string')''')  # noqa
+        version = self.test_db.cur.execute("SELECT * FROM settings WHERE key='version'")
```
PeterSurda, line 222: between 221 and 222 insert here a call to `sql_schema_version`

src/class_sqlThread.py
```diff
@@ -17,128 +19,104 @@
...
-        # From now on, let us keep a 'version' embedded in the messages.dat
-        # file so that when we make changes to the database, the database
-        # version we are on can stay embedded in the messages.dat file. Let us
-        # check to see if the settings table exists yet.
-        item = '''SELECT name FROM sqlite_master WHERE type='table' AND name='settings';'''
-        parameters = ''
-        self.cur.execute(item, parameters)
+            logger.debug(
+                'ERROR trying to initialize database. Error message: %s\n', str(err))
+
```
PeterSurda, line 87: We should also handle `FileNotFoundError`, then log debug that the file is missing, and return.

src/class_sqlThread.py
```diff
@@ -17,128 +19,104 @@
...
+            self.cur.execute(query)
+
+    def upgrade_to_latest(self):
+        """
+            Initialise upgrade level
+        """
+
+        while self.sql_schema_version < self.max_level:
+            self._upgrade_one_level_sql_statement(self.sql_schema_version)
+            self.sql_schema_version = True
```
PeterSurda, line 112: add comment `# bump sql_schema_version by one`

PeterSurda (changes requested):

src/class_sqlThread.py
shekhar-cis: I have checked that `FileNotFoundError` is for `python3`.
or we can do like
`if six.PY2: 
    FileNotFoundError = IOError`

or
`try:
    FileNotFoundError
except NameError:
    # py2
    FileNotFoundError = IOError`

And `IOError` raises `IOError(2, 'No such file or directory')`

src/class_sqlThread.py
```diff
@@ -17,128 +19,107 @@
...
-        # From now on, let us keep a 'version' embedded in the messages.dat
-        # file so that when we make changes to the database, the database
-        # version we are on can stay embedded in the messages.dat file. Let us
-        # check to see if the settings table exists yet.
-        item = '''SELECT name FROM sqlite_master WHERE type='table' AND name='settings';'''
-        parameters = ''
-        self.cur.execute(item, parameters)
+            logger.debug(
+                'ERROR trying to initialize database. Error message: %s\n', str(err))
+
```
PeterSurda, line 90: we seem to be missing a sqlite error handler. This should also log, but stronger severity (e.g. `error`), not `debug`

src/class_sqlThread.py
```diff
@@ -17,128 +19,107 @@
...
-        self.create_function()
-
+    def __get_current_settings_version(self):
+        """
+            Get current setting Version
+        """
+        query = "SELECT value FROM settings WHERE key='version'"
+        parameters = ()
+        self.cur.execute(query, parameters)
+        return int(self.cur.fetchall()[0][0])
```
PeterSurda, line 65:
```
except ValueError:
    return 0
```

src/class_sqlThread.py
```diff
@@ -487,154 +214,237 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+            If the settings version is equal to 2 or 3 then the
+            sqlThread will modify the pubkeys table and change
+            the settings version to 4.
+        """
+        while self.sql_config_settings_version < self.max_setting_level:
+            self.db.upgrade_config_parser_setting_version(self.sql_config_settings_version)
+            self.sql_config_settings_version = self.sql_config_settings_version
+
+    @staticmethod
+    def error_handler(err, custom_error):
```
PeterSurda, line 294: instead of `custom_error` maybe `command`

PeterSurda (changes requested):

src/class_sqlThread.py
```diff
@@ -17,128 +19,112 @@
...
-                '''CREATE TABLE settings (key blob, value blob, UNIQUE(key) ON CONFLICT REPLACE)''')
-            self.cur.execute('''INSERT INTO settings VALUES('version','11')''')
-            self.cur.execute('''INSERT INTO settings VALUES('lastvacuumtime',?)''', (
-                int(time.time()),))
-            self.cur.execute(
-                '''CREATE TABLE objectprocessorqueue'''
-                ''' (objecttype int, data blob, UNIQUE(objecttype, data) ON CONFLICT REPLACE)''')
-            self.conn.commit()
-            logger.info('Created messages database file')
+            self.cur.execute(query, parameters)
```
PeterSurda, line 65: only `return` should be wrapped under `try`

src/class_sqlThread.py
```diff
@@ -487,154 +219,239 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+        elif query == 'exit':
+            self.db.conn.close()
+            logger.info('sqlThread exiting gracefully.')
+            return False
+        elif query == 'movemessagstoprog':
+            logger.debug('the sqlThread is moving the messages.dat file to the local program directory.')
+            self.is_query_movemessagstoprog()
+        elif query == 'movemessagstoappdata':
+            logger.debug('the sqlThread is moving the messages.dat file to the Appdata folder.')
+            try:
```
PeterSurda, line 397: these sections should also be in a separate function/method

src/class_sqlThread.py
```diff
@@ -487,154 +219,239 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+        self.db.cur = self.db.conn.cursor()
+
+    def is_query_deleteandvacuume(self):
+        """
+            When query == 'deleteandvacuume'
+        """
+        try:
+            self.db.cur.execute(''' VACUUM ''')
+        except Exception as err:
+            command = 'deleteandvacuume'
```
PeterSurda, line 348: I don't know if there needs to be a variable `command, why not just use the string directly?

PeterSurda (changes requested):

src/class_sqlThread.py
```diff
@@ -17,128 +19,112 @@
...
+        """
+            Initializing sql
+        """
+        try:
+            with open(os.path.join(self.root_path, "pybitmessage/sql/{}.sql".format(file_name))) as sql_file:
+                sql_as_string = sql_file.read()
+            self.cur.executescript(sql_as_string)
+        except OSError as err:
+            logger.debug('The file is missing. Error message: %s\n', str(err))
+            return
```
PeterSurda, line 86: get rid of this `return`

PeterSurda (changes requested):

src/class_sqlThread.py
PeterSurda: ok I didn't explain it well:
```
self.cur.execute(query, parameters)
try:
    return int(self.cur.fetchall()[0][0])
except ValueError, IndexError
    return 0
```

I added an IndexError just in case.

src/class_sqlThread.py
```diff
@@ -17,128 +19,112 @@
...
-        # From now on, let us keep a 'version' embedded in the messages.dat
-        # file so that when we make changes to the database, the database
-        # version we are on can stay embedded in the messages.dat file. Let us
-        # check to see if the settings table exists yet.
-        item = '''SELECT name FROM sqlite_master WHERE type='table' AND name='settings';'''
-        parameters = ''
-        self.cur.execute(item, parameters)
+            logger.debug(
+                'ERROR trying to initialize database. Error message: %s\n', str(err))
+
```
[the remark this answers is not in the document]
PeterSurda: `return False` should be after the `try/except` block so that it's consistent (returns `True` if no exception, otherwise returns `False`)

PeterSurda (changes requested):
The two changes I suggested should fix the tests too.

src/class_sqlThread.py
```diff
@@ -473,168 +194,249 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
+        """
+            When the query can be default or other '
+        """
+        parameters = helper_sql.sqlSubmitQueue.get()
+        try:
+            self.db.cur.execute(query, parameters)
+            self.rowcount = self.db.cur.rowcount
+            return self.rowcount
+        except Exception as err:
+            if str(err) != 'database or disk is full':
```
PeterSurda, line 341: replace with `error_handler`

PeterSurda (changes requested):

src/class_sqlThread.py
shekhar-cis: `error_handler` is in `else` block.

PeterSurda: the error handling should be unified, this here isn't a special case

shekhar-cis: Added inside `error_handler`.

src/class_sqlThread.py
```diff
@@ -443,28 +170,22 @@ def run(self):  # pylint: disable=too-many-locals, too-many-branches, too-many-s
...
-                        _translate(
-                            "MainWindow",
-                            "Disk full"),
-                        _translate(
-                            "MainWindow",
-                            'Alert: Your disk or data storage volume is full. Bitmessage will now exit.'),
-                        True)))
-                os._exit(0)
-            else:
+            if str(err) != 'database or disk is full':
```
PeterSurda, line 173: this should also use `error_handler`

PeterSurda (changes requested):

PR #2023, opened 2022-08-08
Update kivy_state

shekhar-cis:
Add Default image path to `kivy_state`

### DIFF
--- a/src/bitmessagekivy/kivy_state.py
+++ b/src/bitmessagekivy/kivy_state.py
@@ -5,6 +5,9 @@ Kivy State variables are assigned here, they are separated from state.py
 =================================
 """
 
+import os
+import threading
+
 
 class KivyStateVariables(object):
     """This Class hold all the kivy state variables"""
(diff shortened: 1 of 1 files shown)


src/bitmessagekivy/kivy_state.py
```diff
@@ -33,4 +35,4 @@ def __init__(self):
         self.available_credit = 0
         self.in_sent_method = False
         self.in_search_mode = False
-        self.image_dir = None
+        self.image_dir = os.path.join('./images', 'kivy')
```
PeterSurda, line 38: how about just
`os.path.join('images', 'kivy')`

PeterSurda (changes requested):

src/bitmessagekivy/kivy_state.py
```diff
@@ -33,4 +36,5 @@ def __init__(self):
         self.available_credit = 0
         self.in_sent_method = False
         self.in_search_mode = False
-        self.image_dir = None
+        self.image_dir = os.path.abspath(os.path.join('images', 'kivy'))
+        self.sql_ready = threading.Event()
```
PeterSurda, line 40: different name as mentioned in another PR.

PeterSurda (changes requested):

src/bitmessagekivy/kivy_state.py
PeterSurda: looks ok now

PR #2024, opened 2022-08-09
Add kivy uisignaler

shekhar-cis:


src/bitmessagekivy/uikivysignaler.py
```diff
@@ -0,0 +1,34 @@
...
+
+class UIkivySignaler(Thread):
+    """Kivy ui signaler"""
+
+    def __init__(self, *args, **kwargs):
+        super(UIkivySignaler, self).__init__(*args, **kwargs)
+        self.kivy_state = kivy_state_variables()
+
+    def run(self):
+        self.kivy_state.sql_ready.wait()
```
PeterSurda, line 22: don't use `sql_ready`, create a new `threading.Event` object, just use a different name. Just look at how `sql_ready` is used so that you understand what to do.

src/bitmessagekivy/uikivysignaler.py
```diff
@@ -0,0 +1,34 @@
...
+        while state.shutdown == 0:
+            try:
+                command, data = queues.UISignalQueue.get()
+                if command == 'writeNewAddressToTable':
+                    address = data[1]
+                    App.get_running_app().variable_1.append(address)
```
PeterSurda, line 28: rename `variable_1` to something more helpful

PeterSurda (changes requested):

PR #2029, opened 2022-08-10
Update screen id name for generating new and random address

shekhar-cis:
Update id of random screen; which generate the random identity.

### DIFF
--- a/src/bitmessagekivy/baseclass/maildetail.py
+++ b/src/bitmessagekivy/baseclass/maildetail.py
@@ -135,7 +135,7 @@ class MailDetail(Screen):  # pylint: disable=too-many-instance-attributes
             self.status = data[0][4]
         self.time_tag = ShowTimeHistoy(data[0][4]) if self.kivy_state.detailPageType == 'inbox' \
             else ShowTimeHistoy(data[0][6])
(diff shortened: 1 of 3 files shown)


PeterSurda:
This PR is ok but too small. I'd prefer multiple similar changes in one PR.

shekhar-cis:
> This PR is ok but too small. I'd prefer multiple similar changes in one PR.

ok

src/bitmessagekivy/baseclass/maildetail.py
```diff
@@ -135,7 +135,7 @@ def assign_mail_details(self, data):
             self.status = data[0][4]
         self.time_tag = ShowTimeHistoy(data[0][4]) if self.kivy_state.detailPageType == 'inbox' \
             else ShowTimeHistoy(data[0][6])
-        self.avatarImg = os.path.join(self.kivy_state.imageDir, 'avatar.png') \
```
PeterSurda, line 138: just don't use "avatar", this isn't an avatar.

PeterSurda (changes requested):
normally, the "draft" change would be one PR (or at least one commit), and the id/sc changes would be another one, but for all files together.

PeterSurda (approved):

PR #2046, opened 2022-08-26
Update kivy imports

shekhar-cis:
Add `logger` and `import logging`

### DIFF
--- a/src/bitmessagekivy/baseclass/login.py
+++ b/src/bitmessagekivy/baseclass/login.py
@@ -12,10 +12,11 @@ from kivy.uix.boxlayout import BoxLayout
 from kivy.uix.screenmanager import Screen
 from kivy.app import App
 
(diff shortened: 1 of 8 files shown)


src/bitmessagekivy/baseclass/login.py
```diff
@@ -12,10 +12,11 @@
 from kivy.uix.screenmanager import Screen
 from kivy.app import App
 
-from backend.address_generator import AddressGenerator  # pylint: disable=import-error
 from kivymd.uix.behaviors.elevation import RectangularElevationBehavior
+
+from pybitmessage.backend.address_generator import AddressGenerator  # pylint: disable=import-error
```
PeterSurda, line 17: do we need disable import error?

PeterSurda (changes requested):

PeterSurda (approved):

PR #2052, opened 2022-08-30
Add Kivy mock network

shekhar-cis:
Add network for kivy mock

### DIFF
--- a/src/mock/network.py
+++ b/src/mock/network.py
@@ -1,0 +1,18 @@
+# pylint: disable=too-few-public-methods
+
+"""
+    Mock Network
+"""
+
+
+class objectracker(object):
+    """Mock object tracker"""
+    missingObjects = {}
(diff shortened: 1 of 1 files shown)


src/mock/network.py
```diff
@@ -0,0 +1,18 @@
+# pylint: disable=too-few-public-methods
+
+"""
+    Mock Network
+"""
+
+
+class objectracker(object):
+    """Mock object tracker"""
+    missingObjects = ()
```
PeterSurda, line 10: should be an empty dict, not an empty list

PeterSurda (changes requested):

src/mock/network.py
shekhar-cis: Done

PeterSurda (approved):

PR #2053, opened 2022-08-30
Import network for mock 

shekhar-cis:
Import network for mock


src/mock/kivy_main.py
```diff
@@ -7,6 +7,7 @@
 
 import multiqueue
 from class_addressGenerator import FakeAddressGenerator
+from network import stats, objectracker
```
PeterSurda, line 10: maybe this needs to be
```from pybitmesasge.mock.network import stats, objectracker```
? Not sure.

shekhar-cis: Done.

PeterSurda (approved):

PeterSurda (approved):

PR #2093, opened 2022-10-20
Add python3 compatibility variable in state.py and update network.py

osamacis:
Add python3 compatibility variable in `state.py `and update `network.py`

### DIFF
--- a/src/bitmessagekivy/baseclass/network.py
+++ b/src/bitmessagekivy/baseclass/network.py
@@ -13,7 +13,7 @@ from kivy.uix.screenmanager import Screen
 
 from pybitmessage import state
 
-if os.environ.get('INSTALL_TESTS', False):
+if os.environ.get('INSTALL_TESTS', False) and not state.backend_py3_compatible:
     from pybitmessage.mock import kivy_main
     stats = kivy_main.network.stats
     objectracker = kivy_main.network.objectracker
--- a/src/state.py
+++ b/src/state.py
@@ -69,3 +69,5 @@ ackdataForWhichImWatching = {}
 
 thisapp = None
 """Singleton instance"""
+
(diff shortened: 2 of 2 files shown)


src/bitmessagekivy/baseclass/network.py
```diff
@@ -17,7 +17,8 @@
     from pybitmessage.mock import kivy_main
     stats = kivy_main.network.stats
     objectracker = kivy_main.network.objectracker
-else:
+
+if state.py3_compatible:
```
PeterSurda, line 21: maybe `state.backend_py3_compatible`

PeterSurda (changes requested):

src/bitmessagekivy/baseclass/network.py
osamacis: yes now i have forcefully pushed the changed and now it is same as suggested by you.

PeterSurda (changes requested):
Ok but one thing is missing, it needs to be set `True` when running kivy tests.

src/bitmessagekivy/baseclass/network.py
```diff
@@ -13,12 +13,10 @@
 
 from pybitmessage import state
 
-if os.environ.get('INSTALL_TESTS', False):
+if os.environ.get('INSTALL_TESTS', False) and not state.backend_py3_compatible:
     from pybitmessage.mock import kivy_main
     stats = kivy_main.network.stats
     objectracker = kivy_main.network.objectracker
-else:
```
PeterSurda, line 20: this section should remain

PeterSurda (changes requested):

PeterSurda (approved):

PR #2100, opened 2022-11-02
Added set_common_header in mpybit

osamacis:
Added set_common_header in mpybit

### DIFF
--- a/src/bitmessagekivy/mpybit.py
+++ b/src/bitmessagekivy/mpybit.py
@@ -319,6 +319,14 @@ class NavigateApp(MDApp):
         if self.root.ids.id_newidentity.ids.add_random_bx.children:
             self.root.ids.id_newidentity.ids.add_random_bx.clear_widgets()
 
+    def set_common_header(self):
+        """Common header for all the Screens"""
+        self.root.ids.toolbar.right_action_items = [
+            ['account-plus', lambda x: self.addingtoaddressbook()]]
+        self.root.ids.toolbar.left_action_items = [
+            ['menu', lambda x: self.root.ids.nav_drawer.set_state("toggle")]]
+        return
+
     def open_payment_layout(self, sku):
         """It basically open up a payment layout for kivy UI"""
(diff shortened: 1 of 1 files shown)


src/bitmessagekivy/mpybit.py
```diff
@@ -319,6 +319,16 @@ def reset_login_screen(self):
         if self.root.ids.id_newidentity.ids.add_random_bx.children:
             self.root.ids.id_newidentity.ids.add_random_bx.clear_widgets()
 
+    def set_common_header(self):
+        """Common header for all windows"""
```
PeterSurda, line 323: better docstring, this is too general/inaccurate

src/bitmessagekivy/mpybit.py
```diff
@@ -319,6 +319,16 @@ def reset_login_screen(self):
...
+        self.root.ids.toolbar.right_action_items = [
+            ['account-plus', lambda x: self.addingtoaddressbook()]]
+        # self.root.ids.toolbar.left_action_items = [
```
PeterSurda, line 326: remove comment

src/bitmessagekivy/tests/test_network_screen.py
```diff
@@ -16,7 +16,7 @@ def test_network_status(self):
         self.assert_wait_no_except('//ScreenManager[@current]', timeout=15, value='login')
         # Method to open side navbar
         if self.cli.wait('//ActionTopAppBarButton[@icon~=\"menu\"]', timeout=5):
```
PeterSurda, line 18: I wouldn't use `if` due to asserts

PeterSurda (changes requested):

src/bitmessagekivy/mpybit.py
osamacis: Now this a common header which have two icons one is for side nav bar and one for add new contact address so this will appear in all screens if any address is already created

PeterSurda (approved):

PR #2104, opened 2022-11-07
Add label argument to createChan API endpoint

813492291816:
jsonrpclib throws errors sometimes due to the auto-generated labels that are based on the passphrases I set. Example:

```(-32603, '<class 'api.APIError'>:API Error 0021: Unexpected API Failure - bad interpolation variable reference '%(793569%(4"]'')```

This "%(793569%" is a substring of my label that's based on the passphrase and the error prevents any data from being received.

Because the passphrases I use are generated in a specific way, I can't just exclude certain character combinations as possibilities. Therefore, I propose adding the ability to set the label at the time of the chan creation, which will allow me to set the label to something different/safer than the passphrase.

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -779,10 +779,13 @@ class BMRPCDispatcher(object):
         return queues.apiAddressGeneratorReturnQueue.get()
 
     @command('createChan')
-    def HandleCreateChan(self, passphrase):
+    def HandleCreateChan(self, passphrase, label=None):
         """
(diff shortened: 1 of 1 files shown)


PeterSurda:
I think this should be fixed differently.

813492291816:
I don't know if this is an issue with jsonrpclib or bitmessage, or even how to debug it, but with regard to this PR, I see utility for a number of use cases in allowing the user to set the label of a new chan at creation.

PR #2112, opened 2022-11-10
Api fix `HandleDisseminatePreEncryptedMsg`

### LINKED ISSUE #2110: Incorrect stream offset in `HandleDisseminatePreEncryptedMsg`
A user wanted to use `HandleDisseminatePreEncryptedMsg` API call and it didn't appear to work. After reviewing the code and specification, it looks like when decoding destination stream:

https://github.com/Bitmessage/PyBitmessage/blob/a5a56fe7f1af86f5b8c40ee7d6307df62e0e6b3e/src/api.py#L1313

it uses a protocol v2 specification, and not protocol v3. In v2, the offset for the stream for a message starts at byte 16, but in v3 it starts at a variable position (at the moment it usually is byte 21, which allows for a quick workaround). This API method was introduced around the same time that v3 came out so I guess someone wasn't paying attention.

References:
v2 specification from wiki: https://wiki.bitmessage.org/index.php?title=Protocol_specification&oldid=23575#msg
v3 specification from rtd: https://pybitmessage.readthedocs.io/en/v0.6/protocol.html#object

PeterSurda:
- PoW target formula correction
- correctly decode expiration time / TTL, object type and destination stream
- fixes #2110

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -70,7 +70,7 @@ import time
 import xmlrpclib
 from binascii import hexlify, unhexlify
 from SimpleXMLRPCServer import SimpleXMLRPCRequestHandler, SimpleXMLRPCServer
-from struct import pack
+from struct import pack, unpack
 
 import defaults
 import helper_inbox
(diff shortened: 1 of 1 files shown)


PeterSurda:
tested, working

PR #2113, opened 2022-11-10
Fix kivy trash screen test case

osamacis:
Fix kivy trash screen test case and reordered sc/id


PeterSurda (changes requested):
remove commented code

PeterSurda (approved):

PR #2121, opened 2022-11-16
Refactor mpybit and add msg_composer dropdown

shekhar-cis:
### Description:

- Update in `mpybit.py`: Add `clear_composer()` to reset the fields of composer screen.
- Add `set_navbar_for_composer()` for setting required icons like `send`, `refresh`, `back_arrow`.
- Update `login.py`.
- Update `main.kv`


src/bitmessagekivy/baseclass/login.py
```diff
@@ -74,6 +74,8 @@ def reset_address_spinner(self):
                      if config.getboolean(str(addr), 'enabled')]
         self.manager.parent.ids.content_drawer.ids.identity_dropdown.values = []
         self.manager.parent.ids.content_drawer.ids.identity_dropdown.values = addresses
+        self.manager.parent.ids.id_create.children[1].ids.btn.values = []
```
PeterSurda, line 77: is this `btn` or was it renamed?

PeterSurda (changes requested):

src/bitmessagekivy/baseclass/login.py
shekhar-cis: Fixed

PeterSurda (changes requested):
Split into 2 commits. One with the renaming, and one with the new method.

PeterSurda (approved):

PR #2133, opened 2022-11-29
Refactor telenuim_process: Add a sleep(1) in tearDown

shekhar-cis:
### Code

- Refactor `telenium_process.py`, add `sleep(1)` in `teadDown()` before `super` call


PeterSurda:
Not needed anymore, we're not going forward with this approach.

PR #2136, opened 2022-12-12
Add localization for UI

osamacis:
Add localization for UI


PeterSurda (changes requested):
- remove the translated files and sources
- transform the `Makefile` into a  buildbot job
- add the translated binaries into `setup.py`

PR #2142, opened 2022-12-29
Add transifexrc file

osamacis:
Add transifexrc file for transifex translation


.tx/config
```diff
@@ -0,0 +1,10 @@
+[main]
```
PeterSurda, line 1: remove main section

PeterSurda (changes requested):

.tx/config
```diff
@@ -0,0 +1,6 @@
+[pybitmessage-test.messages-pot]
+file_filter = <lang>po
```
PeterSurda, line 2: subdirectory?

.tx/config
```diff
@@ -0,0 +1,6 @@
...
+minimum_perc = 0
+source_file = messages.pot
```
PeterSurda, line 4: put it into translations directory

PeterSurda (changes requested):

.tx/config
osamacis: yes i have put it here.

PeterSurda:
The file should be in project root

PR #2178, opened 2024-02-26
Code Quality

anand-skss:
# Code Quality improvement

* Suppressed try catch pass B110 bandit warning wherever required


PeterSurda (approved):
- dandelion fluff randomisation does not need to be cryptographically secure 
- bare excepts in this PR are either in OS-level critical code which can't be improved without thorough tests, or,
- as in case of the network thread, will be replaced by a library which is python3 compatible, so there's no need to fix it if it will be removed.

PR #2183, opened 2024-02-27
Try renaming readthedocs config file

PeterSurda:
Hi!

I found also [this](https://docs.readthedocs.io/en/stable/config-file/v2.html#configuration-file-v2-readthedocs-yaml):
```
Read the Docs supports configuring your documentation builds with a configuration file. This file is named .readthedocs.yaml and should be placed in the top level of your Git repository.
```

Something strange has happened with readthedocs config versioning in January. The previous config was also v2.

I hope the renaming helps.


PeterSurda:
Not needed, the problem was a missing secret in the webhook configuration and ReadTheDocs started enforcing this on January 31st.

PR #2201, opened 2024-03-28
Test cases for shared.py

anand-skss:
* Test cases for shared.py

### DIFF
--- a/src/tests/test_shared.py
+++ b/src/tests/test_shared.py
@@ -1,0 +1,152 @@
+"""Test cases for shared.py"""
+
+import unittest
+from pybitmessage.shared import (
+    isAddressInMyAddressBook,
+    isAddressInMySubscriptionsList,
+    checkSensitiveFilePermissions,
+    reloadBroadcastSendersForWhichImWatching,
+    fixSensitiveFilePermissions,
+    MyECSubscriptionCryptorObjects,
+    stat,
+    os,
+)
+
+from .samples import sample_address
+
+try:
+    # Python 3
+    from unittest.mock import patch, PropertyMock
+except ImportError:
+    # Python 2
+    from mock import patch, PropertyMock
+
+# mock os.stat data for file
+PERMISSION_MODE1 = stat.S_IRUSR  # allow Read permission for the file owner.
+PERMISSION_MODE2 = (
+    stat.S_IRWXO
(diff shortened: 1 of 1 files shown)


src/tests/test_shared.py
```diff
@@ -0,0 +1,130 @@
...
+                "st_uid": 1000,
+                "st_gid": 0,
+                "st_size": 1021,
+                "st_atime": 1711587560,
+                "st_mtime": 1709449249,
+                "st_ctime": 1709449603,
+            },
+        )
+        fixSensitiveFilePermissions(filename, False)
+        mock_chmod.asset_called_once()
```
PeterSurda, line 130: this looks like a typo, it should be `assert_called_once`

src/tests/test_shared.py
```diff
@@ -0,0 +1,130 @@
...
+    from mock import patch, PropertyMock
+
+
+class TestShared(unittest.TestCase):
+    """Test class for shared.py"""
+
+    @patch("pybitmessage.shared.sqlQuery")
+    def test_isaddress_in_myaddressbook(self, mock_sql_query):
+        """Test if address is in MyAddressbook"""
+        address = "BM-2cUGaEcGz9Zft1SPAo8FJtfzyADTpEgU9U"
```
PeterSurda, line 29: maybe use addresses from samples

src/tests/test_shared.py
```diff
@@ -0,0 +1,130 @@
...
+        result = checkSensitiveFilePermissions(filename)
+        self.assertTrue(result)
+
+        # test for freebsd system
+        mock_sys.platform = "freebsd7"
+        mock_os_stat.return_value = os.stat_result(
+            sequence=[0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
+            dict={
+                "st_mode": stat.S_IRUSR,
+                "st_ino": 753,
```
PeterSurda, line 93: use constants

PeterSurda (changes requested):

src/tests/test_shared.py
anand-skss: os.stat returns stat_result type object, so, while mocking return value stat_obj is required.

PeterSurda: I mean the numbers should be constants

PeterSurda (approved):

PR #2215, opened 2024-04-18
moved inventory in state - global runtime variable from singleton

anand-skss:
* moved inventory in state - global runtime variable from singleton

### DIFF
--- a/src/inventory.py
+++ b/src/inventory.py
@@ -1,10 +1,9 @@
-"""The Inventory singleton"""
+"""The Inventory"""
 
 # TODO make this dynamic, and watch out for frozen, like with messagetypes
 import storage.filesystem
 import storage.sqlite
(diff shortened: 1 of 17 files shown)


src/inventory.py
```diff
@@ -17,8 +16,7 @@ def create_inventory_instance(backend="sqlite"):
         "{}Inventory".format(backend.title()))()
 
 
-@Singleton
-class Inventory():
+class Inventory:
     """
     Inventory singleton class which uses storage backends
```
PeterSurda, line 21: update the docstring to reflect it's not a singleton anymore but it's supposed to be a variable in global state.

PeterSurda (changes requested):

PeterSurda (approved):

PR #2217, opened 2024-04-25
Moved BMConnectionPool in state - global runtime variable from singleton

anand-skss:
* Moved BMConnectionPool in state - global runtime variable from singleton


PeterSurda (approved):
There are some minor things that may be done differently but for the time being it's ok.

PeterSurda:
superseded by #2220

PR #2220, opened 2024-05-01
Made BMConnectionPool as global runtime variable in connectionpool from singleton

anand-skss:
* Made BMConnectionPool as global runtime variable in connectionpool from singleton
* Moved announcethread and connectionpool imports into start function

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -96,9 +96,9 @@ from helper_sql import (
 from highlevelcrypto import calculateInventoryHash
 
 try:
-    from network import BMConnectionPool
+    from network import connectionpool
 except ImportError:
-    BMConnectionPool = None
+    connectionpool = None
 
 from network import stats, StoppableThread
 from version import softwareVersion
@@ -1475,18 +1475,18 @@ class BMRPCDispatcher(object):
         Returns bitmessage connection information as dict with keys *inbound*,
         *outbound*.
         """
-        if BMConnectionPool is None:
+        if connectionpool is None:
             raise APIError(21, 'Could not import BMConnectionPool.')
         inboundConnections = []
         outboundConnections = []
-        for i in BMConnectionPool().inboundConnections.values():
+        for i in connectionpool.pool.inboundConnections.values():
             inboundConnections.append({
                 'host': i.destination.host,
                 'port': i.destination.port,
                 'fullyEstablished': i.fullyEstablished,
                 'userAgent': str(i.userAgent)
             })
-        for i in BMConnectionPool().outboundConnections.values():
+        for i in connectionpool.pool.outboundConnections.values():
             outboundConnections.append({
                 'host': i.destination.host,
                 'port': i.destination.port,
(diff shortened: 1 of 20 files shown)


src/api.py
```diff
@@ -1475,18 +1475,18 @@ def HandleListConnections(self):
         Returns bitmessage connection information as dict with keys *inbound*,
         *outbound*.
         """
-        if BMConnectionPool is None:
+        if network.BMConnectionPool is None:
```
PeterSurda, line 1478: If we don't conditionally import the connection pool then this check isn't necessary anymore.

src/network/__init__.py
```diff
@@ -4,19 +4,19 @@
 
 try:
     from .announcethread import AnnounceThread
-    from .connectionpool import BMConnectionPool
+    from .connectionpool import BMConnectionPool as BMConnectionPool_cls
+    BMConnectionPool = BMConnectionPool_cls()
```
PeterSurda, line 8: how about we use a different name for the object than the class name? How about `network.connectionpool` for instance and we keep `network.BMConnectionPool` as the class name?

src/network/__init__.py
```diff
@@ -4,19 +4,19 @@
...
 except ImportError:
```
PeterSurda, line 9: I'm not sure this try/except is needed anymore once we fix the circular imports.

src/network/__init__.py
```diff
@@ -4,19 +4,19 @@
...
     AnnounceThread = None
     BMConnectionPool = None
 from .threads import StoppableThread
 
 
-__all__ = ["AnnounceThread", "BMConnectionPool", "StoppableThread"]
```
PeterSurda, line 14: I think we can use the object name here, i.e. `connectionpool` as I recommended above.

src/network/addrthread.py
```diff
@@ -5,10 +5,10 @@
 
 # magic imports!
 import state
+import network
```
PeterSurda, line 8: why not `from network import connectionpool`?

src/network/announcethread.py
```diff
@@ -5,9 +5,9 @@
 
 # magic imports!
 import state
+import network
```
PeterSurda, line 8: again here we could use `from network import connectionpool`

PeterSurda (changes requested):

src/network/__init__.py
anand-skss: done, renamed as pool

src/network/addrthread.py
anand-skss: because it will create a new reference/new copy of connectionpool obj which would be different from network one.

src/network/announcethread.py
anand-skss: same reason as above, and other thing it is also throwing import error

PeterSurda (approved):

PR #2222, opened 2024-05-09
Remove state.shutdown or replaced with self._stopped from some of the network threads

anand-skss:
* Removed state dependency from some of the network threads -> addrthread, announcethread, networkthread, recievequeuethread

### DIFF
--- a/src/network/announcethread.py
+++ b/src/network/announcethread.py
@@ -20,7 +20,7 @@ class AnnounceThread(StoppableThread):
 
     def run(self):
         lastSelfAnnounced = 0
-        while not self._stopped and state.shutdown == 0:
+        while not self._stopped:
             processed = 0
             if lastSelfAnnounced < time.time() - self.announceInterval:
                 self.announceSelf()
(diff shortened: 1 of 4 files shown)


src/network/announcethread.py
```diff
@@ -34,7 +33,7 @@ def announceSelf():
         for connection in connectionpool.pool.udpSockets.values():
             if not connection.announcing:
                 continue
-            for stream in state.streamsInWhichIAmParticipating:
+            for stream in connectionpool.pool.streams:
```
PeterSurda, line 36: `steamsInWhichIAmParticipating` should be done separately from `state.shutdown`

PeterSurda (changes requested):

src/network/announcethread.py
anand-skss: separated this in two diff commits

PeterSurda: What I meant was, this PR should only deal with `state.shutdown`. There should be a separate PR, which removes the variable `state.streamsInWhichIAmParticipating` and replaces it just with `pool.streams`.

PeterSurda (approved):

PR #2226, opened 2024-05-13
Replaced state.streamsInWhichIAmParticipating with pool.streams

anand-skss:
* Replaced state.streamsInWhichIAmParticipating with pool.streams

### DIFF
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -237,8 +237,7 @@ class Main(object):
                 upnpThread = upnp.uPnPThread()
                 upnpThread.start()
         else:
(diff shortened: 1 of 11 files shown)


src/bitmessagemain.py
```diff
@@ -237,8 +238,7 @@ def start(self):
                 upnpThread = upnp.uPnPThread()
                 upnpThread.start()
         else:
-            # Populate with hardcoded value (same as connectToStream above)
-            state.streamsInWhichIAmParticipating.append(1)
```
anand-skss, line 241: This make difference in streamsInWhichIAmParticipating & pool streams

PeterSurda (approved):

PR #2237, opened 2024-05-21
Merge branch 'v0.6' into qt5-wip

kashikoibumi:
I have merged the branch 'v0.6' into 'qt5-wip' in 'Porting bitmessageqt to Qt5 #1389' by @g1itch .

It seems work well in Python2.7 + Qt.py + PyQt5.
But, it seems not work with PyQt4.
I will continue investigating this issue.


kashikoibumi:
In order to run with both PyQt4 and PyQt5, Qt.py (qtpy) should be used directly without using the fallback 'PyQt5', although Qt.py becomes a must-dependency.
It seems OK, so I'm going on.

kashikoibumi:
Now PyBitmessage runs with either of PyQt5 or PyQt4 by using Qt.py.

But, does not pass the tests.py ..
I have no idea for this currently.

kashikoibumi:
I also tried to run with PySide2, but failed with some errors.
It may be sufficient to be runnable with PyQt5 and PyQt4, so I don't seek a path with PySide2 now.

PR #2244, opened 2024-05-26
Moved dandelion runtime var from state to network->dandelion

anand-skss:
* Moved dandelion runtime var from state to network->dandelion

### DIFF
--- a/src/network/bmproto.py
+++ b/src/network/bmproto.py
@@ -15,6 +15,7 @@ import addresses
 import knownnodes
 import protocol
 import state
+import dandelion
 import connectionpool
 from bmconfigparser import config
 from queues import invQueue, objectProcessorQueue, portCheckerQueue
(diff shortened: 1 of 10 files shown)


src/network/bmproto.py
```diff
@@ -337,7 +338,7 @@ def bm_command_getdata(self):
             self.pendingUpload[str(i)] = now
         return True
 
-    def _command_inv(self, dandelion=False):
+    def _command_inv(self, dandelion_enabled=False):
```
PeterSurda, line 341: I wouldn't use `dandelion_enabled` because this is confusing. It's a bit difficult to come up with a good description, I think `extend_dandelion_stem` is the best I could do.

The dandelion functionality in Bitmessage is an implementation of the Dandelion++ proposal for Bitcoin (https://en.bitcoin.it/wiki/BIP_0156). The `tx` and `dandeliontx` network messages in Bitcoin correspond to `inv` and `dinv` commands in Bitmessage.

PeterSurda (changes requested):

PR #2246, opened 2024-05-29
moved dandelion_enabled from state to dandelion instance as enabled attribute

anand-skss:
* moved dandelion_enabled from state to Dandelion  class as enabled attribute
* Move dandelion enabled checks from main into a function is_dandelion_enabled under Dandelion class

### DIFF
--- a/src/bitmessagemain.py
+++ b/src/bitmessagemain.py
@@ -12,6 +12,7 @@ The PyBitmessage startup script
 import os
 import sys
 
+
 try:
     import pathmagic
 except ImportError:
@@ -155,13 +156,6 @@ class Main(object):
         self.setSignalHandler()
 
         set_thread_name("PyBitmessage")
-
-        state.dandelion_enabled = config.safeGetInt('network', 'dandelion')
-        # dandelion requires outbound connections, without them,
-        # stem objects will get stuck forever
-        if state.dandelion_enabled and not config.safeGetBoolean(
-                'bitmessagesettings', 'sendoutgoingconnections'):
-            state.dandelion_enabled = 0
 
         if state.testmode or config.safeGetBoolean(
                 'bitmessagesettings', 'extralowdifficulty'):
--- a/src/network/__init__.py
+++ b/src/network/__init__.py
@@ -1,9 +1,10 @@
 """
 Network subsystem package
 """
-
+from .dandelion import Dandelion
 from .threads import StoppableThread
 
+dandelion_ins = Dandelion()
 
 __all__ = ["StoppableThread"]
 
@@ -20,6 +21,11 @@ def start(config, state):
     from .knownnodes import readKnownNodes
     from .receivequeuethread import ReceiveQueueThread
     from .uploadthread import UploadThread
+
+    # check and set dandelion enabled value at network startup
+    dandelion_ins.init_dandelion_enabled(config)
+    # pass pool instance into dandelion class instance
+    dandelion_ins.init_pool(connectionpool.pool)
 
     readKnownNodes()
     connectionpool.pool.connectToStream(1)
--- a/src/network/bmproto.py
+++ b/src/network/bmproto.py
@@ -15,7 +15,6 @@ import addresses
 import knownnodes
 import protocol
 import state
-import dandelion
 import connectionpool
 from bmconfigparser import config
 from queues import invQueue, objectProcessorQueue, portCheckerQueue
@@ -27,7 +26,7 @@ from network.bmobject import (
     BMObjectUnwantedStreamError
 )
 from network.proxy import ProxyError
-
+from network import dandelion_ins
 from node import Node, Peer
 from objectracker import ObjectTracker, missingObjects
 
@@ -351,14 +350,14 @@ class BMProto(AdvancedDispatcher, ObjectTracker):
             raise BMProtoExcessiveDataError()
 
         # ignore dinv if dandelion turned off
-        if extend_dandelion_stem and not state.dandelion_enabled:
+        if extend_dandelion_stem and not dandelion_ins.enabled:
             return True
 
         for i in map(str, items):
-            if i in state.Inventory and not dandelion.instance.hasHash(i):
+            if i in state.Inventory and not dandelion_ins.hasHash(i):
                 continue
-            if extend_dandelion_stem and not dandelion.instance.hasHash(i):
-                dandelion.instance.addHash(i, self)
+            if extend_dandelion_stem and not dandelion_ins.hasHash(i):
+                dandelion_ins.addHash(i, self)
             self.handleReceivedInventory(i)
 
         return True
@@ -420,9 +419,9 @@ class BMProto(AdvancedDispatcher, ObjectTracker):
             except KeyError:
                 pass
 
-        if self.object.inventoryHash in state.Inventory and dandelion.instance.hasHash(
+        if self.object.inventoryHash in state.Inventory and dandelion_ins.hasHash(
                 self.object.inventoryHash):
-            dandelion.instance.removeHash(
+            dandelion_ins.removeHash(
                 self.object.inventoryHash, "cycle detection")
 
         state.Inventory[self.object.inventoryHash] = (
@@ -541,7 +540,7 @@ class BMProto(AdvancedDispatcher, ObjectTracker):
         if not self.isOutbound:
             self.append_write_buf(protocol.assembleVersionMessage(
                 self.destination.host, self.destination.port,
-                connectionpool.pool.streams, True,
+                connectionpool.pool.streams, dandelion_ins.enabled, True,
                 nodeid=self.nodeid))
             logger.debug(
                 '%(host)s:%(port)i sending version',
--- a/src/network/dandelion.py
+++ b/src/network/dandelion.py
@@ -7,9 +7,6 @@ from random import choice, expovariate, sample
 from threading import RLock
 from time import time
 
-import connectionpool
-import state
-from queues import invQueue
 
 # randomise routes after 600 seconds
 REASSIGN_INTERVAL = 600
@@ -37,6 +34,8 @@ class Dandelion:  # pylint: disable=old-style-class
         # when to rerandomise routes
         self.refresh = time() + REASSIGN_INTERVAL
         self.lock = RLock()
+        self.enabled = None
+        self.pool = None
 
     @staticmethod
     def poissonTimeout(start=None, average=0):
@@ -47,10 +46,23 @@ class Dandelion:  # pylint: disable=old-style-class
             average = FLUFF_TRIGGER_MEAN_DELAY
         return start + expovariate(1.0 / average) + FLUFF_TRIGGER_FIXED_DELAY
 
+    def init_pool(self, pool):
+        """pass pool instance"""
+        self.pool = pool
+
+    def init_dandelion_enabled(self, config):
+        """Check if Dandelion is enabled and set value in enabled attribute"""
+        dandelion_enabled = config.safeGetInt('network', 'dandelion')
+        # dandelion requires outbound connections, without them,
+        # stem objects will get stuck forever
+        if not config.safeGetBoolean(
+                'bitmessagesettings', 'sendoutgoingconnections'):
+            dandelion_enabled = 0
+        self.enabled = dandelion_enabled
+
     def addHash(self, hashId, source=None, stream=1):
-        """Add inventory vector to dandelion stem"""
-        if not state.dandelion_enabled:
-            return
+        """Add inventory vector to dandelion stem return status of dandelion enabled"""
+        assert self.enabled is not None
         with self.lock:
             self.hashMap[hashId] = Stem(
                 self.getNodeStem(source),
@@ -89,7 +101,7 @@ class Dandelion:  # pylint: disable=old-style-class
         """Child (i.e. next) node for an inventory vector during stem mode"""
         return self.hashMap[hashId].child
 
-    def maybeAddStem(self, connection):
+    def maybeAddStem(self, connection, invQueue):
         """
         If we had too few outbound connections, add the current one to the
         current stem list. Dandelion as designed by the authors should
@@ -163,7 +175,7 @@ class Dandelion:  # pylint: disable=old-style-class
                 self.nodeMap[node] = self.pickStem(node)
                 return self.nodeMap[node]
 
-    def expire(self):
+    def expire(self, invQueue):
         """Switch expired objects from stem to fluff mode"""
         with self.lock:
             deadline = time()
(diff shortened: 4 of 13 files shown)


src/bitmessagemain.py
```diff
@@ -156,12 +157,9 @@ def start(self):
...
         set_thread_name("PyBitmessage")
 
-        state.dandelion_enabled = config.safeGetInt('network', 'dandelion')
-        # dandelion requires outbound connections, without them,
-        # stem objects will get stuck forever
-        if state.dandelion_enabled and not config.safeGetBoolean(
-                'bitmessagesettings', 'sendoutgoingconnections'):
-            state.dandelion_enabled = 0
+        # dandelion enabled checks at startup
+        network.dandelion.instance.is_dandelion_enabled(config)
```
PeterSurda, line 161: how about making this into a static method that returns `True` / `False`? Then the instance and the config are decoupled, you pass the result of `is_dandelion_enabled` do the constructor.

src/network/bmproto.py
```diff
@@ -351,7 +351,7 @@ def _command_inv(self, extend_dandelion_stem=False):
             raise BMProtoExcessiveDataError()
 
         # ignore dinv if dandelion turned off
-        if extend_dandelion_stem and not state.dandelion_enabled:
+        if extend_dandelion_stem and not dandelion.instance.enabled:
```
PeterSurda, line 354: what we can do here is to move this check into the `Dandelion.hasHash` method. Then bmproto doesn't need to check.

src/network/invthread.py
```diff
@@ -42,7 +42,7 @@ def handleLocallyGenerated(stream, hashId):
         """Locally generated inventory items require special handling"""
         dandelion.instance.addHash(hashId, stream=stream)
```
PeterSurda, line 43: here, `addHash` could return `True` / `False` depending on whether dandelion is on/off, and we could use this for further checks.

src/network/invthread.py
```diff
@@ -42,7 +42,7 @@ def handleLocallyGenerated(stream, hashId):
...
         for connection in connectionpool.pool.connections():
-            if state.dandelion_enabled and connection != \
+            if dandelion.instance.enabled and connection != \
```
PeterSurda, line 45: here, `objectChildStem` could return value depending on whether dandelion is enabled or not

src/network/invthread.py
```diff
@@ -78,7 +78,7 @@ def run(self):  # pylint: disable=too-many-branches
                             if connection == dandelion.instance.objectChildStem(inv[1]):
                                 # Fluff trigger by RNG
                                 # auto-ignore if config set to 0, i.e. dandelion is off
-                                if random.randint(1, 100) >= state.dandelion_enabled:  # nosec B311
+                                if random.randint(1, 100) >= dandelion.instance.enabled:  # nosec B311
```
PeterSurda, line 81: Hmm, I forgot, the value is actually an int between 0 and 100. We need to rethink.

src/network/invthread.py
```diff
@@ -106,6 +106,6 @@ def run(self):  # pylint: disable=too-many-branches
                 invQueue.task_done()
 
             if dandelion.instance.refresh < time():
-                dandelion.instance.reRandomiseStems()
+                dandelion.instance.reRandomiseStems(connectionpool.pool)
```
PeterSurda, line 109: this check could also be moved down (i.e. `reRandomiseStems` could check check the time.

PeterSurda (changes requested):

src/bitmessagemain.py
anand-skss: but the problem is we need a variable where we can store the return value. currently after checking we are storing in class attr.

src/network/bmproto.py
anand-skss: we can move into Dandelion.hasHash. but `hasHash` here used in some checks in iteration. i think that could not work same as here, like return direct True.

src/network/invthread.py
anand-skss: but for there places we have to use `dandelion.instace.enabled` in this file

src/bitmessagemain.py
PeterSurda: Hmm, now I see the problem. We then probably could create a `Placeholder` instance inside `dandelion.py` and create a proper instance here in `bitmessagemain.py`. Then we can pass the value to the constructor.

src/network/bmproto.py
PeterSurda: The easiest solution is just to use the same approach in `if` as in `for`. It would result in dummy loops but my guess is that it's not a big performance hit. It could be optimised for performance later if needed.

src/bitmessagemain.py
anand-skss: invthread.py", line 55, in run
handleExpiredDandelion(dandelion.instance.expire())

- ivthread initialized before dandelion instance created and checking for dandelion expire. real dandelion instance is required at the starting point of thread.

src/tests/core.py
```diff
@@ -320,15 +321,15 @@ def _decode_msg(data, pattern):
     def test_version(self):
         """check encoding/decoding of the version message"""
         # with single stream
-        msg = protocol.assembleVersionMessage('127.0.0.1', 8444, [1])
+        msg = protocol.assembleVersionMessage('127.0.0.1', 8444, [1], dandelion_ins.enabled)
```
PeterSurda, line 324: we don't really need an instance of dandelion here. We just need a variable and we pass the same variable to `assembleVersionMessage` and `assertEqual`

src/protocol.py
```diff
@@ -337,7 +337,7 @@ def assembleAddrMessage(peerList):
 
 
 def assembleVersionMessage(
-    remoteHost, remotePort, participatingStreams, server=False, nodeid=None
+    remoteHost, remotePort, participatingStreams, dandelion_enabled, server=False, nodeid=None,
```
PeterSurda, line 340: We can have `dandelion_enabled` default to `True`. Then we can put it at the end of arguments as well.

src/network/dandelion.py
```diff
@@ -177,21 +195,19 @@ def expire(self):
                 invQueue.put(row)
         return toDelete
 
-    def reRandomiseStems(self):
+    def reRandomiseStems(self, pool):
```
PeterSurda, line 198: We could use lazy initialisation on `pool` too and keep it as an attribute, then we don't need to pass it later.

PeterSurda (changes requested):

src/network/bmproto.py
```diff
@@ -27,7 +26,7 @@
     BMObjectUnwantedStreamError
 )
 from network.proxy import ProxyError
-
+from network import dandelion_ins
```
PeterSurda, line 29: Why in `bmobject` it does `from . import dandelion_ins` and here it does `from network import dandelion_ins`? Why is it inconsistent?

src/network/__init__.py
```diff
@@ -21,6 +22,11 @@ def start(config, state):
     from .receivequeuethread import ReceiveQueueThread
     from .uploadthread import UploadThread
 
+    # lazy init of dandelion enabled network startup
```
PeterSurda, line 25: It should maybe say "pass instance", I don't think "lazy init" is a good comment. We're not lazy initialising is_enabled or pool, we're lazy initialising dandelion.

src/network/dandelion.py
```diff
@@ -47,15 +49,36 @@ def poissonTimeout(start=None, average=0):
             average = FLUFF_TRIGGER_MEAN_DELAY
         return start + expovariate(1.0 / average) + FLUFF_TRIGGER_FIXED_DELAY
 
+    def init_pool(self, pool):
+        """Lazy intialization of pool attr"""
```
PeterSurda, line 53: Maybe "pass pool instance"

src/network/dandelion.py
```diff
@@ -47,15 +49,36 @@ def poissonTimeout(start=None, average=0):
...
+        self.pool = pool
+
+    def init_is_enabled(self, config):
+        """Lazy intialization of enabled attr"""
```
PeterSurda, line 57: Do we need a separate `init_is_enabled` and `is_dandelion_enabled`?

src/network/dandelion.py
```diff
@@ -86,8 +109,8 @@ def hasHash(self, hashId):
         return hashId in self.hashMap
 
     def objectChildStem(self, hashId):
-        """Child (i.e. next) node for an inventory vector during stem mode"""
-        return self.hashMap[hashId].child
+        """Returns Child (i.e. next) node for an inventory vector during stem mode and dandelion enabled status"""
+        return (self.hashMap[hashId].child, self.enabled)
```
PeterSurda, line 113: this isn't necessary, you can retrieve enabled directly as an object attribute

src/network/dandelion.py
```diff
@@ -179,19 +202,20 @@ def expire(self):
 
     def reRandomiseStems(self):
         """Re-shuffle stem mapping (parent <-> child pairs)"""
+        assert self.pool is not None
+        if self.refresh > time():
+            return
+        if self.refresh > time():
```
PeterSurda, line 208: duplicate

src/network/invthread.py
```diff
@@ -75,10 +75,11 @@ def run(self):  # pylint: disable=too-many-branches
                         except KeyError:
                             continue
                         try:
-                            if connection == dandelion.instance.objectChildStem(inv[1]):
+                            childstem, dandelion_enabled = dandelion_ins.objectChildStem(inv[1])
+                            if connection == childstem:
                                 # Fluff trigger by RNG
                                 # auto-ignore if config set to 0, i.e. dandelion is off
-                                if random.randint(1, 100) >= state.dandelion_enabled:  # nosec B311
+                                if random.randint(1, 100) >= dandelion_enabled:  # nosec B311
```
PeterSurda, line 82: we can just use `dandelion_ins.enabled`

PeterSurda (changes requested):

src/network/invthread.py
anand-skss: As per this (https://github.com/Bitmessage/PyBitmessage/pull/2246#discussion_r1618334811) , here `objectChildStem` is returning  value whether dandelion is enabled or not and that is used here. do i need to modify it like it was before?

src/network/dandelion.py
anand-skss: Fixed

src/network/dandelion.py
anand-skss: https://github.com/Bitmessage/PyBitmessage/pull/2246#discussion_r1618334811 , I've modified it.

src/network/dandelion.py
anand-skss: merged it to `is_dandelion_enabled` by changed it into classmethod from staticmethod. and removed `init_is_enabled`.

src/network/__init__.py
anand-skss: changed comment.

src/network/dandelion.py
anand-skss: Done

src/network/bmproto.py
anand-skss: Chenged it to `from network import dandelion_ins` everywhere.

src/network/dandelion.py
anand-skss: Done

src/tests/core.py
anand-skss: Done

src/bitmessagemain.py
anand-skss: Resolved now. moved Dandelion instance  into `network/__init__.py`

src/network/invthread.py
anand-skss: when dandelion is on what would be the probable return value. that help us resolve the fixing comparison between `int` and `bool`

src/network/invthread.py
```diff
@@ -40,10 +40,10 @@ class InvThread(StoppableThread):
     @staticmethod
     def handleLocallyGenerated(stream, hashId):
         """Locally generated inventory items require special handling"""
-        dandelion.instance.addHash(hashId, stream=stream)
+        dandelion_enbled = dandelion_ins.addHash(hashId, stream=stream)
```
anand-skss, line 43: As per https://github.com/Bitmessage/PyBitmessage/pull/2246#discussion_r1618334222. i have returned dandelion enabled value here as we. do i need to use `dandelion_ins.enabled` here as well?

src/network/dandelion.py
```diff
@@ -47,15 +49,31 @@ def poissonTimeout(start=None, average=0):
             average = FLUFF_TRIGGER_MEAN_DELAY
         return start + expovariate(1.0 / average) + FLUFF_TRIGGER_FIXED_DELAY
 
+    def init_pool(self, pool):
+        """pass pool instance"""
+        self.pool = pool
+
+    def is_dandelion_enabled(self, config):
```
PeterSurda, line 56: I would call this `init_dandelion_enabled`

src/network/dandelion.py
```diff
@@ -6,10 +6,10 @@
...
 from threading import RLock
 from time import time
-
-import connectionpool
-import state
-from queues import invQueue
+try:
+    from queues import invQueue
+except ImportError:
+    from pybitmessage.queues import invQueue
```
PeterSurda, line 12: I would also pass `invQueue` using dependency injection. Then we can simplify import.

src/network/dandelion.py
```diff
@@ -47,15 +49,31 @@ def poissonTimeout(start=None, average=0):
...
+    def init_pool(self, pool):
+        """pass pool instance"""
+        self.pool = pool
+
+    def is_dandelion_enabled(self, config):
+        """Check if Dandelion is enabled and set value in enabled attribute"""
+        dandelion_enabled = config.safeGetInt('network', 'dandelion')
+        # dandelion requires outbound connections, without them,
+        # stem objects will get stuck forever
+        if dandelion_enabled and not config.safeGetBoolean(
```
PeterSurda, line 61: We probably don't need to check for `dandelion_enabled` here.

src/network/invthread.py
PeterSurda: I think using `dandelion_ins.enabled` makes more sense.

PeterSurda (changes requested):

src/network/dandelion.py
anand-skss: Value is coming from config  at startup. so i think it is required.

src/network/invthread.py
anand-skss: Done

src/network/dandelion.py
anand-skss: done

src/network/dandelion.py
PeterSurda: I mean we're checking twice, should work ok if we just check once.

src/network/dandelion.py
anand-skss: At Constructor level, dependency injection is not working, having same issue. Dandelion class have two fns having invQueue dependency. `from pybitmessage.queues import invQueue` is working but for portable it fails.

anand-skss: i think try except option would be better here until project fully migrated to py3

anand-skss: Used dependency injection at function level.

PeterSurda (approved):

Issue #2247, opened 2024-05-29
msgid is stored as text type


kashikoibumi:
msgid is stored as text type.
It should be binary blob.
```
$ sqlite3 messages.dat
SQLite version 3.40.1 2022-12-28 14:03:47
Enter ".help" for usage hints.
sqlite> SELECT TYPEOF(msgid) FROM inbox LIMIT 1;
text
```

hash in inventory is OK.
```
sqlite> SELECT TYPEOF(hash) FROM inventory LIMIT 1;
blob
```

PeterSurda:
This is true, but there are potential problems with upgrades. That's why the sql thread should be refactored first and added tests. These are the PRs that tried that but they are too big and need cleaning up: #1794 #1999 #2150

kashikoibumi:
The type of values whose declared type is blob:
* inbox.msgid : text
* inbox.sighash : text
* inventory.hash : blob
* inventory.payload : blob
* inventory.tag : blob
* objectprocessorqueue.data : (empty)
* pubkeys.transmitdata : text
* sent.msgid : text
* sent.toripe : text
* sent.ackdata : text
* settings.key : text
* settings.value : text, integer

PeterSurda:
It actually isn't consistent across versions, there may be older databases which are different. That's why we need tests for this, even more for the data type change.

kashikoibumi:
How about this:
* Change all code at once to store all blob data as blob by using sqlite3.Binary(). Then,
* Search any blob data:
  * First, try SELECT by blob keys using sqlite3.Binary(key).
  * When there is no match found, as a fallback, try SELECT by text key using CAST(? AS TEXT) in SQLite sintax.

I think this should be compatible to any past versions and future updates, including migrations to Python3.

kashikoibumi:
I have implemented a quick workaround: #2248

PR #2248, opened 2024-05-30
Quick workaround for BLOB as TEXT problem (#2247)

kashikoibumi:
Quick workaround for BLOB as TEXT problem (#2247).

This patch fixes each SQL query to be both workable with BLOB-keys and TEXT-keys.
In each query, it tries first BLOB-key by using sqlite3.Binary() function.
When it failed, then it tries next TEXT-key by using CAST(? AS TEXT) syntax.

The modifications are all around the code base where SQL queries exist, but the logic is simple.

### DIFF
--- a/src/api.py
+++ b/src/api.py
@@ -67,6 +67,7 @@ import subprocess  # nosec B404
 import time
 from binascii import hexlify, unhexlify
 from struct import pack, unpack
+import sqlite3
 
 import six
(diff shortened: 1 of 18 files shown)


kashikoibumi:
I have added a script to revert the BLOB-keys into TEXT-keys.
* revert_blob_to_text.sh

By applying this script, old version of PyBitmessage before this patch can be run again.

PR #2250, opened 2024-05-30
Runnable with both Python3 and Python2, with both PyQt5 and PyQt4 by using Qt.py

kashikoibumi:
Runnable with both Python3 and Python2, with both PyQt5 and PyQt4 by using Qt.py.


PeterSurda:
Can I create a gitea account for you on a private instance? You'll have more access to code and reviews by other developers that way. I only need an email address for that, you can access is through tor. I don't have is codespaces support in gitea but you don't use that anyway so there shouldn't be any drawbacks.

kashikoibumi:
I hope to access to the Gitea! My email address is now shown on my public profile.

PR #2256, opened 2024-06-23
Moved Multiqueue, invQueue and addrQueue to network module

anand-skss:
* Moved Multiqueue, invQueue and addrQueue to network module

### DIFF
--- a/src/network/addrthread.py
+++ b/src/network/addrthread.py
@@ -7,7 +7,7 @@ from six.moves import queue
 # magic imports!
 import connectionpool
 from protocol import assembleAddrMessage
-from queues import addrQueue  # FIXME: init with queue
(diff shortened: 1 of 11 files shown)


src/network/addrthread.py
```diff
@@ -7,7 +7,7 @@
 # magic imports!
 import connectionpool
 from protocol import assembleAddrMessage
-from queues import addrQueue  # FIXME: init with queue
```
anand-skss, line 10: Please check Fixme comment, i'm not sure whether i remove this or not.

PeterSurda: I'm not sure either, let's leave it as it is.

PeterSurda (approved):
Looks ok, just next time I'd put the helper_random -> random change into a separate PR.

PR #2257, opened 2024-06-24
Moved portCheckerQueue and receiveDataQueue to network module

anand-skss:
* Moved portCheckerQueue and receiveDataQueue to network module

### DIFF
--- a/src/network/connectionchooser.py
+++ b/src/network/connectionchooser.py
@@ -5,11 +5,14 @@ Select which node to connect to
 import logging
 import random
 
+from six.moves import queue
+
 import knownnodes
 import protocol
 import state
+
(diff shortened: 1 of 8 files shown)


src/network/connectionchooser.py
```diff
@@ -9,7 +9,7 @@
 import protocol
 import state
 from bmconfigparser import config
-from queues import queue, portCheckerQueue
+from network import queue, portCheckerQueue
```
PeterSurda, line 12: it's probably better to split like this:

```
from six.moves import queue
from network import portCheckerQueue
```

PeterSurda (changes requested):

PeterSurda (approved):

PR #2269, opened 2024-10-09
kivy-code-quality-2

PeterSurda:


PeterSurda:
I'll override the code quality check, because the `super` it complains about is in python3-only code, and there it's correct.

